PR #68, opened 2023-03-17
chore: Fix release steps for node server. Update manual release.

kinyoklion:


.github/workflows/release-please.yml
```diff
@@ -65,7 +65,7 @@ jobs:
     permissions:
       id-token: write
       contents: write
-    if: ${{ needs.release-please.outputs.package-sdk-server-released }}
```
kinyoklion, line 68: This was wrong, which caused the auto publish to fail.

.github/workflows/manual-publish.yml
```diff
@@ -44,6 +44,17 @@ jobs:
         with:
           workspace_name: ${{ env.WORKSPACE_NAME }}
           workspace_path: ${{ inputs.workspace_path }}
+      - uses: ./actions/release-secrets
```
kinyoklion, line 47: I forgot to add these steps to the manual publish after figuring out everything that was needed to do during automation.

Get the auth token, then configure yarn.

yusinto (approved):

PR #74, opened 2023-04-10
feat: cloudflare sdk base

yusinto:
This is the base branch for the cloudflare sdk. Please review this PR first before the following child prs:
* #75 - share MockEventSource
* #76 - adding mniflare tests
* #78 - adding ci


packages/shared/sdk-server/src/store/index.ts
yusinto, line 1: I'm exporting this so the cloudflare unit tests can use async/await when testing feature store functionality. It's much better to use async/await for unit tests than promise thenables.

packages/sdk/cloudflare/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "build": "yarn tsc",
+    "tsw": "yarn tsc --watch",
+    "start": "yarn tsw",
+    "lint": "eslint . --ext .ts",
+    "prettier": "prettier --write '**/*.@(js|ts|tsx|json|css)'",
+    "test": "jest --ci --runInBand",
+    "check": "yarn prettier && yarn lint && yarn tsc && yarn test"
+  },
+  "dependencies": {
+    "@cloudflare/workers-types": "^4.20230321.0",
```
InTheCloudDan, line 28: Are the types necessary as a `dependency` or can it be moved to `devDependency`?

packages/sdk/cloudflare/src/createLDClient/createOptions.ts
```diff
@@ -0,0 +1,67 @@
...
+import { BasicLogger, LDLogger, LDOptions, SafeLogger } from '@launchdarkly/js-server-sdk-common';
+// import { version } from '../../package.json';
+import createFeatureStore from './createFeatureStore';
+
+type SupportedLDOptions = Pick<LDOptions, 'logger' | 'featureStore'>;
+const allowedOptions = ['logger', 'featureStore'];
+
+const defaults = {
+  stream: false,
+  sendEvents: false,
```
InTheCloudDan, line 11: Are we keeping this for now because that's how it was in the original SDK? For that SDK it was set to `false` because it wasn't technically possible to polyfill the API we were using.

InTheCloudDan (commented):
Only a couple of minor questions from me.

packages/sdk/cloudflare/src/platform/eventSource.ts
```diff
@@ -0,0 +1,33 @@
+// TODO: Dry this file duplicated in this repo
```
yusinto, line 1: This duplication has been removed in #75

packages/sdk/cloudflare/package.json
yusinto: I also had the same question but eslint/typescript complained. It's actually not @types/* so I think that's why.

packages/sdk/cloudflare/src/createLDClient/createOptions.ts
yusinto: I tried `sendEvents: true` In #76 and ran miniflare tests. The tests passes but jest exited disgracefully (no offence jest):

```
Jest did not exit one second after the test run has completed.

'This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```
Is this needed for go live on Apr 28? I'm focusing on feature parity and there's still other chores outstanding like documentation, releasing and CI.

packages/sdk/cloudflare/package.json
InTheCloudDan: ahh got it, thanks for clarifying!

packages/sdk/cloudflare/src/createLDClient/createOptions.ts
InTheCloudDan: For parity no, and Tim C should probably be consulted before a default change to the behavior is implemented.

yusinto: Yep. This will be a major release though so changing sendEvents to true shouldn't be an issue. It does make sense to sendEvents and it's extremely valuable.

I will attempt to include this for go-live only if other necessary tasks are done. If not we can always release a minor version to support this. I'm sure there are many other improvements in addition to this after go-live.

packages/sdk/cloudflare/src/createLDClient/createFeatureStore.ts
```diff
@@ -0,0 +1,76 @@
...
+  LDFeatureStore,
+  LDFeatureStoreDataStorage,
+  LDFeatureStoreItem,
+  LDFeatureStoreKindData,
+} from '@launchdarkly/js-server-sdk-common';
+import noop from '../utils/noop';
+
+const createFeatureStore = (kvNamespace: KVNamespace, sdkKey: string, logger: LDLogger) => {
+  const key = `LD-Env-${sdkKey}`;
+  const store: LDFeatureStore = {
```
kinyoklion, line 14: Ah, I see, I looked through vercel first. So I had commented there.

Would it be possible for this to just be a class? `class CloudFlareFeatureStore implements LDFeatureStore`.

packages/sdk/cloudflare/src/platform/info.ts
```diff
@@ -0,0 +1,18 @@
+import type { Info, PlatformData, SdkData } from '@launchdarkly/js-server-sdk-common';
+
+import { name, version } from '../../package.json';
+
+export default class CloudflareInfo implements Info {
+  platformData(): PlatformData {
```
kinyoklion, line 6: Are there any `additional` fields that would be useful to help debug issues? I did a quick look and it doesn't look like they provide very much runtime information (and apparently they update their runtime weekly).

kinyoklion (commented):
Few comments. Overall looking very good!

packages/sdk/cloudflare/src/index.ts
```diff
@@ -0,0 +1,86 @@
...
+ *   Optional configuration settings. The only supported options for the Cloudflare SDK
+ *   are 'logger' and 'featureStore'.
+ * @return
+ *   The new {@link LDClient} instance.
+ */
+const init = (
+  kvNamespace: KVNamespace,
+  sdkKey: string,
+  options: LDOptions = {}
+): LDClientSubset => {
```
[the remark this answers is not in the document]
yusinto: > Should we re-export LDClientSubset as LDClient

Yep this is exported in #76 [here](https://github.com/launchdarkly/js-core/pull/76/files#r1165876662). I call it LDClientCloudflare though, to avoid confusion with the `full` LDClient. Is this ok with you?

packages/sdk/cloudflare/src/platform/eventSource.ts
yusinto: Ok done in #75 which deals with making it shareable.

packages/sdk/cloudflare/src/index.ts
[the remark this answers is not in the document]
kinyoklion: Yeah, I think that is probably fine.

packages/sdk/cloudflare/src/createLDClient/createFeatureStore.ts
yusinto: It is possible but I am opinionated about using constructors in javascript. I'm old school and prefer javascript functions to constructors, just because that's how the history of js developed. Constructors are ok, but they are more suggestive of OOP which js strictly speaking was not.

I understand and see other parts of the repo are designed with constructors, but perhaps we can compromise and live with both styles in this monorepo?

packages/sdk/cloudflare/src/platform/info.ts
yusinto: I also searched high and low and can't find anything useful.

packages/sdk/cloudflare/src/createLDClient/createFeatureStore.ts
kinyoklion: I can compromise.

Generally speaking I hard disagree when using typescript, and implementing an interface. But it uses structural typing, so it isn't the end of the world.

kinyoklion: (A feature store is a polymorphic type and is OOP.)

packages/sdk/cloudflare/src/platform/info.ts
kinyoklion: Cool. That is fine, I was just curious.

kinyoklion (approved):

packages/sdk/cloudflare/src/createLDClient/createFeatureStore.ts
yusinto: 🙏 Appreciate your understanding and thank you for compromising.

PR #78, opened 2023-04-12
chore: add ci config

yusinto:
Adding ci config for cloudflare.


packages/sdk/cloudflare/package.json
```diff
@@ -16,9 +16,9 @@
     "/dist"
   ],
   "scripts": {
-    "build": "yarn tsc",
+    "build": "rimraf dist && yarn tsc",
```
kinyoklion, line 19: rimraf is probably fine, but `yarn clean` does delete the output files. It doesn't delete the folders though, which may be inconvenient.

kinyoklion (approved):

PR #81, opened 2023-04-14
feat: fix typedoc and export common types

yusinto:
Typedoc does not play nice 100% so this is best-effort.

### DIFF
--- a/packages/sdk/cloudflare/src/index.ts
+++ b/packages/sdk/cloudflare/src/index.ts
@@ -44,7 +44,7 @@ export type LDClient = Pick<
 export type LDOptions = Pick<LDOptionsCommon, 'logger' | 'featureStore'>;
 
 /**
- * Creates an instance of the LaunchDarkly client.
+ * Creates an instance of the Cloudflare LaunchDarkly client.
  *
  * Applications should instantiate a single instance for the lifetime of the worker.
  * The client will begin attempting to connect to the configured Cloudflare KV as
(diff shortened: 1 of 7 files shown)


packages/sdk/cloudflare/src/index.ts
```diff
@@ -8,32 +8,50 @@
...
-import {
-  LDClient,
+import type { KVNamespace } from '@cloudflare/workers-types';
+import type {
+  LDClient as LDClientCommon,
   LDFlagsState,
   LDFlagsStateOptions,
-  LDOptions,
+  LDOptions as LDOptionsCommon,
   LDContext,
```
kinyoklion, line 17: I am confused about exports a bit. I feel like you would need to return at least these, for people to conveniently use the SDK, as well as for documentation generation to work well. Normally when you gen the docs it will say which items are present in the docs, but which there are not exported types for. The internal ones you can ignore, because we may have a field in a type that is used internally, but we don't want people to use.

	LDContext it is a parameter to variation calls.
	LDEvaluationDetail for handling the result of variationDetail.
		LDFlagValue used by LDEvaluationDetail
		LDEvaluationReason used by LDEvaluationDetail
		
	LDFlagsStateOptions as a parameter for allFlagsState.
	LDFlagsState for the return value from allFlagsState.
		LDFlagSet used by LDFlagsState

kinyoklion: There is a secondary problem with the reduced types, because their comments may reference other types that it doesn't know about. Which is possibly we can ignore that.

kinyoklion: (I missed a line so you can ignore the deleted comment.)

kinyoklion (commented):
The docs look much fuller now. :+1:

kinyoklion (approved):

PR #83, opened 2023-04-17
feat: create sdk-server-edge package

yusinto:
To enable sharing of common code between cloudflare and vercel, this PR:
* moves common code for the cloudflare and vercel sdks to a new package under packages/sdk-server-edge. 
* the cloudflare project is refactored to use this package.
* sendEvents is turned on and hanging jest tests have been fixed.

Remaining tasks which will be covered in separate PRs:
* add inline comments to fix typedoc
* add example cloudflare app

This PR should unblock the vercel team with their work. The vercel sdk should follow the cloudflare sdk requiring only the  implementation featureStore and platformInfo objects.


package.json
```diff
@@ -3,6 +3,7 @@
   "workspaces": [
     "packages/shared/common",
     "packages/shared/sdk-server",
+    "packages/shared/sdk-server-edge",
```
yusinto, line 6: Let me know if you have better name suggestions.

packages/sdk/cloudflare/package.json
```diff
@@ -11,22 +11,22 @@
   ],
   "main": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",
-  "type": "module",
   "files": [
     "/dist"
   ],
   "scripts": {
+    "doc": "../../../scripts/build-doc.sh .",
```
yusinto, line 18: This is useful for local dev.

packages/sdk/cloudflare/src/index.ts
```diff
@@ -59,44 +33,12 @@ export type LDOptions = Pick<LDOptionsCommon, 'logger' | 'featureStore'>;
...
-      callback?: (err: Error | null, res: LDFlagsState | null) => void
-    ): Promise<LDFlagsState> {
-      return client.allFlagsState(context, o, callback);
-    },
-    waitForInitialization(): Promise<LDClient> {
-      return client.waitForInitialization();
-    },
-  };
-};
+export const init = (sdkKey: string, kvNamespace: KVNamespace) =>
```
yusinto, line 39: I've simplified the public facing api to just sdkKey and the cloudflare KV. The previous version supports `options` which only supports `logger` and `featureStore` however I removed that in this new version because:

1. The BasicLogger included in server-common is completely capable of logging. I don't think it's necessary to burden the customers with a choice to override it? Unless there's a compelling reason a customer wants to specify their own loggers? Please share your knowledge if you know.
2. In the previous version, options.featureStore is redundant because the validation logic throws an error if kvNamespace is unspecified or is not valid. This tells me that options.featureStore is not really used/useful. Again please let me know if you think otherwise.

packages/shared/common/src/logging/BasicLogger.ts
```diff
@@ -31,6 +31,10 @@ export default class BasicLogger implements LDLogger {
 
   private formatter?: (...args: any[]) => string;
 
+  static get() {
+    return new BasicLogger({});
+  }
```
yusinto, line 36: This is useful to easily get a logger quickly.

packages/shared/sdk-server-edge/src/api/createOptions.ts
```diff
@@ -0,0 +1,31 @@
+import { BasicLogger, LDFeatureStore } from '@launchdarkly/js-server-sdk-common';
+
+const defaults = {
+  stream: false,
+  sendEvents: true,
```
yusinto, line 5: I fixed the hanging jest tests by closing the ldClient after all tests. So this seems to work and I'm including it.

packages/sdk/cloudflare/src/index.ts
kinyoklion: For #1 generally speaking it is good to have a way to adjust the log level. The basic logger is created at the info level I think (generally we export the basic logger so you can change the log level). The other thing is if you are using some logging framework (like winston) it is nice to be able to easily connect directly to it. Another thing, which maybe we want to change, is that the logger writes to stderr always. When it would be nice if it wrote things that were not errors to stdout. That would, generally speaking, increase the complexity of setting just a destination, because the basic logger also supports that.

We get logging questions pretty often for node.

For #2 that sounds reasonable to me.

packages/shared/common/src/logging/BasicLogger.ts
kinyoklion: We should put a comment block explaining when this should be used. Specifically it would be bad to ever use this in any SDK code, at least SDK code that lets the user set a logger.

packages/sdk/cloudflare/src/index.test.ts
```diff
@@ -50,4 +55,20 @@ describe('worker', () => {
       testFlag2: false,
     });
   });
+
+  test('isOffline is not supported', async () => {
+    expect(() => ldClient.isOffline()).toThrowError(/is not supported/);
+  });
+
+  test('flush is not supported', async () => {
```
kinyoklion, line 63: If we enable sendEvents, then likely we should support flush. Possibly right now. Flush didn't have a purpose without events, but it does have one if we do have events.

kinyoklion: I know we will want track as well, as discussed elsewhere, but what about identify?

packages/shared/sdk-server-edge/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+# Changelog
```
kinyoklion, line 1: I probably need to write a document, but when we make the first release you will have to manually edit this file, because release-please will put the release note before this notice. But once you move it to the correct place all future releases will be correct. This is just a note, note a request to change anything.

packages/shared/sdk-server-edge/src/api/LDClient.ts
```diff
@@ -0,0 +1,69 @@
...
+    throw new Error('isOffline is not supported');
+  }
+
+  override secureModeHash(): string {
+    throw new Error('secureModeHash is not supported');
+  }
+
+  override flush(): Promise<void> {
+    throw new Error('flush is not supported');
+  }
```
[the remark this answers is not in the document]
yusinto: https://launchdarkly.slack.com/archives/C01M24B06N9/p1681752269076069

packages/shared/common/src/logging/BasicLogger.ts
yusinto: Ok I added comments. Is this sufficient?

kinyoklion (approved):

kinyoklion (approved):

PR #90, opened 2023-04-20
fix: semver util import error

yusinto:
Semver [v7.5.0](https://github.com/npm/node-semver/compare/v7.4.0...v7.5.0#diff-7917f8ef42ff71df5bde71c259d15feab8b147e3e6c2fde11b56cbf42cadea16R19) adds a version check on the node util class with a `require('util')` statement which breaks nodejs compatibility in a worker environment. This should have been `require('node:util')`. The workaround is for our customers to use a bundler for the application, but that's not very nice.

So this ticket we'll freeze the semver package version to 7.4.0 which does not have `require('util')` statement.

### DIFF
--- a/packages/sdk/cloudflare/README.md
+++ b/packages/sdk/cloudflare/README.md
@@ -20,6 +20,13 @@ or yarn:
 yarn add -D @launchdarkly/cloudflare-server-sdk
 ```
 
+In `wrangler.toml` in your worker project turn on the Node.js compatibility flag.
+This allows the SDK to use `node:events`:
+
+```toml
+compatibility_flags = [ "nodejs_compat" ]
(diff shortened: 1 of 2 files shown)


kinyoklion (approved):
I wonder if we should comment in the dependencies about why it is pinned.

"//" is reserved by npm/yarn for comments.

```
"//": "The semver package is pinned because 7.5.0 introduced a bad node dependency."
```

Or something like that?

yusinto:
> "//" is reserved by npm/yarn for comments.

I didn't know this until I see you do it. TIL. Done.

```json
"//": "Pinned because 7.5.0 introduced require('util') without the node: prefix",
```

PR #98, opened 2023-04-26
fix: fix packagejson import

monsagri:
I was getting build errors complaining about package.json only being able to be used as a default import/export in ESM


ldhenry (approved):
Works for me.

PR #103, opened 2023-04-26
chore: remove beta text, add licence, homepage and repository fields …

yusinto:
* repository, homepage and licence fields are missing
* remove beta text from cloudflare readme
* minor readme improvements


release-please-config.json
```diff
@@ -1,19 +1,19 @@
 {
   "packages": {
     "packages/shared/common": {
-      "release-as": "1.0.0"
+      "bump-minor-pre-major": true
```
kinyoklion, line 4: "bump-minor-pre-major" is just to support 0.x.x releases. Otherwise it starts with 1.0.  So we don't need it anymore for these packages.

kinyoklion:
image
It would be good to add cloudflare and vercel to this table as well.

release-please-config.json
yusinto, line 1: @kinyoklion I saw your earlier comments and so I made changes here. Will this work?

kinyoklion:
If we want this release this, then we may want to call it a "fix:".

kinyoklion (approved):

yusinto:
@kinyoklion please re-review I duplicated the licence files to all packages. There maybe other smarter/more efficient ways to do this but I think it's ok to have licence files in each package directory. I also renamed LICENCE.txt to just LICENCE.

kinyoklion (approved):

PR #107, opened 2023-04-28
fix: fixed bug where the feature store does not deserialize KV values…

yusinto:
* removed { type: 'json' } from KV calls to return default string values from the KV
* expose deserializePoll and use it to deserialize string KV values
* fixed unit tests

### DIFF
--- a/packages/sdk/cloudflare/example/package.json
+++ b/packages/sdk/cloudflare/example/package.json
@@ -5,7 +5,7 @@
   "module": "./dist/index.mjs",
   "packageManager": "yarn@3.4.1",
   "dependencies": {
-    "@launchdarkly/cloudflare-server-sdk": "0.0.1"
+    "@launchdarkly/cloudflare-server-sdk": "2.0.1"
   },
   "devDependencies": {
     "@cloudflare/workers-types": "^4.20230321.0",
--- a/packages/sdk/cloudflare/package.json
+++ b/packages/sdk/cloudflare/package.json
@@ -34,6 +34,7 @@
     "lint": "eslint . --ext .ts",
     "prettier": "prettier --write '**/*.@(js|ts|tsx|json|css)' --ignore-path ../../../.prettierignore",
     "test": "NODE_OPTIONS=\"--experimental-vm-modules --no-warnings\" jest --ci --runInBand",
+    "coverage": "yarn test --coverage",
     "check": "yarn prettier && yarn lint && yarn build && yarn test && yarn doc"
   },
   "dependencies": {
--- a/packages/sdk/cloudflare/package.json
+++ b/packages/sdk/cloudflare/package.json
@@ -33,7 +33,7 @@
     "start": "rimraf dist && yarn tsw",
     "lint": "eslint . --ext .ts",
     "prettier": "prettier --write '**/*.@(js|ts|tsx|json|css)' --ignore-path ../../../.prettierignore",
-    "test": "NODE_OPTIONS=\"--experimental-vm-modules --no-warnings\" jest --ci --runInBand --coverage",
+    "test": "NODE_OPTIONS=\"--experimental-vm-modules --no-warnings\" jest --ci --runInBand",
     "check": "yarn prettier && yarn lint && yarn build && yarn test && yarn doc"
   },
   "dependencies": {
--- a/packages/sdk/cloudflare/src/createFeatureStore.test.ts
+++ b/packages/sdk/cloudflare/src/createFeatureStore.test.ts
@@ -64,10 +64,10 @@ describe('createFeatureStore', () => {
 
   describe('all', () => {
     test('all flags', async () => {
-      const flag = await asyncFeatureStore.all({ namespace: 'features' });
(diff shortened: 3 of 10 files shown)


packages/sdk/cloudflare/package.json
```diff
@@ -33,7 +33,7 @@
     "start": "rimraf dist && yarn tsw",
     "lint": "eslint . --ext .ts",
     "prettier": "prettier --write '**/*.@(js|ts|tsx|json|css)' --ignore-path ../../../.prettierignore",
-    "test": "NODE_OPTIONS=\"--experimental-vm-modules --no-warnings\" jest --ci --runInBand --coverage",
```
yusinto, line 36: Coverage is nice but noisy. I'm removing it for now.

packages/sdk/cloudflare/src/createFeatureStore.test.ts
```diff
@@ -7,6 +7,16 @@ import * as testData from './utils/testData.json';
 describe('createFeatureStore', () => {
   const sdkKey = 'sdkKey';
   const kvKey = `LD-Env-${sdkKey}`;
+  const {
+    testFlag1: { debugEventsUntilDate: d1, ...testFlag1Subset },
+    testFlag2: { debugEventsUntilDate: d2, ...testFlag2Subset },
+    testFlag3: { debugEventsUntilDate: d3, ...testFlag3Subset },
+  } = testData.flags;
```
yusinto, line 14: The deserializePoll function adds `debugEventsUntilDate` which doesn't exist in `testData.json`. So I'm omitting them here to be ignored in tests.

packages/sdk/cloudflare/src/createFeatureStore.test.ts
```diff
@@ -18,7 +28,7 @@ describe('createFeatureStore', () => {
   let asyncFeatureStore: AsyncStoreFacade;
 
   beforeEach(() => {
-    mockGet.mockImplementation(() => Promise.resolve(testData));
+    mockGet.mockImplementation(() => Promise.resolve(JSON.stringify(testData)));
```
yusinto, line 31: KV calls now return default strings.

packages/sdk/cloudflare/example/package.json
```diff
@@ -5,7 +5,7 @@
   "module": "./dist/index.mjs",
   "packageManager": "yarn@3.4.1",
   "dependencies": {
-    "@launchdarkly/cloudflare-server-sdk": "0.0.1"
+    "@launchdarkly/cloudflare-server-sdk": "2.0.1"
```
kinyoklion, line 8: Not a problem for right now, but I wonder if we can get release please to update this file as well.

packages/sdk/cloudflare/package.json
kinyoklion: I usually just make a second command "coverage". Which runs basically "yarn test -- --coverage" (or however you forward the parameter)

packages/sdk/cloudflare/src/createFeatureStore.test.ts
kinyoklion: I see it in the test data?
image

I don't see something in `deserializePoll` that would affect them.

packages/shared/sdk-server/src/store/serialization.ts
```diff
@@ -22,6 +22,7 @@ export function reviver(this: any, key: string, value: any): any {
 }
 
 interface FlagsAndSegments {
+  [key: string]: { [name: string]: Flag } | { [name: string]: Segment };
```
kinyoklion, line 25: I know you probably did this to appease typescript, but this really changes the shape of this object. Which makes it less clear what is correct code in other places. I think I have some thoughts about how to handle this at the calling side.

packages/sdk/cloudflare/example/package.json
yusinto: That will be the best. Any idea how?

packages/sdk/cloudflare/package.json
yusinto: done

packages/sdk/cloudflare/src/createFeatureStore.test.ts
yusinto: I'll investigate.

packages/sdk/cloudflare/example/package.json
kinyoklion: I do not know. I am looking into making some other build robustness changes. Maybe I will figure it out. Nothing important to solve right this second.

packages/sdk/cloudflare/src/createFeatureStore.test.ts
yusinto: I set debugEventsUntilDate to 2000 so it's non null so it's not stripped by serializer.ts.

packages/shared/sdk-server/src/store/serialization.ts
yusinto: Merged #111 which contained your suggestions. Thank you.

kinyoklion (approved):

PR #108, opened 2023-04-28
fix: vercel serialization bug

yusinto:
The same cloudflare fix in #107 is applied to vercel in this PR.

@monsagri can you please test this branch in your example app?

### DIFF
--- a/packages/sdk/vercel/src/createPlatformInfo.ts
+++ b/packages/sdk/vercel/src/createPlatformInfo.ts
@@ -1,6 +1,6 @@
 import type { Info, PlatformData, SdkData } from '@launchdarkly/js-server-sdk-common-edge';
 
-import packageJson from '../package.json';
+import { name, version } from '../package.json';
 
 class VercelPlatformInfo implements Info {
   platformData(): PlatformData {
@@ -11,8 +11,8 @@ class VercelPlatformInfo implements Info {
 
   sdkData(): SdkData {
     return {
-      name: packageJson.name,
-      version: packageJson.version,
+      name,
+      version,
     };
   }
 }
--- a/packages/sdk/vercel/src/createPlatformInfo.ts
+++ b/packages/sdk/vercel/src/createPlatformInfo.ts
@@ -1,6 +1,6 @@
 import type { Info, PlatformData, SdkData } from '@launchdarkly/js-server-sdk-common-edge';
 
-import { name, version } from '../package.json';
+import packageJson from '../package.json';
 
 class VercelPlatformInfo implements Info {
   platformData(): PlatformData {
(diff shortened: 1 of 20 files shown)


packages/sdk/vercel/src/createPlatformInfo.ts
```diff
@@ -1,6 +1,6 @@
 import type { Info, PlatformData, SdkData } from '@launchdarkly/js-server-sdk-common-edge';
 
-const packageJson = require('../package.json');
```
yusinto, line 3: @monsagri I saw on slack there was an issue with importing this. Can you expound on the issue a little more? Using `require` will break esm modules so we want to avoid that.

monsagri: Yeah, you're totally right about `require`, that was my bad yesterday.
I don't think we can use package.json fields as named exports though, so I'll change it to be a default import for now.

```

Failed to compile
../js-core/packages/sdk/vercel/dist/esm/src/createPlatformInfo.js
Can't import the named export 'name' (imported as 'name') from default-exporting module (only default export is available)
This error occurred during the build process and can only be dismissed by fixing the error.
```

packages/sdk/vercel/tsconfig.json
```diff
@@ -16,7 +16,8 @@
     "stripInternal": true,
     "moduleResolution": "node",
     "types": ["jest", "node"],
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "esModuleInterop": true,
```
monsagri, line 20: We need `esModuleInterop` to be able to import package.json as default import and be consistent with our build instructions/output

monsagri (approved):
Thanks for taking this on and being so quick with it!

I've made a few changes as discussed over slack and I think we're good to go. 

I'll follow up with vercel about updating their SDK to support returning a non parsed config string

packages/sdk/vercel/tsconfig.json
yusinto: Both the cloudflare and server-node packages don't need esModuleInterop and all the tests passed without it. Did you have an issue without it? Two syntaxes which should work:

```ts
import { name, version } from '../package.json';
//or
import * as packageJson from '../package.json';
```

Can you kindly try these please? Thank you.

monsagri: f I try and use the package with the named import, I see this error below,

```
Failed to compile
../js-core/packages/sdk/vercel/dist/esm/src/createPlatformInfo.js
Can't import the named export 'name' (imported as 'name') from default-exporting module (only default export is available)
```

This error occurred during the build process of the example project and can only be fixed by changing the import.

Using a default import works fine, however then without esModuleInterOp the tests fail as a default import from the package.json isn’t supported

he tests will pass since they use the tsconfig in the package, but have you tried using the built package in a real world setting?

yusinto: #109 has an example nextjs app which works without esModuleInterop and works with the vercel sdk with the serialization fix.

yusinto:
Closing this in favour of #114.

PR #110, opened 2023-04-28
chore: Improve release robustness.

kinyoklion:
This makes a couple improvements to the release process. 

1. Further improve the documentation publishing process.
~2. Publish packages in order.~

Cannot publish in order the way I was attempting. Because doing so caused it to only publish packages when the dependency package was published. Versus it just being satisfied in that case.

This could be done with some more complex logic.

This never happened, but imagine if common failed, and we published server anyway, now server would be broken.

So this commit adds dependencies between jobs. Overall the release process builds will take a little longer, but be safer.


kinyoklion:
Realized a problem with dependency order publishing. Working on it.

yusinto (approved):

PR #111, opened 2023-04-28
fix: Suggestions for serialization.

kinyoklion:


packages/sdk/cloudflare/src/createFeatureStore.ts
```diff
@@ -15,11 +15,11 @@ const createFeatureStore = (kvNamespace: KVNamespace, sdkKey: string, logger: LD
   const store: LDFeatureStore = {
     get(
       kind: DataKind,
-      flagKey: string,
```
kinyoklion, line 18: Isn't just flags.

packages/sdk/cloudflare/src/createFeatureStore.ts
```diff
@@ -32,7 +32,17 @@ const createFeatureStore = (kvNamespace: KVNamespace, sdkKey: string, logger: LD
           if (!item) {
             throw new Error(`Error deserializing ${kindKey}`);
           }
-          callback(item[kindKey][flagKey]);
+          switch (kind.namespace) {
```
kinyoklion, line 35: This isn't super elegant, but lets us keep the types the same, and we can make it more elegant in the next update.

packages/shared/sdk-server/src/store/serialization.ts
```diff
@@ -158,6 +157,14 @@ export function deserializeAll(data: string): AllData | undefined {
   return parsed;
 }
 
+/**
```
kinyoklion, line 160: Just because this isn't a permanent solution, and isn't a function we want a customer to use.

packages/shared/sdk-server/src/store/serialization.ts
```diff
@@ -22,7 +22,6 @@ export function reviver(this: any, key: string, value: any): any {
 }
 
 interface FlagsAndSegments {
-  [key: string]: { [name: string]: Flag } | { [name: string]: Segment };
```
kinyoklion, line 25: FlagsAndSegments isn't intended to be addressed by strings. In places where it isn't flags or segments it would generally be an error. So, this is the main thing I was concerned about.

yusinto: Maybe long term `DataKind` needs to be `flags | segments` instead of string. But this is out of scope and out of time for my ambitions now.

```ts
export interface DataKind {
  namespace: 'flags' | 'segments';
}

yusinto (approved):

PR #113, opened 2023-04-28
chore: commonize edge feature store logic

yusinto:
* Created new class EdgeFeatureStore which encapsulates shared logic for edge providers
* Refactored cloudflare to use it


packages/shared/sdk-server-edge/src/api/EdgeFeatureStore.ts
```diff
@@ -0,0 +1,120 @@
...
+export class EdgeFeatureStore implements LDFeatureStore {
+  private readonly rootKey: string;
+
+  private readonly description: string;
+
+  private edgeProvider: EdgeProvider;
+
+  private logger: LDLogger;
+
+  constructor(edgeProvider: EdgeProvider, sdkKey: string, description: string, logger: LDLogger) {
```
kinyoklion, line 24: Not requesting this as a change, but did you know you could write:
```
export class EdgeFeatureStore implements LDFeatureStore {
  constructor(private readonly edgeProvider: EdgeProvider, private readonly sdkKey: string, private description: string, private logger: LDLogger) {
```
It will automatically make all the class members if you put their specifiers in the constructor.

yusinto: That's super cool!

kinyoklion: Oh, and you don't need to assign them inside either. It will automatically do that as well.

packages/shared/sdk-server-edge/src/api/EdgeFeatureStore.ts
```diff
@@ -0,0 +1,119 @@
...
+    private readonly edgeProvider: EdgeProvider,
+    private readonly sdkKey: string,
+    private readonly description: string,
+    private logger: LDLogger
+  ) {
+    this.edgeProvider = edgeProvider;
+    this.rootKey = `LD-Env-${sdkKey}`;
+    this.description = description;
+    this.logger = logger;
+  }
```
kinyoklion, line 28:
```suggestion
  ) {
  }
```

kinyoklion, line 28:
```suggestion
  ) {
    this.rootKey = `LD-Env-${sdkKey}`;
  }
```

kinyoklion (approved):

PR #115, opened 2023-04-29
fix: bump cloudflare sdk dep

yusinto:
fix: bump cloudflare sdk dep

### DIFF
--- a/packages/sdk/cloudflare/example/package.json
+++ b/packages/sdk/cloudflare/example/package.json
@@ -5,7 +5,7 @@
   "module": "./dist/index.mjs",
   "packageManager": "yarn@3.4.1",
   "dependencies": {
-    "@launchdarkly/cloudflare-server-sdk": "2.0.1"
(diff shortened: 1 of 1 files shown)


kinyoklion:
This is a "fix" which will make another release, is that what you want?

yusinto:
For now, a patch release is fine I think. Otherwise the example app fails due to the serialization bug, which is not very good for our customers. Is that ok with you please?

kinyoklion (approved):

kinyoklion:
> For now, a patch release is fine I think. Otherwise the example app fails due to the serialization bug, which is not very good for our customers. Is that ok with you please?

I approved, but I wasn't sure if someone would be running an example from the package, or the repo. If it is just the repo, then you just need a "chore". But if you want it updated in the package, then this is the right thing to do.

yusinto:
Good point re your last comment. Preferably both scenarios should work, so the caret added should somewhat solve this kind of patch releases to a certain extent. Agree we may need to re-structure the example location.

PR #119, opened 2023-05-16
feat: Use version number from release process instead of package.json.

kinyoklion:
I ended up using the release-please extra files instead of the `.default`. The edge environment doesn't seem to agree with TS expectations here, so `.default` would violate the local typings.

The test file still loads from pacakge.json, which passes here, and should pass as the release PRs update things. If the module name changed, it would fail, which is good actually.

After this is merged we can check the release PR.


ldhenry (approved):
Thanks!

PR #123, opened 2023-05-25
feat: Add rollup config for Akamai SDK

dariushar:
Adding rollup configuration to Akamai SDK
Creating bundles for both esm and cjs
Polyfills for `node:events` and `node:timers`

### DIFF
--- a/packages/sdk/akamai/package.json
+++ b/packages/sdk/akamai/package.json
@@ -27,7 +27,7 @@
   ],
   "scripts": {
     "doc": "../../../scripts/build-doc.sh .",
-    "build": "../../../scripts/build-package.sh npm run build-ts",
+    "build": "../../../scripts/build-package.sh build-ts",
     "build-ts": "rollup -c rollup.config.js",
     "clean": "rimraf dist",
     "tsw": "yarn tsc --watch",
--- a/packages/sdk/akamai/package.json
+++ b/packages/sdk/akamai/package.json
@@ -27,7 +27,8 @@
   ],
   "scripts": {
     "doc": "../../../scripts/build-doc.sh .",
-    "build": "../../../scripts/build-package.sh",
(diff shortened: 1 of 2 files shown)


kinyoklion (changes requested):
Looks like this introduces a problem building.

packages/sdk/akamai/package.json
```diff
@@ -54,6 +55,8 @@
     "launchdarkly-js-test-helpers": "^2.2.0",
     "prettier": "^2.8.7",
     "rimraf": "^5.0.0",
+    "rollup": "^3.23.0",
+    "rollup-plugin-node-polyfills": "^0.2.1",
```
kinyoklion, line 59: Are we using node polyfills for akamai? What is needing polyfilled at this point? What is currently being polyfilled?

kinyoklion: I see. I would strongly recommend against including `node:events` at least.

kinyoklion: I know I've generally laid this out before, but js-core is designed to allow bringing you own platform. The edge SDK platform doesn't have to be used for this provider. It happens that cloudflare and vercel share the same platform constraints.

Akamai doesn't share those constraints, but has more constraints, so I think it should be using its own platform. It mostly could be empty because you don't need most of the functionality.

kinyoklion: To clarify, most code is in `shared/sdk-server` and applies to server use cases. It uses no node APIs. It does use web APIs, of which 1 will need a polyfill of some kind for akamai (setTimeout).

A platform just provides, information, network requests, crypto, etc. 

These edge providers basically just need info and crypto and can have no ops for most everything else.

dariushar:
This will be completed in a separate PR

PR #128, opened 2023-06-01
feat: Akamai SDK example, rollup config, edgeKV logic

dariushar:
This PR includes the changes for releasing the initial version of the SDK (alpha version).

We have decided to create two different SDKs. One for using the EdgeKV feature store and one allowing the user send a specific feature store they would like to use. We're doing this because Akamai is expecting a file called `edgekv_token.js` but that file won't exists if the user decides to specify their feature store. To avoid trying to dynamically load the `edgekv_token.js` file, we have decided to split the SDK into two different packages.

Akamai Updates:
- Added logic for retrieving flag data from EdgeKV
- Added tests for the Akamai classes
- Added rollup
- Added example


packages/sdk/akamai-edgekv/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+[![NPM][sdk-akamai-npm-badge]][sdk-akamai-npm-link]
+[![Actions Status][sdk-akamai-ci-badge]][sdk-akamai-ci]
+[![Documentation][sdk-akamai-ghp-badge]][sdk-akamai-ghp-link]
+[![NPM][sdk-akamai-dm-badge]][sdk-akamai-npm-link]
+[![NPM][sdk-akamai-dt-badge]][sdk-akamai-npm-link]
+
+The LaunchDarkly Akamai SDK is designed primarily for use in Akamai Edgeworkers. It follows the server-side LaunchDarkly model for multi-user contexts. It is not intended for use in desktop and embedded systems applications.
+
+For more information, see the [complete reference guide for this SDK](https://docs.launchdarkly.com/sdk/server-side/akamai).
```
mmrj, line 11:
```suggestion
For more information, see the [complete reference guide for this SDK](https://docs.launchdarkly.com/sdk/edge/akamai).
```
Docs will be in our new-ish [Edge SDKs](https://docs.launchdarkly.com/sdk/edge) section when available.

packages/sdk/akamai-edgekv/example/bundle.json
```diff
@@ -0,0 +1,4 @@
+{
+  "edgeworker-version": "__VERSION__",
```
kinyoklion, line 2: Some akamai magic?

dariushar: Not entirely sure of the purpose but I know the value is being set in the `rollup.config` on this [line](https://github.com/launchdarkly/js-core/pull/128/files/93d8b94bb29cbd3eb509f03ff51971cd71beedff#diff-05f09962aa0125ee7d10ce89fc18a361a3a49a0f8811e60036441948715f95b3R41)

@ctawiah Do you have any more info about the purpose of this?

ctawiah: This came Henry's POC but its not Akamai specific, it just uses the version from package.json to set the edgeworker version when its bundled. Its mostly useful if we're uploading the example bundle via Akamai's control center for testing and automates the manual task of bumping the version

release-please-config.json
```diff
@@ -12,6 +12,20 @@
         "src/createPlatformInfo.ts"
       ],
       "bump-minor-pre-major": true
+    },
+    "packages/sdk/akamai": {
```
ldhenry, line 16: I assume most customers will want to use the `akamai-edgekv` version instead of this one. It may be misleading for `akamai` to be something that people should ignore. Maybe we could call it `akamai-base` instead. What do you think?

dariushar: I like that idea. I can make those updates after the All Tech meeting.

kinyoklion (approved):

ctawiah (approved):
🎉

PR #136, opened 2023-06-05
chore: Scaffold node redis store package.

kinyoklion:
First note that this is NOT targeting main. It is targeting a feature branch.

Scaffolds the location for the node redis package. This just makes packages and updates readmes and things.

This does not add it to the release-please configuration. That will be done when it is ready to be released.

The existing package uses the "redis" package. For this implementation I am planning to use "ioredis" because it has sentinel support.

### DIFF
--- a/packages/store/node-server-sdk-redis/README.md
+++ b/packages/store/node-server-sdk-redis/README.md
@@ -60,7 +60,9 @@ By default, the store will try to connect to a local Redis instance on port 6379
 
 To reduce traffic to Redis, there is an optional in-memory cache that retains the last known data for a configurable amount of time. This is on by default; to turn it off (and guarantee that the latest feature flag data will always be retrieved from Redis for every flag evaluation), configure the store as follows:
 
-        const store = RedisFeatureStore({ cacheTTL: 0 });
+```typescriot
+const store = new RedisFeatureStore({ cacheTTL: 0 });
+```
 
 ## Contributing
 
--- a/packages/store/node-server-sdk-redis/README.md
+++ b/packages/store/node-server-sdk-redis/README.md
@@ -1,0 +1,87 @@
+# LaunchDarkly Server-Side SDK for Node.js
+
+[![NPM][node-redis-npm-badge]][node-redis-npm-link]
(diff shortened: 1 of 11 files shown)


packages/store/node-server-sdk-redis/README.md
```diff
@@ -0,0 +1,87 @@
+# LaunchDarkly Server-Side SDK for Node.js
+
+[![NPM][node-redis-npm-badge]][node-redis-npm-link]
+[![Actions Status][node-redis-ci-badge]][node-redis-ci]
+[![Documentation](https://img.shields.io/static/v1?label=GitHub+Pages&message=API+reference&color=00add8)](https://launchdarkly.github.io/js-core/packages/store/node-server-sdk-redis/docs/)
+
+This library provides a Redis-backed persistence mechanism (feature store) for the [LaunchDarkly Node.js SDK](https://github.com/launchdarkly/js-core/packages/sdk/server-node), replacing the default in-memory feature store. The underlying Redis client implementation is [ioredis](https://github.com/luin/ioredis).
+
+The minimum version of the LaunchDarkly Server-Side SDK for Node for use with this library is 8.0.0.
```
kinyoklion, line 9: I have the 8.0.0 text here, because that is what node server will be when it releases. Depending on how this comes together I may need to add a clarifying node.

packages/store/node-server-sdk-redis/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+This package is compatible with Node.js versions 14 and above.
+
+## Getting started
+
+Refer to [Using Redis as a persistent feature store](https://docs.launchdarkly.com/sdk/features/storing-data/redis#nodejs-server-side).
+
+## Quick setup
+
+This assumes that you have already installed the LaunchDarkly Node.js SDK.
```
kinyoklion, line 29: This is an estimate of what things may look like. The old readme had a section like this.

louis-launchdarkly (approved):
This looks like a good start - I am less equipped to tell if there is other things we want to add, but given this is a working feature branch, we can always make more changes.

PR #141, opened 2023-06-07
chore: rename Edge Config client

dferber90:
Renames the edgeClient variable to edgeConfigClient to be more descriptive and avoid confusion.


ldhenry (approved):
Thanks!

PR #144, opened 2023-06-07
fix: Export PersistentDataStoreWrapper.

kinyoklion:
fix: Fix a bug in the UpdateQueue, used for persistence, which would prevent tasks from executing correctly.
fix: Update package.json to use --topological-dev when building workspaces. This will ensure packages which have devDependencies on each other, are built in correct order. Currently only "dependencies" affect build order. This will affect store packages because they use a peer+dev dependency on the SDK they are associated with.

### DIFF
--- a/packages/shared/sdk-server/src/store/PersistentDataStoreWrapper.ts
+++ b/packages/shared/sdk-server/src/store/PersistentDataStoreWrapper.ts
@@ -86,8 +86,6 @@ function deserialize(
  * instance of {@link PersistentDataStore}. It provides optional caching behavior and other logic
  * that would otherwise be repeated in every data store implementation. This makes it easier to
  * create new database integrations by implementing only the database-specific logic.
- *
- * @internal
  */
 export default class PersistentDataStoreWrapper implements LDFeatureStore {
   private isInitialized = false;
--- a/packages/shared/sdk-server/src/store/UpdateQueue.ts
+++ b/packages/shared/sdk-server/src/store/UpdateQueue.ts
@@ -22,12 +22,13 @@ export default class UpdateQueue {
         // count could hit 0, and overlapping execution chains could be started.
         this.queue.shift();
         // There is more work to do, so schedule an update.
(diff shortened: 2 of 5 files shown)


packages/shared/sdk-server/src/store/PersistentDataStoreWrapper.ts
```diff
@@ -86,8 +86,6 @@ function deserialize(
  * instance of {@link PersistentDataStore}. It provides optional caching behavior and other logic
  * that would otherwise be repeated in every data store implementation. This makes it easier to
  * create new database integrations by implementing only the database-specific logic.
- *
- * @internal
```
kinyoklion, line 90: Needs to be used by persistent store packages.

packages/shared/sdk-server/src/store/UpdateQueue.ts
```diff
@@ -22,12 +22,13 @@ export default class UpdateQueue {
         // count could hit 0, and overlapping execution chains could be started.
         this.queue.shift();
         // There is more work to do, so schedule an update.
-        if (this.enqueue.length > 0) {
```
kinyoklion, line 25: This was a bug, `enqueue` is a function. The function has 2 parameters, so its length is 2. Which caused the queue to always trigger a task update. 

This caused multiple execution of tasks, as well as failing to dispatch other tasks.

packages/shared/sdk-server/src/store/index.ts
```diff
@@ -1,4 +1,5 @@
 import AsyncStoreFacade from './AsyncStoreFacade';
+import PersistentDataStoreWrapper from './PersistentDataStoreWrapper';
 import { deserializePoll } from './serialization';
 
-export { AsyncStoreFacade, deserializePoll };
+export { AsyncStoreFacade, PersistentDataStoreWrapper, deserializePoll };
```
kinyoklion, line 5: Exporting to be used in persistent store packages.

louis-launchdarkly (approved):

PR #150, opened 2023-06-08
ci: update build docs script

dariushar:
Due to an issue with rollup and typedoc, the Akamai EdgeKV package and the Akamai Base package didn't generate docs correctly. To fix this:

- I updated the ci action to use the `doc` command found in the workspace instead of always using the build-docs script.
- Added the `doc` script to the packages that didn't have the script listed


kinyoklion (approved):

ctawiah (approved):
👍🏽

PR #151, opened 2023-06-08
test: Akamai base and Akamai edgekv tests

dariushar:
Adding tests for Akamai base and Akamai edgeKV index files

Ran into this error when running the tests so I've added `moduleNameMapper` to the `jest.config.json` files
Screenshot 2023-06-08 at 10 26 12 AM


packages/sdk/akamai-base/jest.config.json
```diff
@@ -5,5 +5,8 @@
   "modulePathIgnorePatterns": ["dist"],
   "testEnvironment": "node",
   "moduleFileExtensions": ["ts", "tsx", "js", "jsx", "json", "node"],
-  "collectCoverageFrom": ["src/**/*.ts"]
+  "collectCoverageFrom": ["src/**/*.ts"],
+  "moduleNameMapper": {
```
kinyoklion, line 9: I've not experienced this on other packages. Do you know the underlying reason you need this?

dariushar: I'm not entirely sure. My assumption is that it has something to do with the rollup config since none of the other sdks are having this issue. I can try to investigate further tomorrow.

ctawiah: hmm not sure this is the case either, my 2 cents is that is probably related to the `module` value set in tsconfig.json because we need this package's output in only ES module format so perhaps thats causing the jest import resolution issue.

kinyoklion (approved):

PR #160, opened 2023-06-12
fix: Correctly handle excluded big segments.

kinyoklion:
fix: Correctly handle cached big segment status for excluded users.
fix: Handle log formatting correctly when there is a prefix. 
chore: Implement support for big segments contract tests.
fix: Do not query big segment status for a context which doesn't match the big segment kind. Verify the kind before checking the generation.

### DIFF
--- a/contract-tests/BigSegmentTestStore.js
+++ b/contract-tests/BigSegmentTestStore.js
@@ -1,0 +1,29 @@
+import got from 'got';
+
+export default class BigSegmentTestStore {
+  /**
+   * Create a big segment test store suitable for use with the contract tests.
+   * @param {string} callbackUri Uri on the test service to direct big segments
+   * calls to.
+   */
+  constructor(callbackUri) {
+    this._callbackUri = callbackUri;
+  }
+
+  async getMetadata() {
+    const data = await got.get(`${this._callbackUri}/getMetadata`, { retry: { limit: 0 } }).json();
+    return data;
+  }
+
+  async getUserMembership(contextHash) {
+    const data = await got.post(`${this._callbackUri}/getMembership`, {
+      retry: { limit: 0 },
+      json: {
+        contextHash
+      }
+    }).json();
+    return data?.values;
+  }
+
+  close() { }
+}
--- a/contract-tests/index.js
+++ b/contract-tests/index.js
@@ -1,8 +1,8 @@
-const express = require('express');
-const bodyParser = require('body-parser');
+import express from 'express';
+import bodyParser from 'body-parser';
 
-const { Log } = require('./log');
-const { newSdkClientEntity, badCommandError } = require('./sdkClientEntity');
+import { Log } from './log.js';
+import { newSdkClientEntity, badCommandError } from './sdkClientEntity.js';
 
 const app = express();
 let server = null;
@@ -26,6 +26,7 @@ app.get('/', (req, res) => {
       'all-flags-details-only-for-tracked-flags',
       'all-flags-with-reasons',
       'tags',
+      'big-segments'
     ],
   });
 });
--- a/contract-tests/sdkClientEntity.js
+++ b/contract-tests/sdkClientEntity.js
@@ -1,10 +1,12 @@
-const ld = require('node-server-sdk');
+import ld from 'node-server-sdk';
+import BigSegmentTestStore from './BigSegmentTestStore.js';
 
-const { Log, sdkLogger } = require('./log');
+import { Log, sdkLogger } from './log.js';
 
 const badCommandError = new Error('unsupported command');
+export { badCommandError };
 
-function makeSdkConfig(options, tag) {
+export function makeSdkConfig(options, tag) {
   const cf = {
     logger: sdkLogger(tag),
   };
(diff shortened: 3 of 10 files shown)


contract-tests/BigSegmentTestStore.js
```diff
@@ -0,0 +1,29 @@
+import got from 'got';
```
kinyoklion, line 1: This is a big segment store implementation for the contract tests.
I used 'got' here because it is just a test dependency and it made this simple. We can use anything though if this becomes a problem.

contract-tests/index.js
```diff
@@ -1,8 +1,8 @@
-const express = require('express');
```
kinyoklion, line 1: I wanted to use an ESM (got), so I made my life easier and switched to imports. The contract tests are still JS, but I think we are now beyond the need to as a compatibility check. So we could change them to TS at some point when we have time.

contract-tests/sdkClientEntity.js
```diff
@@ -33,17 +35,31 @@ function makeSdkConfig(options, tag) {
       version: options.tags.applicationVersion,
     };
   }
+  if (options.bigSegments) {
```
kinyoklion, line 38: Support bigSegment options.

contract-tests/sdkClientEntity.js
```diff
@@ -106,7 +122,7 @@ async function newSdkClientEntity(options) {
         return undefined;
 
       case 'getBigSegmentStoreStatus':
-        return undefined;
+        return await client.bigSegmentStoreStatusProvider.requireStatus();
```
kinyoklion, line 125: Support querying big segments status for contract tests.

packages/shared/common/__tests__/logging/BasicLogger.test.ts
```diff
@@ -48,11 +48,15 @@ describe('given a logger with a custom name', () => {
     logger.info('b');
     logger.warn('c');
     logger.error('d');
+    logger.debug('This %s is %s', 'log', 'working');
```
kinyoklion, line 51: There was a bug in the logger with formatting with a prefix. So I added some test coverage for that.

packages/shared/sdk-server/src/BigSegmentsManager.ts
```diff
@@ -9,6 +9,10 @@ const DEFAULT_STATUS_POLL_INTERVAL_SECONDS = 5;
 const DEFAULT_USER_CACHE_SIZE = 1000;
 const DEFAULT_USER_CACHE_TIME_SECONDS = 5;
 
+interface MembershipCacheItem {
```
kinyoklion, line 12: There was a bug with caching the membership for a context which had no membership. That membership is null/undefined, so caching undefined provided no difference between there having not been status fetched and there not being any status.

packages/shared/sdk-server/src/evaluation/Evaluator.ts
```diff
@@ -465,9 +465,11 @@ export default class Evaluator {
     state: EvalState,
     segmentsVisited: string[]
   ): Promise<MatchOrError> {
-    const includeExclude = matchSegmentTargets(segment, context);
-    if (includeExclude !== undefined) {
-      return new Match(includeExclude);
+    if (!segment.unbounded) {
```
kinyoklion, line 468: After checking an unbounded segment membership we then need to allow checking to continue, but we don't need to evaluate the includes/excludes.

packages/shared/sdk-server/src/evaluation/Evaluator.ts
```diff
@@ -501,6 +503,13 @@ export default class Evaluator {
       return this.simpleSegmentMatchContext(segment, context, state, segmentsVisited);
     }
 
+    const bigSegmentKind = segment.unboundedContextKind || 'user';
```
kinyoklion, line 506: I moved this earlier, but really the order here should not matter.

packages/sdk/server-node/src/index.ts
```diff
@@ -13,10 +13,10 @@ import {
   BasicLoggerOptions,
   LDLogger,
   LDOptions,
-  LDClient,
 } from '@launchdarkly/js-server-sdk-common';
 import { EventEmitter } from 'events';
 import LDClientImpl from './LDClientNode';
+import { LDClient } from './api/LDClient';
```
kinyoklion, line 19: This type correctly specifies the node client type, which includes a big segment status provider. Versus the LDClient & Emitter, which is lacking the big segment status provider.

contract-tests/index.js
```diff
@@ -26,6 +26,7 @@ app.get('/', (req, res) => {
       'all-flags-details-only-for-tracked-flags',
       'all-flags-with-reasons',
       'tags',
+      'big-segments'
```
kinyoklion, line 29: Add the big segments capability. 

This doesn't test stores, but it does test common logic/evaluation/caching.

packages/shared/sdk-server/src/evaluation/Evaluator.ts
```diff
@@ -465,9 +465,11 @@ export default class Evaluator {
     state: EvalState,
     segmentsVisited: string[]
   ): Promise<MatchOrError> {
-    const includeExclude = matchSegmentTargets(segment, context);
-    if (includeExclude !== undefined) {
-      return new Match(includeExclude);
+    if (!segment.unbounded) {
+      const includeExclude = matchSegmentTargets(segment, context);
+      if (includeExclude !== undefined) {
```
yusinto, line 470:
```suggestion
      if (includeExclude) {
```

packages/shared/sdk-server/src/evaluation/Evaluator.ts
```diff
@@ -501,6 +503,13 @@ export default class Evaluator {
       return this.simpleSegmentMatchContext(segment, context, state, segmentsVisited);
     }
 
+    const bigSegmentKind = segment.unboundedContextKind || 'user';
+    const keyForBigSegment = context.key(bigSegmentKind);
+
+    if (keyForBigSegment === undefined) {
```
yusinto, line 509:
```suggestion
    if (!keyForBigSegment) {
```

kinyoklion: I think this one would be fine, as empty keys are not valid.

yusinto (approved):

packages/shared/sdk-server/src/evaluation/Evaluator.ts
kinyoklion: It may be good to change this to an enum like type in the future.

PR #162, opened 2023-06-13
feat: Allow imports from paths.

kinyoklion:


packages/sdk/server-node/package.json
```diff
@@ -10,6 +10,21 @@
   "type": "commonjs",
   "main": "./dist/src/index.js",
   "types": "./dist/src/index.d.ts",
+  "exports": {
```
kinyoklion, line 13: This seems to make everything happy. I tried require and import in JS with node, and I tried import in TS with node. Everything worked and showed types. I needed the "typesVersions" before vscode intellisense worked correctly for integrations.

I have not done this much, so I am not a domain expert here.

`import { TestData } from '@launchdarkly/node-server-sdk/integrations';`

I am still exporting `integrations` from the main package. Probably safe for now.

PR #167, opened 2023-06-14
chore: Workaround x-release-please-version problem.

kinyoklion:
Workaround for https://github.com/googleapis/release-please/issues/1978

Basically this just runs a post processing step to put the version number into the build.

So we build as normal, then it replaces `__LD_VERSION__` in the built package with the version from the package.json.

Once the release-please bug is fixed we can remove this workaround.

### DIFF
--- a/packages/sdk/vercel/src/createPlatformInfo.test.ts
+++ b/packages/sdk/vercel/src/createPlatformInfo.test.ts
@@ -14,9 +14,6 @@ describe('Vercel Platform Info', () => {
   it('sdkData shows correct information', () => {
     const platformData = createPlatformInfo();
 
-    expect(platformData.sdkData()).toEqual({
-      name: packageJson.name,
-      version: packageJson.version,
-    });
+    expect(platformData.sdkData().name).toEqual(packageJson.name);
   });
 });
--- a/scripts/replace-version.sh
+++ b/scripts/replace-version.sh
@@ -1,0 +1,14 @@
+# Run this script like:
+# ./scripts/replace-version.sh packages/sdk/node
+
+# It will look for the string __LD_VERSION__ in the `dist` directory and replace 
+# any instances with the version from the package.json.
+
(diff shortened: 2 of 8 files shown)


packages/sdk/vercel/src/createPlatformInfo.test.ts
```diff
@@ -14,9 +14,6 @@ describe('Vercel Platform Info', () => {
   it('sdkData shows correct information', () => {
     const platformData = createPlatformInfo();
 
-    expect(platformData.sdkData()).toEqual({
-      name: packageJson.name,
-      version: packageJson.version,
-    });
+    expect(platformData.sdkData().name).toEqual(packageJson.name);
```
kinyoklion, line 17: Cannot test the version here because it isn't processed in the jest build.

scripts/replace-version.sh
```diff
@@ -0,0 +1,14 @@
...
+
+# It will look for the string __LD_VERSION__ in the `dist` directory and replace 
+# any instances with the version from the package.json.
+
+# This is a workaround for a bug with the node-workspace plugin with release-please.
+# https://github.com/googleapis/release-please/issues/1978
+
+set -e
+
+version=$(node -p "let pj = require('./$1/package.json');\`\${pj.version}\`");
```
kinyoklion, line 12: Get the version from the package.json in the specified directory.

scripts/replace-version.sh
```diff
@@ -0,0 +1,14 @@
...
+
+find "$1/dist" -type f -exec sed -i "s/__LD_VERSION__/$version/g" {} +
```
kinyoklion, line 14: Find `__LD_VERSION__` in any files in `/dist` and replace it with the actual version.

ctawiah (approved):
Works for me

dariushar (approved):

PR #171, opened 2023-06-15
chore: Remove release-as for redis/dynamo.

kinyoklion:
chore: Run Redis/Dynamo during releases to allow redis/dynamo tests to pass.

### DIFF
--- a/.github/workflows/manual-publish.yml
+++ b/.github/workflows/manual-publish.yml
@@ -41,6 +41,18 @@ jobs:
         with:
           node-version: 16.x
           registry-url: 'https://registry.npmjs.org'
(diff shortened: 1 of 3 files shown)


.github/workflows/manual-publish.yml
```diff
@@ -41,6 +41,18 @@ jobs:
         with:
           node-version: 16.x
           registry-url: 'https://registry.npmjs.org'
+      - name: "Setup Redis"
```
kinyoklion, line 44: When running either redis or dynamo releases start the servers. I need to refactor this.

Long term I need a better way to compose CI setup for releases.

yusinto (approved):

PR #174, opened 2023-06-16
fix: Fix typescript exports for both cjs and es modules

ctawiah:
- Resolved an issue that caused typescript types import to fail for ES modules
- Improved Akamai SDK usability as a hybrid module.


dariushar (approved):
lgtm

PR #177, opened 2023-06-20
feat: Vercel 1.0.0

monsagri:
Intending to use release-now to release 1.0.0 for GA today, [judging from their docs](https://github.com/googleapis/release-please#how-do-i-change-the-version-number) having that message above in the merge commit should do it?


yusinto:
@monsagri Have you followed the initial package publish steps [here](https://github.com/launchdarkly/js-core/blob/main/contributing/initial-package-publish.md)?

You can also use the cloudflare [initial release PR](https://github.com/launchdarkly/js-core/pull/101/files) for reference.

monsagri:
> @monsagri Have you followed the initial package publish steps [here](https://github.com/launchdarkly/js-core/blob/main/contributing/initial-package-publish.md)?
> 
> You can also use the cloudflare [initial release PR](https://github.com/launchdarkly/js-core/pull/101/files) for reference.

Thanks for the pointers, I've updated to follow that process!
I wondered if we could avoid the hardcoded release-as with the commit message as referenced in the docs.

packages/sdk/vercel/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@launchdarkly/vercel-server-sdk",
-  "version": "0.4.3",
```
kinyoklion, line 3: I would be very careful manually updating these. It should be ok as long as there are release notes. Release please has problems if it doesn't make any changes in a project when it releases it. Because then it cannot find the release PR associated with the project. 

So, generally you always have a change because it updates the version and changes there changelog. So make sure at least one of those will still be true.

release-please-config.json
```diff
@@ -14,7 +14,8 @@
       "extra-files": [
         "src/createPlatformInfo.ts"
       ],
-      "bump-minor-pre-major": true
+      "bump-minor-pre-major": true,
```
kinyoklion, line 17: 1.0.0 will be "major" so you can remove "bump-minor-pre-major" it just served to stop it going 1.0.0 automatically. (If you removed it, and had feat changes, then it would update to 1.0.0 on its own).

kinyoklion:
Do you have a beta notice you can remove?

This PR isn't going to make a release. There aren't any releasable units for the SDK.

kinyoklion:
Explanation of releasable units: https://launchdarkly.atlassian.net/wiki/spaces/PD/pages/2425487795/Conventional+Commits+with+Release-Please

I would at least change the commit message of this PR, when you merge it, to "feat: Vercel 1.0.0" or something like that.

kinyoklion (approved):
I am approving. I changed the PR title to something releasable. This contains changes in your example repo, so it will release them.

Note, I think I mentioned this before, but examples are not going to get automatic version bumps, which may be inconvenient later. It is fine for this release though.

PR #189, opened 2023-06-29
chore: Add support for user-type contract tests.

kinyoklion:
This adds support for the `user-type` capability of the contract tests.

All tests pass.

### DIFF
--- a/contract-tests/sdkClientEntity.js
+++ b/contract-tests/sdkClientEntity.js
@@ -90,9 +90,9 @@ export async function newSdkClientEntity(options) {
       case 'evaluate': {
         const pe = params.evaluate;
         if (pe.detail) {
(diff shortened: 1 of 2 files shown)


contract-tests/sdkClientEntity.js
```diff
@@ -90,9 +90,9 @@ export async function newSdkClientEntity(options) {
       case 'evaluate': {
         const pe = params.evaluate;
         if (pe.detail) {
-          return await client.variationDetail(pe.flagKey, pe.context, pe.defaultValue);
+          return await client.variationDetail(pe.flagKey, pe.context || pe.user, pe.defaultValue);
```
kinyoklion, line 93: Basically for any request it can pass a user or a context, we handle them interchangeably.

PR #190, opened 2023-07-03
fix: move "default" conditional exports to be last

seanparmelee:
**Requirements**

I was trying to upgrade our Next.js application to use `@launchdarkly/node-server-sdk`, but was running into the following error when trying to build our app:

```
Module not found: Default condition should be last one
```

According to https://nodejs.org/api/packages.html#conditional-exports, the `default` conditional export should be the last one listed. After making that change the app built successfully. I created a simple reproduction here: https://github.com/seanparmelee/ld-server-nextjs.

### DIFF
--- a/packages/sdk/server-node/package.json
+++ b/packages/sdk/server-node/package.json
@@ -12,12 +12,12 @@
   "types": "./dist/src/index.d.ts",
   "exports": {
     ".": {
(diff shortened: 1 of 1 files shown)


kinyoklion:
@seanparmelee,

Thank you for finding and fixing this!

kinyoklion (approved):

PR #191, opened 2023-07-04
fix: adhere to specs by making 'default' export the last one

Katona:
**Requirements**

**Related issues**

This PR makes the `default` entry within `exports` sections of `package.json` the last one as mandated by the [specs](https://nodejs.org/api/packages.html#packages_conditional_exports). This inconsistency is actually causing error when building apps relying on `@launchdarkly/node-server-sdk` with Next.js (which uses Webpack internally):

```
Module not found: Default condition should be last one
```

I have moved the `default` entry in the `exports` section so it has become the last one.

I haven't considered any.

**Additional context**

It appears to effect Next.js (and possibly any other project using Webpack), but plain Node.js project without bundler has no problem, maybe Node.js is not that rigorous regarding this rule.

### DIFF
--- a/packages/sdk/server-node/package.json
+++ b/packages/sdk/server-node/package.json
@@ -12,12 +12,12 @@
   "types": "./dist/src/index.d.ts",
   "exports": {
     ".": {
-      "default": "./dist/src/index.js",
-      "types": "./dist/src/index.d.ts"
+      "types": "./dist/src/index.d.ts",
+      "default": "./dist/src/index.js"
     },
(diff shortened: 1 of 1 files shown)


kinyoklion:
@Katona, 

Thank you for the contribution.

This is also addressed in: https://github.com/launchdarkly/js-core/pull/190

By another contributor.

Thank you,
Ryan

kinyoklion:
Closing because an equivalent PR has been merged. Again thank you for your contribution!

PR #193, opened 2023-07-05
chore: Fix timing issues for flaky tests.

kinyoklion:


packages/shared/sdk-server/__tests__/LDClient.evaluation.test.ts
```diff
@@ -158,6 +158,17 @@ describe('given an offline client', () => {
   });
 });
 
+class InertUpdateProcessor implements LDStreamProcessor {
```
kinyoklion, line 161: Update processor that never finishes initialization.

PR #194, opened 2023-07-05
fix: Preserve fallthrough variation when cloning test data.

kinyoklion:


packages/shared/sdk-server/__tests__/LDClient.evaluation.test.ts
```diff
@@ -34,6 +34,33 @@ describe('given an LDClient with test data', () => {
     client.close();
   });
 
+  it('evaluates a flag which has a fallthrough and a rule', async () => {
```
kinyoklion, line 37: This is a clone of the test case the user encountered.

packages/shared/sdk-server/__tests__/integrations/test_data/TestData.test.ts
```diff
@@ -199,7 +202,14 @@ describe('given a TestData instance', () => {
       },
     ];
 
-    expect(flagCopy.build(1).rules).toEqual(flagRules);
+    const builtFlag = flagCopy.build(1);
+    expect(builtFlag.fallthrough).toEqual({ variation: 1 });
```
kinyoklion, line 206: This test only verified rules before. Now it covers the basics as well.

louis-launchdarkly (approved):

PR #205, opened 2023-07-17
chore: bump semver dep

yusinto:
Updating semver to latest. The reason it was pinned to 7.5.0 was because 7.5.0 introduced require('util') without the node: prefix. This has been fixed in [7.5.1](https://github.com/npm/node-semver/releases/tag/v7.5.1) so we'll update to latest which is 7.5.4.

### DIFF
--- a/packages/shared/sdk-server/package.json
+++ b/packages/shared/sdk-server/package.json
@@ -27,10 +27,9 @@
     "lint:fix": "yarn run lint -- --fix"
   },
   "license": "Apache-2.0",
-  "//": "Pinned semver because 7.5.0 introduced require('util') without the node: prefix",
   "dependencies": {
     "@launchdarkly/js-sdk-common": "1.0.2",
-    "semver": "7.4.0"
+    "semver": "7.5.4"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


kinyoklion:
What confidence level do we have that this will execute nicely in cloudflare, vercel, akamai? Are the current unit tests sufficiently representative of those environments?

@dariushar @ctawiah @monsagri

louis-launchdarkly (dismissed):

kinyoklion (changes requested):
Requesting changes until my question is addressed.

dariushar (approved):
Someone else may be a little more confident than I am but I took a quick look at the Akamai and Vervel SDKs and I don't see any issues being caused by this

ctawiah:
> What confidence level do we have that this will execute nicely in cloudflare, vercel, akamai? Are the current unit tests sufficiently representative of those environments?
> 
> @dariushar @ctawiah @monsagri

LGTM

ctawiah (approved):

monsagri (approved):

monsagri:
Looks good to me from what I can see!

kinyoklion (approved):

PR #207, opened 2023-07-18
chore: improved LDEmitter

yusinto:
* Improved emitter design
* Improved jest config
* Added unit tests

### DIFF
--- a/packages/shared/sdk-client/jest.config.js
+++ b/packages/shared/sdk-client/jest.config.js
@@ -1,7 +1,0 @@
-module.exports = {
-  transform: { '^.+\\.ts?$': 'ts-jest' },
-  testMatch: ['**/__tests__/**/*test.ts?(x)'],
-  testEnvironment: 'node',
-  moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json', 'node'],
-  collectCoverageFrom: ['src/**/*.ts'],
-};
--- a/packages/shared/sdk-client/src/api/LDEmitter.ts
+++ b/packages/shared/sdk-client/src/api/LDEmitter.ts
@@ -38,7 +38,10 @@ export default class LDEmitter {
   }
 
   off(name: EventName) {
-    this.listeners.get(name)?.forEach((l) => this.et.removeEventListener(name, l));
+    this.listeners.get(name)?.forEach((l) => {
+      this.et.removeEventListener(name, l);
+    });
+    this.listeners.delete(name);
   }
 
   emit(name: EventName, ...detail: any[]) {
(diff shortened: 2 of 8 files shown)


packages/shared/sdk-client/jest.config.js
yusinto, line 1: Renamed to jest.config.json to be consistent with other projects in the repo.

packages/shared/sdk-client/src/api/LDEmitter.ts
yusinto, line 1: I'm omitting `maybeReportError` from the legacy api for now. I don't understand why anyone will trust a method like `maybeXXX`. Secondly I don't understand why maybe reporting an error is the responsibility of an Emitter.

packages/shared/sdk-client/src/api/LDEmitter.ts
```diff
@@ -1,20 +1,21 @@
...
+ * This is an event emitter using the standard built-in EventTarget web api.
  * https://developer.mozilla.org/en-US/docs/Web/API/EventTarget
  *
  * In react-native use event-target-shim to polyfill EventTarget. This is safe
  * because the react-native repo uses it too.
  * https://github.com/mysticatea/event-target-shim
  */
-export default class LDEventTarget extends EventTarget {
+export default class LDEmitter {
+  private et: EventTarget = new EventTarget();
```
yusinto, line 12: I choose not to extend because we are designing this to follow EventEmitter's interface. Further, for react-native, we may need to refactor `EventTarget` to be a generic so we can swap it out with a polyfill. For example:

```ts
// react-native will pass in a polyfill.
// browsers will use the built-in api.
export default class LDEmitter<IEventTarget> {
```

packages/shared/sdk-client/src/api/LDEmitter.ts
```diff
@@ -25,24 +26,30 @@ export default class LDEventTarget extends EventTarget {
...
-      listener(...listener.arguments, ...detail);
+      // invoke listener with args from CustomEvent
+      listener(...detail);
     };
     this.saveListener(name, customListener);
+    this.et.addEventListener(name, customListener);
+  }
+
+  off(name: EventName) {
+    this.listeners.get(name)?.forEach((l) => this.et.removeEventListener(name, l));
```
kinyoklion, line 41: It feels like this should be removing those listeners from our local listener list as well?

kinyoklion (changes requested):

packages/shared/sdk-client/src/api/LDEmitter.ts
yusinto: Thanks!

packages/shared/sdk-client/src/api/LDEmitter.ts
yusinto: More for ease of porting. However, I want to know your thoughts. Please expound.

kinyoklion: If it makes porting easier, then I am fine with it. I just wanted to make sure there wasn't some technical reason we specifically wanted to keep that interface.

kinyoklion (approved):

PR #208, opened 2023-07-19
chore: major prettier update

yusinto:
This PR only contains config changes. The post-prettier-ran after effects are in #209 where all the CI checks are green.

* Updated prettier to v3.0
* Added `prettier-plugin-sort-imports` to sort imports

The main files to review are the root `.prettierrc` and root `package.json`. The other package.json updates are just mirrors of the root.


.prettierrc
```diff
@@ -1,4 +1,10 @@
 {
   "singleQuote": true,
-  "printWidth": 100
+  "printWidth": 100,
+  "semi": false,
+  "plugins": ["@trivago/prettier-plugin-sort-imports"],
+  "importOrder": ["^@launchdarkly/(.*)$", "^[./]"],
+  "importOrderCaseInsensitive": true,
+  "importOrderSeparation": true,
+  "importOrderSortSpecifiers": true
```
yusinto, line 9: I got rid of semicolons in other projects and I don't miss them. Let me know if you think they are necessary, I'm more than happy to negotiate.

Sorting imports make our code nicer and it's already part of gonfalon. We are using the prettier plugin here which is better and more modern, compared to the eslint plugin in gonfalon.

package.json
```diff
@@ -29,20 +29,22 @@
     "contract-test-service": "npm --prefix contract-tests install && npm --prefix contract-tests start",
     "contract-test-harness": "curl -s https://raw.githubusercontent.com/launchdarkly/sdk-test-harness/master/downloader/run.sh \\ | VERSION=v2 PARAMS=\"-url http://localhost:8000 -debug -stop-service-at-end $TEST_HARNESS_PARAMS\" sh",
     "contract-tests": "npm run contract-test-service & npm run contract-test-harness",
-    "prettier": "npx prettier --write \"**/*.{js,ts,tsx,json,yaml,yml,md}\" --loglevel warn"
+    "prettier": "npx prettier --write \"**/*.{js,ts,tsx,json,yaml,yml,md}\" --log-level warn",
+    "check": "yarn && yarn prettier && yarn lint && tsc && yarn build"
```
yusinto, line 33: Added `check` as a quick way to validate everything still works. I included tsc to validate typescript and its json config files.

package.json
```diff
@@ -29,20 +29,22 @@
...
-    "eslint-plugin-prettier": "^4.2.1",
-    "prettier": "^2.8.4",
+    "eslint-config-airbnb-typescript": "^17.1.0",
+    "eslint-config-prettier": "^8.8.0",
+    "eslint-plugin-import": "^2.27.5",
+    "eslint-plugin-prettier": "^5.0.0",
+    "prettier": "^3.0.0",
     "typedoc": "0.23.26",
-    "typescript": "^4.6.3"
+    "typescript": "^5.1.6"
```
yusinto, line 47: Worth pointing out major typescript update here. I don't think this is a big deal because all the other sub projects are already on v5.0.

tsconfig.json
```diff
@@ -30,12 +30,6 @@
     },
     {
       "path": "./packages/sdk/akamai-base/tsconfig.ref.json"
-    },
-    {
-      "path": "./packages/store/node-server-sdk-redis/tsconfig.ref.json"
-    },
-    {
-      "path": "./packages/store/node-server-sdk-dynamodb/tsconfig.ref.json"
```
yusinto, line 38: `yarn check` [revealed](https://github.com/launchdarkly/js-core/pull/208/files#r1268734480) these don't exist, which is true so I deleted them.

packages/store/node-server-sdk-dynamodb/package.json
```diff
@@ -31,26 +31,27 @@
     "ioredis": "^5.3.2"
   },
   "peerDependencies": {
-    "@aws-sdk/client-dynamodb": "^3.53.0",
+    "@aws-sdk/client-dynamodb": "^3.370.0",
     "@launchdarkly/node-server-sdk": "8.x"
```
yusinto, line 35: Callout here I had to update this because typescript was erroring. I suspect upgrading from ts 4 to 5 needs this updating.

packages/store/node-server-sdk-dynamodb/package.json
```diff
@@ -31,26 +31,27 @@
     "ioredis": "^5.3.2"
   },
   "peerDependencies": {
-    "@aws-sdk/client-dynamodb": "^3.53.0",
+    "@aws-sdk/client-dynamodb": "^3.370.0",
```
yusinto, line 34: Callout here I had to update this because typescript was erroring. I suspect upgrading from ts 4 to 5 needs this updating.

.prettierrc
kinyoklion: I strongly favor semicolons, but I think it is only personal preference. I don't have any reason to add them, and I don't think there is any reason to remove them either. So, I'm not going to stop you.

packages/store/node-server-sdk-dynamodb/package.json
kinyoklion: We chose 53 for specific reasons, and if it doesn't work, then we should figure that out. If it cannot be resolved, then we should choose another version that isn't likely to inconvenience customers.

The peer dependency shouldn't have much of a bearing on if it works though. Things are compiled with the dev dependency. Currently it is the lowest to help maintain compatibility.

What is typescript error?

.prettierrc
yusinto: Thank you for the feedback. I have re-instated semicolons.

packages/store/node-server-sdk-dynamodb/package.json
yusinto: I rolled back these changes to reproduce the error but could not. As a result, I am not going to update the aws packages in this PR and keep them the same as main.

tsconfig.json
kinyoklion: Originally I had things compiling effectively as one large project. Now running tsc just double compiles some parts of it, because the build runs tsc on each thing in order. Because the builds had extra requirements like rollup.

kinyoklion (approved):

PR #212, opened 2023-07-24
feat: Add migrationVariation method.

kinyoklion:
For now I am adding this to server common, which means that edge SDKs would inherit the behavior. We may not want that, because they don't support events, but I can always exclude it or move the implementation.

### DIFF
--- a/packages/shared/sdk-server/src/LDClientImpl.ts
+++ b/packages/shared/sdk-server/src/LDClientImpl.ts
@@ -10,7 +10,15 @@ import {
   subsystem,
   internal,
 } from '@launchdarkly/js-sdk-common';
-import { LDClient, LDFlagsStateOptions, LDOptions, LDStreamProcessor, LDFlagsState } from './api';
+import {
+  LDClient,
+  LDFlagsStateOptions,
+  LDOptions,
+  LDStreamProcessor,
+  LDFlagsState,
+  LDMigrationStage,
+  IsMigrationStage,
+} from './api';
 import { BigSegmentStoreMembership } from './api/interfaces';
 import BigSegmentsManager from './BigSegmentsManager';
 import BigSegmentStoreStatusProvider from './BigSegmentStatusProviderImpl';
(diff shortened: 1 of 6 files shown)


packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -260,6 +268,23 @@ export default class LDClientImpl implements LDClient {
...
+    key: string,
+    context: LDContext,
+    defaultValue: LDMigrationStage,
+    callback?: (err: any, res: LDMigrationStage) => void
+  ): Promise<LDMigrationStage> {
+    const stringValue = await this.variation(key, context, defaultValue as string);
+    if (stringValue && !IsMigrationStage(stringValue)) {
+      const error = new Error(`Unrecognized MigrationState for "${key}"; returning default value.`);
+      this.onError(error);
+      callback?.(error, defaultValue);
```
kinyoklion, line 281: Most of the implementation here is just to handle the compatibility we have with callbacks. Everything supports both a callback and a promise. The error handling is also a little unique, we raise error events or put errors in callbacks.

packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -260,6 +268,23 @@ export default class LDClientImpl implements LDClient {
     return res.detail;
   }
 
+  async variationMigration(
+    key: string,
+    context: LDContext,
+    defaultValue: LDMigrationStage,
```
kinyoklion, line 274: As long as they do not break the typing the default value is of type LDMigrationStage, so I am not validating the default value.

kinyoklion: Though if it was not the correct type, and it came back out as the default, it would still get validated.

packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -260,6 +268,23 @@ export default class LDClientImpl implements LDClient {
...
+    callback?: (err: any, res: LDMigrationStage) => void
+  ): Promise<LDMigrationStage> {
+    const stringValue = await this.variation(key, context, defaultValue as string);
+    if (!IsMigrationStage(stringValue)) {
+      const error = new Error(`Unrecognized MigrationState for "${key}"; returning default value.`);
+      this.onError(error);
+      callback?.(error, defaultValue);
```
kinyoklion, line 281: Unrelated to this change, but I may consider removing callback support entirely for this method.

PR #213, opened 2023-07-25
feat: Add migration configuration and basic migration.

kinyoklion:
This does not have consistency tracking, error tracking, or latency tracking.

Most documentation is temporary.


packages/shared/sdk-server/src/api/options/LDMigrationOptions.ts
```diff
@@ -0,0 +1,127 @@
...
+    }
+  | {
+      success: false;
+      error: any;
+    };
+
+/**
+ * Configuration class for configuring serial execution of a migration.
+ */
+export class LDSerialExecution {
```
kinyoklion, line 66: I am not 100% on these being classes. But it can be easily changed if it proves cumbersome to use from JS.

packages/shared/sdk-server/src/Migration.ts
```diff
@@ -0,0 +1,221 @@
...
+  LDMethodResult,
+} from './api/options/LDMigrationOptions';
+import { LDMigration, LDMigrationResult } from './api/LDMigration';
+
+type MultipleReadResult<TMigrationRead> = {
+  fromOld: LDMethodResult<TMigrationRead>;
+  fromNew: LDMethodResult<TMigrationRead>;
+};
+
+async function safeCall<TResult>(
```
kinyoklion, line 18: I don't trust people to not have exceptions, so this handles an exception and turns it into an error return.

packages/shared/sdk-server/src/api/options/LDMigrationOptions.ts
```diff
@@ -0,0 +1,127 @@
...
+  Disabled,
+}
+
+/**
+ * Migration methods may return an LDMethodResult.
+ * The implementation includes methods for creating results conveniently.
+ *
+ * An implementation may also throw an exception to represent an error.
+ */
+export type LDMethodResult<TResult> =
```
kinyoklion, line 53: I've gone with a rust-like tagged union.

kinyoklion: (I'm not going to add monadic operation.)

packages/shared/sdk-server/__tests__/Migration.test.ts
```diff
@@ -0,0 +1,245 @@
...
+
+    await client.waitForInitialization();
+  });
+
+  afterEach(() => {
+    client.close();
+  });
+
+  describe.each([
+    [new LDSerialExecution(LDExecutionOrdering.Fixed), 'serial fixed'],
```
kinyoklion, line 42: I don't have builders, but these "execution" classes ensure correct combinations without having a potentially unused config at the top level.

packages/shared/sdk-server/src/api/LDMigration.ts
```diff
@@ -0,0 +1,62 @@
+import { LDContext } from '@launchdarkly/js-sdk-common';
+import { LDMigrationStage } from './data/LDMigrationStage';
+
+/**
+ * Specifies the origin of the result or error.
+ *
+ * Results from `readOld` or `writeOld` will be 'old'.
+ * Results from `readNew` or `writeNew` will be 'new'.
+ */
+export type LDMigrationOrigin = 'old' | 'new';
```
keelerm84, line 10: In the spec, I was thinking it might be useful to designate whether they were old or new sources, as well as if they were authoritative or not. This way the caller doesn't have to encode our same mapping table, and we can more easily change it in the future. What do you think?

keelerm84 (approved):
These changes look good. One point for consideration, but nothing that couldn't be addressed in a follow-up.

packages/shared/sdk-server/src/api/LDMigration.ts
kinyoklion: Yeah, I see how, at least for writes, that information could provide some value. I will try it.

keelerm84 (approved):

PR #214, opened 2023-07-26
feat: Add migration operation input event and tracker.

kinyoklion:


packages/shared/sdk-server/__tests__/LDClient.migrations.test.ts
```diff
@@ -76,30 +76,17 @@ describe('given an LDClient with test data', () => {
...
     td.update(td.flag(flagKey).valueForAll('potato'));
     const res = await client.variationMigration(flagKey, { key: 'test-key' }, LDMigrationStage.Off);
-    expect(res).toEqual(LDMigrationStage.Off);
+    expect(res.value).toEqual(LDMigrationStage.Off);
     expect(errors.length).toEqual(1);
     expect(errors[0].message).toEqual(
       'Unrecognized MigrationState for "bad-migration"; returning default value.'
     );
   });
-
```
kinyoklion, line 92: I decided to remove the callback interface from this function. It is old, and this is a new function.

packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -271,18 +273,36 @@ export default class LDClientImpl implements LDClient {
...
     key: string,
     context: LDContext,
-    defaultValue: LDMigrationStage,
-    callback?: (err: any, res: LDMigrationStage) => void
-  ): Promise<LDMigrationStage> {
-    const stringValue = await this.variation(key, context, defaultValue as string);
-    if (!IsMigrationStage(stringValue)) {
+    defaultValue: LDMigrationStage
+  ): Promise<LDMigrationDetail> {
+    const convertedContext = Context.fromLDContext(context);
```
kinyoklion, line 278: For now I still need to double convert the context, because the other code converts it at the last possible moment.

packages/shared/sdk-server/src/MigrationOpTracker.ts
```diff
@@ -0,0 +1,121 @@
...
+   * @param value The latency of the operation in milliseconds (TODO is it MS?).
+   */
+  latency(origin: LDMigrationOrigin, value: number) {
+    this.latencyMeasurement[origin] = value;
+  }
+
+  createEvent(): LDMigrationOpEvent | undefined {
+    if (this.operation && this.context.valid) {
+      const measurements: LDMigrationMeasurement[] = [];
+      if (
```
kinyoklion, line 68: This section code is not very elegant. I could probably find a way to refine it. Maybe split it up.

packages/shared/sdk-server/src/api/data/LDMigrationOpEvent.ts
```diff
@@ -0,0 +1,34 @@
+import { LDEvaluationReason } from '@launchdarkly/js-sdk-common';
+
+export type LDMigrationOp = 'read' | 'write';
+
+interface LDMigrationEvaluation {
```
kinyoklion, line 5: I should document this.

kinyoklion:
The linter has taken leave of its senses. But I am going to fix that after Yus merges all his linter changes.

keelerm84 (approved):

packages/shared/sdk-server/src/MigrationOpTracker.ts
```diff
@@ -0,0 +1,125 @@
...
+  constructor(
+    private readonly flagKey: string,
+    private readonly context: Context,
+    private readonly defaultStage: LDMigrationStage,
+    private readonly stage: LDMigrationStage,
+    private readonly reason: LDEvaluationReason,
+    private readonly variation?: number,
+  ) {}
+
+  op(op: LDMigrationOp) {
```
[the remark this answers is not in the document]
kinyoklion: It is required, but it isn't available when the tracker is constructed. So it has to be added along the way and then conditionally generate the event.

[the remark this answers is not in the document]
keelerm84: Ah yes, that makes sense.

PR #216, opened 2023-07-27
feat: Add support for exclude from summaries.

kinyoklion:
feat: Add `sampleWeight` to flag data model.

### DIFF
--- a/packages/shared/common/package.json
+++ b/packages/shared/common/package.json
@@ -24,7 +24,7 @@
     "build": "npx tsc",
     "clean": "npx tsc --build --clean",
     "lint": "npx eslint . --ext .ts",
-    "lint:fix": "yarn run lint -- --fix"
+    "lint:fix": "yarn run lint --fix"
   },
   "license": "Apache-2.0",
   "devDependencies": {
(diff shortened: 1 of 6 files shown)


packages/shared/common/package.json
```diff
@@ -24,7 +24,7 @@
     "build": "npx tsc",
     "clean": "npx tsc --build --clean",
     "lint": "npx eslint . --ext .ts",
-    "lint:fix": "yarn run lint -- --fix"
+    "lint:fix": "yarn run lint --fix"
```
keelerm84, line 27: Deja vu, we meet again.

packages/shared/common/src/internal/events/EventSummarizer.ts
```diff
@@ -52,7 +52,7 @@ export default class EventSummarizer {
   private contextKinds: Record<string, Set<string>> = {};
 
   summarizeEvent(event: InputEvent) {
-    if (isFeature(event)) {
+    if (isFeature(event) && !event.excludeFromSummaries) {
```
keelerm84, line 55: If `excludeFromSummaries` is optional, that means it defaults to undefined right? And `!undefined == true`? If so, I think this logic might be incorrect.

kinyoklion: `!undefined == true` also `!false == true` and `!true == false`

keelerm84: Ignore me. I'm a fool.

PR #217, opened 2023-07-31
DO NOT MERGE: feat: add the ability to send events using the Vercel Edge SDK

ldhenry:
This PR adds the ability to send events using the Vercel Edge SDK. The Edge SDK requires the client-side ID to retrieve feature flags from the Vercel Edge Config, so I had to add a new option for the `serverSideKey` so we can set the correct authorization header.

I have tested this locally and also published an alpha version of this branch to test in a deployed environment.


packages/shared/sdk-server-edge/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@launchdarkly/js-server-sdk-common-edge",
-  "version": "1.0.6",
+  "version": "1.1.0-alpha.0",
```
kinyoklion, line 3: This will break local linking for all the edge SDKs. I would prefer we never make manual changes to any of the package.json files, but I do see how it could be unavoidable sometimes. 

Generally these should be handled by the conventional commit tags and the "release-as" settings. 

The way this is right now the other SDKs are linking against the release version because the local version doesn't match what is in the repository. This means that testing those will be all kinds of inconsistent.

kinyoklion (commented):
I commented incorrectly. Didn't realize fetch was implemented, just didn't get called.

packages/shared/sdk-server-edge/package.json
kinyoklion: I think you can revert these now that you published your pre-release. Then release please will update them based on your conventional commit. Importantly it will also update all the other edge SDKs.

ldhenry: I just did this as so I could run the manual Release Github workflow on this branch. I will revert this change if/when we merge this into `main`.

Can you give me an example of how to do this with a conventional commit tag without merging this PR into main?

ldhenry: Cool. I'll revert now.

kinyoklion: So, if you want to release from a branch, as you did, then that is fine. But it is in PR against main now, so I have to be explicit.

ldhenry: Makes sense. Thanks for your help.

ldhenry:
I added DO NOT MERGE to the title because we are waiting on customer feedback of the beta build.

kinyoklion:
@ldhenry The user agent change is now merged, so you can remove the wrapper name/version.

packages/sdk/vercel/src/index.ts
```diff
@@ -36,17 +36,21 @@ export type { LDClient };
  * (`new LDClient()/new LDClientImpl()/new LDClient()`); the SDK does not currently support
  * this.
  *
- * @param sdkKey
- *  The client side SDK key. This is only used to query the edgeConfig above,
+ * @param clientSideID
+ *  The client-side ID. This is only used to query the edgeConfig above,
```
yusinto, line 40: Can we use the `sdkKey` use that to both query edgeConfig and also connect to streamer to send events? This reduces confusion having both clientSideID and serverSideKey which seems redundant to me.

packages/shared/sdk-server-edge/src/utils/validateOptions.ts
```diff
@@ -1,11 +1,16 @@
 import { LDOptions as LDOptionsCommon } from '@launchdarkly/js-server-sdk-common';
 
+type EdgeOptions = {
+  // The LaunchDarkly server-side key. This option is required if sendEvents is enabled.
+  serverSideKey?: string;
+};
```
yusinto, line 6: As I mentioned [above](https://github.com/launchdarkly/js-core/pull/217/files#r1305357502), we can avoid this additional option if we just reuse `sdkKey`. I think it's do-able.

yusinto (changes requested):

packages/sdk/vercel/src/index.ts
ldhenry: Unfortunately the env data is stored in the edge config DB under the client side ID so we need both

kinyoklion: As a secondary note I don't think you would want to use the SDK key for this, even if it was starting now. If the SDK key was used to store the edge config, then you would create some tricky situations if you wanted to rotate the SDK key. Client side ID is fixed.

(Though I would have overall preferred some identifier that wasn't the client-side ID. Simply because this makes it really conceptually confusing.)

ldhenry: yup - maybe we could eventually use something like `projectKey:envKey` for this but it would be a pretty big hassle to migrate over.

ldhenry:
Closing in favor of #256

PR #218, opened 2023-07-31
feat: Implement migration op event and connect to tracking.

kinyoklion:


packages/shared/common/src/Context.ts
```diff
@@ -15,9 +15,6 @@ import { TypeValidators } from './validators';
 // This is to reduce work on the hot-path. Later, for event processing, deeper
 // cloning of the context will be done.
 
-// Validates a kind excluding check that it isn't "kind".
```
kinyoklion, line 18: I moved kind validation into a common location and out of contexts.

packages/shared/sdk-server/__tests__/Migration.test.ts
```diff
@@ -176,8 +174,8 @@ describe('given an LDClient with test data', () => {
       it('uses the correct authoritative source', async () => {
         const migration = new Migration<string, boolean>(client, {
           execution,
-          latencyTracking: LDLatencyTracking.Disabled,
```
kinyoklion, line 179: I decided to switch these to just booleans, because it is a little easier JS wise, and because they are already named, so not like they will get mixed up.

keelerm84 (approved):

PR #220, opened 2023-08-02
feat: Add configuration overrides and metrics data kinds.

kinyoklion:


packages/shared/sdk-server/__tests__/data_sources/StreamingProcessor.test.ts
```diff
@@ -164,32 +174,24 @@ describe('given a stream processor with mock event source', () => {
   });
 
   describe('when patching a message', () => {
-    it('updates a patched flag', async () => {
+    it.each([
```
kinyoklion, line 177: Refactored these tests a little to simplify things.

packages/shared/sdk-server/__tests__/store/serialization.test.ts
```diff
@@ -151,11 +151,13 @@ const segmentWithBucketBy = {
   deleted: false,
 };
 
-function makeAllData(flag?: any, segment?: any): any {
+function makeAllData(flag?: any, segment?: any, override?: any, metric?: any): any {
```
kinyoklion, line 154: This pattern was ported from the old implementation. I don't really care for it, so maybe it is in for a change in the future.

packages/shared/sdk-server/src/store/Metric.ts
```diff
@@ -0,0 +1,5 @@
+import { Versioned } from '../evaluation/data/Versioned';
+
+export interface Metric extends Versioned {
+  samplingRatio: number;
```
kinyoklion, line 4: Currently I have this required, but is it?

packages/shared/sdk-server/src/store/serialization.ts
```diff
@@ -23,13 +25,15 @@ export function reviver(this: any, key: string, value: any): any {
   return value;
 }
 
-interface FlagsAndSegments {
+interface AllData {
```
kinyoklion, line 28: These names are internal, so I changed them to be less confusing. Should not hurt anything.

packages/shared/sdk-server/src/store/Metric.ts
yusinto: Is there a default?

yusinto (approved):

packages/shared/sdk-server/src/store/Metric.ts
kinyoklion: There is, but also currently this type specifically configured a sampling ratio, as it contains nothing else.

So the sampling ratio for a normal track event will be 1. If LD wants to change it from 1, then it will include one of these metric objects with a new value.

The question is really, if LD provides a metric, will it always have a samplingRatio. Or could we add other fields and then exclude this one.

keelerm84: Throughout other specs, we have worked on the premise that if it isn't set, it will default to 1. Seems silly here but maybe worth doing just to future proof and stay consistent?

keelerm84 (approved):

PR #223, opened 2023-08-03
fix: Ensure that test data user targets are handled correctly.

kinyoklion:
User targets require special logic to put the variation/values into `targets` instead of just in `contextTargets`. This was not being done, so user specific targets were not evaluating correctly.

Github issue: https://github.com/launchdarkly/node-server-sdk/issues/283


packages/shared/sdk-server/src/integrations/test_data/TestDataFlagBuilder.ts
```diff
@@ -394,18 +394,24 @@ export default class TestDataFlagBuilder {
 
     if (this.data.targetsByVariation) {
```
kinyoklion, line 395: The change in this file is the actual fix. Everything else is testing.

yusinto (approved):

PR #226, opened 2023-08-03
feat: Allow specifying the user agent per-sdk implementation.

kinyoklion:
Currently `NodeJSClient` is being used as the user agent for all server SDKs.

This was not apparent because only node was sending events. Events are being added to Vercel and it revealed this issue.

### DIFF
--- a/packages/sdk/server-node/src/platform/NodeInfo.ts
+++ b/packages/sdk/server-node/src/platform/NodeInfo.ts
@@ -37,6 +37,7 @@ export default class NodeInfo implements platform.Info {
     return {
       name: packageJson.name,
       version: packageJson.version,
+      userAgent: 'NodeJSClient',
       // No wrapper name/version at the moment.
     };
   }
--- a/packages/sdk/server-node/src/platform/NodeInfo.ts
+++ b/packages/sdk/server-node/src/platform/NodeInfo.ts
@@ -37,7 +37,7 @@ export default class NodeInfo implements platform.Info {
     return {
       name: packageJson.name,
       version: packageJson.version,
-      userAgent: 'NodeJSClient',
+      userAgentBase: 'NodeJSClient',
       // No wrapper name/version at the moment.
     };
   }
--- a/packages/sdk/vercel/src/createPlatformInfo.ts
+++ b/packages/sdk/vercel/src/createPlatformInfo.ts
@@ -11,6 +11,7 @@ class VercelPlatformInfo implements Info {
     return {
       name: '@launchdarkly/vercel-server-sdk',
       version: '__LD_VERSION__',
+      userAgent: 'VercelEdgeSDK',
     };
   }
 }
--- a/packages/sdk/vercel/src/createPlatformInfo.ts
+++ b/packages/sdk/vercel/src/createPlatformInfo.ts
@@ -11,7 +11,7 @@ class VercelPlatformInfo implements Info {
     return {
       name: '@launchdarkly/vercel-server-sdk',
       version: '__LD_VERSION__',
-      userAgent: 'VercelEdgeSDK',
+      userAgentBase: 'VercelEdgeSDK',
     };
   }
 }
--- a/packages/shared/akamai-edgeworker-sdk/src/platform/info/index.ts
+++ b/packages/shared/akamai-edgeworker-sdk/src/platform/info/index.ts
@@ -17,7 +17,7 @@ class AkamaiPlatformInfo implements Info {
     return {
       name: this.sdkName,
       version: this.sdkVersion,
-      userAgent: 'Akamai',
+      userAgent: 'AkamaiEdgeSDK',
     };
   }
 }
--- a/packages/shared/akamai-edgeworker-sdk/src/platform/info/index.ts
+++ b/packages/shared/akamai-edgeworker-sdk/src/platform/info/index.ts
@@ -17,7 +17,7 @@ class AkamaiPlatformInfo implements Info {
     return {
       name: this.sdkName,
       version: this.sdkVersion,
-      userAgent: 'AkamaiEdgeSDK',
+      userAgentBase: 'AkamaiEdgeSDK',
     };
   }
 }
--- a/packages/shared/akamai-edgeworker-sdk/src/platform/info/index.ts
+++ b/packages/shared/akamai-edgeworker-sdk/src/platform/info/index.ts
@@ -17,6 +17,7 @@ class AkamaiPlatformInfo implements Info {
     return {
       name: this.sdkName,
       version: this.sdkVersion,
+      userAgent: 'Akamai',
     };
   }
 }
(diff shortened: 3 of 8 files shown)


packages/shared/akamai-edgeworker-sdk/src/platform/info/index.ts
```diff
@@ -17,6 +17,7 @@ class AkamaiPlatformInfo implements Info {
     return {
       name: this.sdkName,
       version: this.sdkVersion,
+      userAgent: 'Akamai',
```
kinyoklion, line 20: @ctawiah or @dariushar What should the user agent be for Akamai? It isn't sending events, but if it ever does, then it may matter.

packages/sdk/vercel/src/createPlatformInfo.ts
```diff
@@ -11,6 +11,7 @@ class VercelPlatformInfo implements Info {
     return {
       name: '@launchdarkly/vercel-server-sdk',
       version: '__LD_VERSION__',
+      userAgent: 'VercelEdgeSDK',
```
kinyoklion, line 14: @ldhenry Once this is merged, then you won't need the wrapper name/version.

packages/shared/sdk-server/src/data_sources/defaultHeaders.ts
```diff
@@ -12,7 +12,7 @@ export default function defaultHeaders(
   const sdkData = info.sdkData();
   const headers: { [key: string]: string } = {
     authorization: sdkKey,
-    'user-agent': `NodeJSClient/${sdkData.version}`,
+    'user-agent': `${sdkData.userAgent ? sdkData.userAgent : 'NodeJSClient'}/${sdkData.version}`,
```
kinyoklion, line 15: Just uses the default for compatibility. Only really matters if someone uses a newer version of this package than their SDK shipped with.

louis-launchdarkly (commented):
I see a test failure about Akamai, and there already has a comment discussing Akamai. The other part looks good

kinyoklion:
> I see a test failure about Akamai, and there already has a comment discussing Akamai. The other part looks good

Should be resolved now. I'll wait to merge/release until I hear from some of the edge SDK folks.

louis-launchdarkly (approved):

packages/shared/akamai-edgeworker-sdk/src/platform/info/index.ts
ctawiah: Our integration userAgent typically follow this pattern `launchdarkly/{integration-name}` so i suggested an update to the name.

ctawiah, line 20:
```suggestion
      userAgent: 'launchdarkly/akamai',
```

kinyoklion: So. the user agent gets formed with a version. So this would be launchdarkly/akamai/version. I think this would cause problems with our event pipeline.

ctawiah: oh hmm, in the case we do `launchdarkly-akamai`

kinyoklion: Ok, so generally speaking these things we know are launchdarkly. So we have not had launchdarkly in the names before. They have not been terribly consistent. 

Like node is "NodeJSClient". We typically have had client/server in them.

I am leaning toward making it the same as the vercel one. AkamaiEdgeSDK.

ctawiah: Ok, that works.

packages/sdk/server-node/src/platform/NodeInfo.ts
```diff
@@ -37,6 +37,7 @@ export default class NodeInfo implements platform.Info {
     return {
       name: packageJson.name,
       version: packageJson.version,
+      userAgentBase: 'NodeJSClient',
```
yusinto, line 40: I am curious if we can just use `packageJson.name` instead? If we do then the useragent string will look like `@launchdarkly/js-server-sdk-common/1.0.6` which is longer, but it's ok?

kinyoklion: Well, this one needs to specifically be `NodeJSClient`. Because it not a new user agent name and there isn't a good reason to create new statistics.

The second thing, as I mentioned in another comment is, what would be the effect of this be on our pipeline. All of our user agents are of the form `thing/version`.

yusinto (approved):

packages/shared/sdk-server/src/data_sources/defaultHeaders.ts
```diff
@@ -12,7 +12,9 @@ export default function defaultHeaders(
   const sdkData = info.sdkData();
   const headers: { [key: string]: string } = {
     authorization: sdkKey,
-    'user-agent': `NodeJSClient/${sdkData.version}`,
+    'user-agent': `${sdkData.userAgentBase ? sdkData.userAgentBase : 'NodeJSClient'}/${
+      sdkData.version
+    }`,
```
yusinto, line 17: Eslint complained about using unnecessary ternary here, so to fix that:

```suggestion
'user-agent': `${userAgentBase ?? 'NodeJSClient'}/${version}`,
```

husseinsakr:
Does this mean we can now set the userAgent from the node server sdk? or is this internal stuff?

kinyoklion:
> Does this mean we can now set the userAgent from the node server sdk? or is this internal stuff?

It is internal stuff basically.

Our edge SDKs used to be wrappers around our node SDK, but when we moved to this repo we refactored that. We wanted to remove any node dependencies from the edge SDKs to prevent the need for a large number of polyfills. So we moved to a model with a common library with platform implementations per SDK.

This change allowed those SDKs to change the user agent.

What is the use case you have for changing the user agent? 

Thank you,
Ryan

husseinsakr:
Hi @kinyoklion , thank you for the reply. I missed your comment. 

Our use case is that it we today deploy the launchdarkly relay proxy in our environments and have wrapped that solution as well. Our clients connect to it but we have no visibility from the relay proxy on who is connected to it. 

I figured if we are able to set a custom userAgent in our clients. We can tell from the relay proxy which clients are connected to it by reading from the header.

kinyoklion:
@husseinsakr Yeah, this wouldn't really help your use case.

The user-agent, is used for some features, like reporting out of date SDKs or available upgrades, so changing it wouldn't be great.

In some SDKs we do allow custom headers to be added, which would potentially solve your problem. If you added your own custom header, then you could use it to collect such information.

This SDK doesn't support it, but it would likely be the right feature to request. Or there could be some more relay specific option that could help. You may wish to discuss it with support.

Thanks,
Ryan

PR #227, opened 2023-08-03
feat: Add custom event support to tracker.

kinyoklion:


packages/shared/sdk-server/src/MigrationOpTracker.ts
```diff
@@ -53,9 +56,13 @@ export default class MigrationOpTracker implements LDMigrationTracker {
...
   }
 
+  custom(measurement: LDMigrationCustomMeasurement) {
+    this.customMeasurements.push(measurement);
+  }
+
   createEvent(): LDMigrationOpEvent | undefined {
     if (this.operation && Object.keys(this.contextKeys).length) {
-      const measurements: LDMigrationMeasurement[] = [];
+      const measurements: LDMigrationMeasurement[] = [...this.customMeasurements];
```
yusinto, line 65:
```suggestion
      // ts should be able to infer the type automatically?
      // also why not just assign the array directly?
      const measurements = this.customMeasurements;
```

packages/shared/sdk-server/src/api/data/LDMigrationDetail.ts
```diff
@@ -19,10 +23,47 @@ export enum LDConsistencyCheck {
...
+   *
+   * @param measurement The custom measurement to track.
+   */
+  custom(measurement: LDMigrationCustomMeasurement): void;
+
+  /**
+   * Create a migration op event. If the event could not be created, because of a missing
+   * operation, then an event will not be returned.
+   */
   createEvent(): LDMigrationOpEvent | undefined;
```
yusinto, line 67:
```suggestion
  /**
   * Create a migration op event. If the event could not be created, because of a missing
   * operation, then undefined is returned.
   */
  createEvent(): LDMigrationOpEvent | undefined;
```

yusinto (approved):

packages/shared/sdk-server/src/MigrationOpTracker.ts
kinyoklion: If it was assigned, and you added another custom event to the same tracker, then it would end up inside the already generated event.

Similarly if you changed the event that was generated it would contaminate the state of the tracker.

Generally it may not be used that way, but it is exposed to outside developers, so I prefer the isolation.

The type should infer now, so I can remove that part.

PR #228, opened 2023-08-05
feat: reuse EventProcessor for dom

yusinto:
Please don't be alarmed with 55 file changes. Most are just import path changes due to existing files being moved from sdk-server to shared/common to be re-used across server and client sdks:

1. Move EventSender and its tests
2. Move errors.ts
3. Move NullEventProcessor
4. Move defaultHeaders and httpErrorMessage to utils/http.ts

In addition, please also note these important changes:

1. `EventSender` is now intrinsic to EventProcessor and is excluded from the EventProcessor constructor.
2. `contextDeduplicator` is now optional because client sdks don't use them
3. `defaultHeaders` arguments order have changed to `defaultHeaders(sdkKey, info, tags)`


packages/shared/common/src/api/options/LDClientContext.ts
```diff
@@ -17,7 +17,7 @@ export interface LDBasicConfiguration {
   /**
    * True if the SDK was configured to be completely offline.
    */
-  offline: boolean;
+  offline?: boolean;
```
yusinto, line 20: Client sdks don't use this option so this is optional now.

packages/shared/common/src/internal/events/EventSender.ts
```diff
@@ -1,58 +1,43 @@
...
-    this.diagnosticEventsUri = `${clientContext.basicConfiguration.serviceEndpoints.events}/diagnostic`;
-
-    this.requests = clientContext.platform.requests;
+  constructor(clientContext: ClientContext) {
+    const { basicConfiguration, platform } = clientContext;
+    const { sdkKey, serviceEndpoints, tags } = basicConfiguration;
+    const { crypto, info, requests } = platform;
 
-    this.crypto = clientContext.platform.crypto;
+    this.defaultHeaders = defaultHeaders(sdkKey, info, tags);
```
yusinto, line 26: As mentioned in PR desc, `defaultHeaders` arguments now take sdkKey, info and tags. It doesn't need the entire config so this is why I optimized it.

packages/shared/sdk-server/__tests__/data_sources/defaultHeaders.test.ts
yusinto, line 1: Moved to `shared/common/utils/http.test.ts`

packages/shared/common/src/utils/http.ts
```diff
@@ -0,0 +1,45 @@
...
+  'x-launchdarkly-wrapper'?: string;
+  'x-launchdarkly-tags'?: string;
+};
+
+export function defaultHeaders(sdkKey: string, info: Info, tags?: ApplicationTags): LDHeaders {
+  const { userAgentBase, version, wrapperName, wrapperVersion } = info.sdkData();
+
+  const headers: LDHeaders = {
+    authorization: sdkKey,
+    'user-agent': `${userAgentBase ?? 'NodeJSClient'}/${version}`,
```
yusinto, line 16: I applied the fix from #226 to specify a custom user-agent.

packages/shared/common/jest.config.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   transform: { '^.+\\.ts?$': 'ts-jest' },
-  testMatch: ['**/__tests__/**/*test.ts?(x)'],
+  testMatch: ['**/*.test.ts?(x)'],
```
yusinto, line 3: Include test files which are next alongside code files.

packages/shared/common/src/internal/events/EventSender.test.ts
```diff
@@ -0,0 +1,214 @@
...
+
+    expect(status).toEqual(LDDeliveryStatus.Succeeded);
+    expect(serverTime).toEqual(Date.now());
+    expect(error).toBeUndefined();
+    expect(mockFetch).toHaveBeenCalledTimes(1);
+    expect(mockFetch).toHaveBeenCalledWith(`${basicConfig.serviceEndpoints.events}/bulk`, {
+      body: JSON.stringify(testEventData1),
+      headers: analyticsHeaders(uuid),
+      method: 'POST',
+    });
```
yusinto, line 121: This test file is highly modified to use:

1. Fake timers to test server date time
2. Mock fetch to test http posts and response

The logic of each tests are kept to ensure no cases are lost.

packages/shared/common/src/options/ClientContext.ts
```diff
@@ -35,16 +38,18 @@ export default class ClientContext implements LDClientContext {
     sdkKey: string,
     configuration: {
       logger?: LDLogger;
-      offline: boolean;
+      offline?: boolean;
       serviceEndpoints: ServiceEndpoints;
+      tags?: ApplicationTags;
```
yusinto, line 43: Callout here the addition of ApplicationTags so we can easily access them from basicConfiguration.

packages/shared/sdk-server/src/data_sources/defaultHeaders.ts
yusinto, line 1: `defaultHeaders` and `httpErrorMessage` have been [moved](https://github.com/launchdarkly/js-core/pull/228/files#r1286283168) to `shared/common/utils/http`.

packages/shared/sdk-server/src/events/EventSender.ts
yusinto, line 1: `EventSender`, its test file and `NullEventProcessor` have been moved to `shared/common/internal`

packages/shared/common/src/internal/events/EventSender.ts
kinyoklion: It technically already only took tags. With the idea being if we needed to add additional information we didn't have to do call-site adjustment. It happened to work with passing it the config currently.

In the past it used two items of config, but after I worked through the platform abstraction it only took one. (Wrapper used to be part of the config, I think it took me a view iterations to arrive at the current arrangement).

I am not sure what else it could potentially need at this point though. So I don't have a strong opinion on it.

```
export interface DefaultHeaderOptions {
  tags: ApplicationTags;
}

export default function defaultHeaders(
  sdkKey: string,
  config: DefaultHeaderOptions,
  info: Info,
)
```

packages/shared/common/src/internal/events/EventProcessor.ts
```diff
@@ -76,62 +76,52 @@ interface LDDiagnosticsManager {
...
     config: EventProcessorOptions,
     clientContext: ClientContext,
-    private readonly eventSender: LDEventSender,
-    private readonly contextDeduplicator: LDContextDeduplicator,
+    private readonly contextDeduplicator?: LDContextDeduplicator,
     private readonly diagnosticsManager?: LDDiagnosticsManager,
   ) {
     this.capacity = config.eventsCapacity;
     this.logger = clientContext.basicConfiguration.logger;
+    this.eventSender = new EventSender(clientContext);
```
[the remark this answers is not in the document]
kinyoklion: I like the compos-ability, but I don't have strong opinion on it. It was left this way in case we wanted to allow people to provide an implementation, and because it IoC practices. Currently we aren't and changing it back wouldn't be hard.

An example of why, would be if someone wants to use unix domain sockets, of if they wanted some introspection.

kinyoklion (approved):

.gitignore
```diff
@@ -20,3 +20,6 @@ yarn-error.log
 .DS_Store
 .vscode
 dump.rdb
+packages/shared/common/src/**/*.d.ts
```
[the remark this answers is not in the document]
yusinto: I'm not sure. I found some files have them generated locally, and I'm not sure how they were being generated. I thought running `yarn tsc` manually somehow did it but I could not repro. I'll rollback this change.

packages/shared/common/src/index.ts
```diff
@@ -9,5 +9,7 @@ export * from './options';
 export * from './utils';
 
 export * as internal from './internal';
+export * from './errors';
+export * as mocks from './mocks';
```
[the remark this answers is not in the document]
yusinto: That's a good point. You are worried the public may see the mocks and use them in ways which can be problematic. If they were moved to internal, is that something you are willing to accept please?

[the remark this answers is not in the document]
kinyoklion: Yeah, I think that would be better. It makes it clear.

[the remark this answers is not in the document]
kinyoklion: As long as our client side builds are good and shakeable it also shouldn't cause bloat.

PR #234, opened 2023-08-11
feat: Use callbacks for evaluation hotpath.

kinyoklion:
This PR changes the evaluation to use callbacks instead of async. In the case where there is no IO there are many cases where execution will just by recursive without being deferred in the task system. 

Recursion does have its limits for collections though, large collections do use an immediately resolved promise to prevent the iteration from causing a stack overflow.

Generally speaking the callback depth here should be similar to that in the previous implementation, so it should not cause problems. Though I prefer greatly the async approach in regards to this.


packages/shared/sdk-server/src/evaluation/Evaluator.ts
```diff
@@ -101,13 +110,52 @@ export default class Evaluator {
   }
 
   async evaluate(flag: Flag, context: Context, eventFactory?: EventFactory): Promise<EvalResult> {
```
kinyoklion, line 112: For unit test compatibility I kept an async version.

yusinto (approved):
Using callbacks increases complexity make this more difficult to maintain. I feel nervous we are going back to callbacks, simultaneously understand the need for performance. I would like to request in the future we further investigate improving performance with async/await so we can have a win-win.

Are you able to tag this commit "revert to callbacks" so it's easily detectable for future reference please? Thank you.

kinyoklion:
> Using callbacks increases complexity make this more difficult to maintain. I feel nervous we are going back to callbacks, simultaneously understand the need for performance. I would like to request in the future we further investigate improving performance with async/await so we can have a win-win.
> 
> Are you able to tag this commit "revert to callbacks" so it's easily detectable for future reference please? Thank you.

Well, we never used callbacks in this implementation. So it isn't really reverting, and it isn't 100% equivalent to the callback implementation from the old implementation.

I can put callbacks in the title though.

packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -268,7 +269,7 @@ export default class LDClientImpl implements LDClient {
     });
   }
 
-  async allFlagsState(
+  allFlagsState(
```
kinyoklion, line 272: Because we have to make a manual promise anyway removing the async here may save a small amount of overhead.

packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -277,75 +278,80 @@ export default class LDClientImpl implements LDClient {
...
-              return;
-            }
-            this.evaluator.evaluateCb(flag, evalContext, (res) => {
-              if (res.isError) {
-                this.onError(
-                  new Error(
-                    `Error for feature flag "${flag.key}" while evaluating all flags: ${res.message}`,
-                  ),
-                );
+      const doEval = (valid: boolean) =>
```
kinyoklion, line 291: Some refactoring here to allow the callback version of the `initialized` function for the sore.

yusinto (approved):

PR #235, opened 2023-08-11
feat: Optimize segment lookup for large segments.

kinyoklion:


packages/shared/sdk-server/src/store/serialization.ts
```diff
@@ -104,6 +128,33 @@ export function processFlag(flag: Flag) {
  * @internal
  */
 export function processSegment(segment: Segment) {
+  if (segment?.included?.length && segment.included.length > 100) {
```
yusinto, line 131: Maybe declare `100` as a constant at the top of this file and comment why 100? For the benefit of future devs and also easy updating if required.

packages/shared/sdk-server/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "rootDir": "src",
     "outDir": "dist",
-    "target": "es2017",
+    "target": "ESNext",
```
yusinto, line 5: Should we update all projects to ESNext? I am supportive of that.

yusinto (changes requested):
Are we also able to add unit tests please?

packages/shared/sdk-server/tsconfig.json
kinyoklion: Ah, I was testing this. I will undo that for the moment. 

We would need to make sure that ESNext worked on all supported LTS versions of node.

kinyoklion: ESNext is a floating target though, so really we would need to choose a number.

yusinto (approved):

PR #237, opened 2023-08-14
fix: Allow for negation of segment match clauses.

kinyoklion:


packages/shared/sdk-server/__tests__/evaluation/Evaluator.segments.test.ts
```diff
@@ -110,6 +110,20 @@ describe('when evaluating user equivalent contexts for segments', () => {
...
 
+  it('can negate a segment match op', async () => {
+    const segment = {
+      key: 'test',
+      included: ['foo'],
+      version: 1,
+    };
+    const evaluator = new Evaluator(basicPlatform, new TestQueries({ segments: [segment] }));
+    const flag = makeFlagWithSegmentMatch(segment);
+    flag.rules[0].clauses![0].negate = true;
```
kinyoklion, line 121: Basically this should make the results of this test opposite of the above test.

louis-launchdarkly (approved):

PR #240, opened 2023-08-15
fix: Correct double callback in persistent store wrapper.

kinyoklion:


packages/shared/sdk-server/src/store/PersistentDataStoreWrapper.ts
```diff
@@ -166,6 +166,7 @@ export default class PersistentDataStoreWrapper implements LDFeatureStore {
         const value = deserialize(persistKind, descriptor);
         this.itemCache?.set(cacheKey(kind, key), value);
         callback(itemIfNotDeleted(value));
+        return;
```
kinyoklion, line 169: In the case where there was item the callback was getting executed again. When using the async facade this worked fine, which is what the SDK has been doing. The callback based implementation does not use the facade, and was encountering this problem.

PR #243, opened 2023-08-16
feat: Remove custom events.

kinyoklion:


packages/shared/sdk-server/src/MigrationOpEventConversion.ts
```diff
@@ -54,27 +49,13 @@ function areValidValues(values: { old?: number; new?: number }) {
 function validateMeasurement(
   measurement: LDMigrationMeasurement,
 ): LDMigrationMeasurement | undefined {
+  // Here we are protecting ourselves from JS callers. TypeScript says that
+  // it cannot be an empty string, but those using JS can do what they want.
+  // @ts-ignore
```
yusinto, line 54: 👍

packages/shared/sdk-server/src/api/data/LDMigrationOpEvent.ts
```diff
@@ -53,8 +44,7 @@ export interface LDMigrationErrorMeasurement {
 export type LDMigrationMeasurement =
   | LDMigrationLatencyMeasurement
   | LDMigrationErrorMeasurement
-  | LDMigrationConsistencyMeasurement
-  | LDMigrationCustomMeasurement;
```
yusinto, line 57: So this will break existing applications. This is ok, if the plan is for a major release?

keelerm84 (approved):

packages/shared/sdk-server/src/api/data/LDMigrationOpEvent.ts
kinyoklion: None of this is released. All on a feature branch.

yusinto (approved):

PR #245, opened 2023-08-18
feat: Event sampling.

kinyoklion:


packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -410,16 +456,42 @@ export default class LDClientImpl implements LDClient {
...
     const converted = MigrationOpEventToInputEvent(event);
-    if (converted) {
-      this.eventProcessor.sendEvent(converted);
+    if (!converted) {
+      return;
     }
+    // Async immediately invoking function expression to get the flag from the store
+    // without requiring track to be async.
+    (async () => {
+      // TODO: Would it be better to pass this through.
```
kinyoklion, line 483: https://app.shortcut.com/launchdarkly/story/213635/change-the-variationmigration-method-to-return-the-sampling-ratio-so-it-doesn-t-have-to-be-looked-up-when-tracking-events

packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -445,10 +517,10 @@ export default class LDClientImpl implements LDClient {
     context: LDContext,
     defaultValue: any,
     eventFactory: EventFactory,
-  ): Promise<EvalResult> {
+  ): Promise<[EvalResult, Flag?]> {
```
kinyoklion, line 520: Need to pass back the original flag to get the sampling ratio.

packages/shared/common/__tests__/internal/events/EventProcessor.test.ts
```diff
@@ -302,6 +313,102 @@ describe('given an event processor', () => {
...
+      makeSummary(1000, 1000, 1, 11),
+    ]);
+  });
+
+  it('excludes index events that are not sampled', async () => {
+    // @ts-ignore
+    shouldSample.mockImplementation((ratio) => ratio === 2);
+    Date.now = jest.fn(() => 1000);
+    eventProcessor.sendEvent({
+      kind: 'feature',
```
louis-launchdarkly, line 387: Is this suppose to say `index`?

louis-launchdarkly (commented):
Thank you for walking through the change with me during 1:1. There is a single comment for the unit test, but everything else make sense and looks good.

packages/shared/common/__tests__/internal/events/EventProcessor.test.ts
kinyoklion: No. Basically index events are created via 'evaluation' events. So here we have a feature input event. Based on a single input event the event processor may create up to 4 output events (feature, debug, index, summary). 

Here we have a sampling ratio of 2 for the feature, which the mock will say to sample.
```
shouldSample.mockImplementation((ratio) => ratio === 2);
```

The index sampling is 1, which the mock will say not to sample.

Then we check that it sampled them, and we check that it did not produce an index event, but just a feature event and summary event.

louis-launchdarkly (approved):

PR #246, opened 2023-08-21
feat: Cache interim eval results during allFlagsState.

kinyoklion:
When evaluating all flags we will often evaluate the same flag several times. For instance a prereq flag will be evaluated as part of evaluating all flags, as well as each time it appears as a prereq.

For a segment it will be evaluated each time it is referenced in a clause.

This PR tracks evaluations and re-uses results for a single allFlagsState call.

This method cannot be used when evaluating a single flag. This is because during single flag evaluation we evaluate prerequisites of prerequisites and need to generate events for however many layers deep there are. Using a cached result would affect the total prerequisite evaluations.


packages/shared/sdk-server/src/evaluation/Evaluator.ts
```diff
@@ -65,12 +65,35 @@ function computeUpdatedBigSegmentsStatus(
...
+}
+
+function isMissingMarker(u: unknown): u is MissingMarker {
+  if ((u as any).isMissing) {
+    return true;
+  }
+  return false;
+}
+
+export interface EvalCache {
```
kinyoklion, line 79: So, I exported this (just inside the package), but I have not exported the constituent parts. Basically just let the client pass in an empty list. I could export more. Basically the state needs to be held externally, to re-use between eval calls for evaluating all flags, but it shouldn't be interacted with.

PR #255, opened 2023-08-24
chore: Change akamai module to nodenext

kinyoklion:
I am putting these back to nodenext. After pinning typescript I found that rollup and typescript are not agreeing on what these options should be set to.


ctawiah (approved):
👍🏽

PR #256, opened 2023-08-26
feat: edge sdks should send events to bulk/environment endpoint

yusinto:
Edge sdks use clientSideID not sdk-key. We should be able to post events to the eventsUri/bulk/clientSideID endpoint for edge sdks without needing to specify the sdk key in the authorization header.

This is an alternate solution to #217. Both prs are attempting to enable events for edge sdks.

Don't be alarmed with the filecount, it's mostly shell file fixes unrelated to this to include a directive which seems to be needed for local build.


packages/shared/sdk-server/src/events/EventSender.ts
```diff
@@ -28,21 +28,26 @@ export default class EventSender implements subsystem.LDEventSender {
...
-    this.eventsUri = `${clientContext.basicConfiguration.serviceEndpoints.events}/bulk`;
-
-    this.diagnosticEventsUri = `${clientContext.basicConfiguration.serviceEndpoints.events}/diagnostic`;
-
-    this.requests = clientContext.platform.requests;
+    const isClientSideID = !sdkKey.startsWith('sdk-');
+    this.eventsUri = isClientSideID ? `${events}/events/bulk/${sdkKey}` : `${events}/bulk`;
+    this.diagnosticEventsUri = isClientSideID
+      ? `${events}/events/diagnostic/${sdkKey}`
+      : `${events}/diagnostic`;
```
yusinto, line 47: Use the appropriate endpoints based on [this](https://launchdarkly.atlassian.net/wiki/spaces/PD/pages/2129461878/Service+endpoints+used+by+SDKs).

yusinto:
I still need to test this PR in a real application but based on existing js sdk code, this PR should work.

packages/sdk/cloudflare/example/src/index.ts
```diff
@@ -2,16 +2,19 @@ import { init as initLD } from '@launchdarkly/cloudflare-server-sdk';
...
     // start using ld
-    const client = initLD(sdkKey, env.LD_KV);
+    const client = initLD(clientSideID, env.LD_KV, { sendEvents: true });
     await client.waitForInitialization();
     const flagValue = await client.variation(flagKey, context, false);
     const flagDetail = await client.variationDetail(flagKey, context, false);
     const allFlags = await client.allFlagsState(context);
+    await client.flush((err, res) => {
+      console.log(`============ flushed events result: ${res}. error: ${err}`);
+    });
```
yusinto, line 17: GOTCHA: critical to flush otherwise events are not going to be sent.

packages/shared/sdk-server/src/events/EventSender.ts
```diff
@@ -66,6 +76,9 @@ export default class EventSender implements subsystem.LDEventSender {
     }
     let error;
     try {
+      this.logger?.error(`===== uri: ${uri}`);
+      this.logger?.error(`===== headers: ${JSON.stringify(headers)}`);
+      this.logger?.error(`===== events: ${JSON.stringify(events)}`);
```
yusinto, line 81: Debugging only and will remove prior to merging.

scripts/build-doc.sh
```diff
@@ -1,3 +1,5 @@
+#!/bin/bash
```
yusinto, line 1: Locally `yarn build` failed with a weird yarn error without these.

packages/sdk/cloudflare/example/src/index.ts
```diff
@@ -2,16 +2,19 @@ import { init as initLD } from '@launchdarkly/cloudflare-server-sdk';
 
 export default {
   async fetch(request: Request, env: Bindings): Promise<Response> {
-    const sdkKey = 'test-sdk-key';
-    const flagKey = 'testFlag1';
-    const context = { kind: 'user', key: 'test-user-key-1', email: 'test@gmail.com' };
+    const clientSideID = '59b2b2596d1a250b1c78baa4';
```
yusinto, line 5: Debugging only and will remove prior to merging.

link-dev.sh
yusinto, line 1: The cloudflare example app doesn't work with modern yarn workspaces so I have to do this manually.

yusinto:
I ran the cloudflare example app using this code which uses clientSideID and events seem to be sent.

Screenshot 2023-08-27 at 7 36 47 PM

kinyoklion:
I am not sure about this. The events that can be generated from a client SDK are not 100% the same. So I do not know how the endpoints for a client SDK handle the data.

The server-side SDKs, which these are, generate prerequisite record events. Would the client SDK endpoint for events handle those correctly? If we have a definitive answer, then I can see pursuing it.

kinyoklion:
Secondarily it appears these changes break many contract tests, so some investigation is need there.

packages/shared/sdk-server/src/data_sources/defaultHeaders.ts
```diff
@@ -11,12 +11,16 @@ export default function defaultHeaders(
...
-    authorization: sdkKey,
     'user-agent': `${sdkData.userAgentBase ? sdkData.userAgentBase : 'NodeJSClient'}/${
       sdkData.version
     }`,
   };
 
+  // edge sdks use clientSideID and don't need the authorization header
+  if (sdkKey.startsWith('sdk-')) {
+    headers.authorization = sdkKey;
+  }
```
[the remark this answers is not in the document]
kinyoklion: This far SDK keys have been opaque to the SDK, so likely this will cause problems in testing infrastructure. It may be why the contract tests break. I think we would need some discussions to agree to the SDKs treating keys as non-opaque.

packages/shared/sdk-server/src/events/EventSender.ts
kinyoklion: Similarly uncomfortable with this change. If we want to do something like this, then likely we should move the full URL to earlier config. Then use another piece of either config or platform to control it.

At the most basic level `SdkAuthType` which has `Mobile`, `Server`, `ClientSide`. This would get similarly messy if we add more hybrid SDKs like these edge SDKs. 

At the more complex side just make a function that can be provided by the SDK implementation to get the endpoints. I would generally prefer this, edge SDKs can control their destiny and don't have to have implementation details reaching as deep. It also would allow for more runtime config, which could happen for something like RN (if an SDK ever supports RN and web targets).

yusinto:
Thanks for your feedback. I submitted this PR having done due diligence checking the event recorder logic how it handles events for server and client sdks.

My investigation reveals both the [/bulk](https://github.com/launchdarkly/event-recorder/blob/89b0dc55660d31473f5b4345a0623f85806afd47/event-recorder/event-recorder.go#L814) and [/events/bulk/{envId}](https://github.com/launchdarkly/event-recorder/blob/89b0dc55660d31473f5b4345a0623f85806afd47/event-recorder/event-recorder.go#L833) endpoints are handled equally by the [postBulkEvents](https://github.com/launchdarkly/event-recorder/blob/89b0dc55660d31473f5b4345a0623f85806afd47/event-recorder/event-recorder.go#L136) handler. You'll see this postBulkEvents contain the [RecordRequest](https://github.com/launchdarkly/event-recorder/blob/89b0dc55660d31473f5b4345a0623f85806afd47/event-recorder/event-recorder.go#L141) logic you are concerned with so I think this is covered for both server and client sdks.

I understand your concerns though and you want us to be thorough with this kind of hybrid logic because we never had to mix the two before. The edge sdks are unique in this way, it is truly somewhat of a hybrid. This approach opens the door for the future for such sdks so I'm hopeful this works. I will assign someone from the events pipeline to review my work.

I will investigate the contract tests failures, I suspect they are related to the `sdk-` key prefix which is necessary now. If those tests use some arbitrary strings as sdkKey then those will fail.

yusinto:
Added @samstokes as reviewer from the Events pipeline team.

packages/shared/sdk-server-edge/src/api/LDClient.ts
```diff
@@ -12,11 +12,11 @@ import createOptions from './createOptions';
 export class LDClient extends LDClientImpl {
   emitter: EventEmitter;
 
-  // sdkKey is only used to query featureStore, not to initialize with LD servers
-  constructor(sdkKey: string, platformInfo: Info, options: LDOptions) {
+  // clientSideID is only used to query featureStore, not to initialize with LD servers
+  constructor(clientSideID: string, platformInfo: Info, options: LDOptions) {
```
samstokes, line 16: This suggests that this change might have broader implications than just sending events. If the SDK uses a client-side id rather than SDK key _for authentication to the flag delivery services_ (streamer and flagbearer), it will actually receive different information - it'll need to specify user attributes along with the call, and will receive evaluated flag payloads for that user, rather than the flag evaluation ruleset.

I am guessing this actually means this comment is stale, rather than an actual change in behavior, because I assume the SDK has already been modified to get flag values from the edge store rather than the flag delivery endpoints.

I find this comment confusing, though, and possibly misleading - could we update it to clarify what if anything the client-side id is actually used for here?

packages/shared/sdk-server/src/events/EventSender.ts
samstokes: I'll defer to y'all to decide on code organization, but just to agree from a security standpoint it'd be nice if there was a clearer distinction between `sdkKey` and `clientSideID` in the code, rather than storing them both in a variable called `sdkKey` and relying on just a prefix to disambiguate them. _(Not a blocker for me, just a +1 to the existing comment.)_

SDK keys are sensitive and client-side IDs are not, and it'd be bad if we accidentally sent an SDK key in a URL - which aren't generally treated as sensitive, and could get copy-pasted, logged by proxies, etc.

samstokes (commented):
LGTM from an event pipeline standpoint. (Leaving it to the SDK team to do the approving.)

Re:
> The events that can be generated from a client SDK are not 100% the same. So I do not know how the endpoints for a client SDK handle the data.

I can confirm that both endpoints behave identically in terms of what event payloads they will accept. e.g. even though only browser SDKs generate `click` events, the `/bulk` endpoint that handles server-side SDK payloads will happily accept a `click` event. It all goes into the same Kinesis stream :)

Re the specific concern about prerequisite events:
> The server-side SDKs, which these are, generate prerequisite record events. Would the client SDK endpoint for events handle those correctly? If we have a definitive answer, then I can see pursuing it.

I don't think anything in our event pipeline does anything special with prerequisites. Definitely the events endpoint handlers don't.

(AFAIK there's nothing special about prerequisite _events_, right - they're just regular evaluation events? And the only special consideration is that client-side SDKs don't send them.

packages/shared/sdk-server-edge/src/api/LDClient.ts
kinyoklion: It is indeed confusing. Edge SDKs operate as server SDKs, but they don't connect to LD to get the flag configuration. They are similar to daemon mode. The expectation is that some edge specific store contains the flag configuration. At some point it was decided that the flag configuration would be stored by the client side ID. So the client side ID is the key, or component of the key, where the flag configuration is stored.

kinyoklion: So, they have to have that ID already. So it is ideal if they can also use it for events. There is another PR which takes both keys and uses them for different purposes.

samstokes: I see. Maybe this is because the client-side ID is also the environment id, which is the logical "owner" of the flag configuration? I guess I retract my suggestion to update the comment. (Maybe the previous wording which said that the SDK key was "used to query featureStore" was incorrect?)

yusinto:
```suggestion
  // clientSideID is only used to query featureStore and send analytics, not to initialize with LD servers
  constructor(clientSideID: string, platformInfo: Info, options: LDOptions) {
```

kinyoklion: I think the problem is that featureStore is a bit confusing. It is used to initialize the special edge SDK feature store. It isn't really used to query it. It is used to query the edge key-value store.

yusinto:
> from a security standpoint it'd be nice if there was a clearer distinction between sdkKey and clientSideID in the code

Thanks @samstokes ! I have improved the code to provide a better distinction between sdkKey and clientSideID.

kinyoklion (approved):

ldhenry (commented):
This looks great @yusinto. Thanks or confirming the client-side endpoint can handle all of the events correctly. Before merging and releasing, I think it would be good to publish a pre-release version so we can test it in a deployed environment. I ran out of time today but I can give it a shot tomorrow.

yusinto:
Thanl you @ldhenry ! I'll go ahead and do a pre-release. Right now the cloudflare sdk is at 2.1.3. I think this should just be a minor release, so I'll release 2.2.0-alpha. Any objections?

release-please-config.json
```diff
@@ -7,9 +7,14 @@
...
-    "packages/sdk/cloudflare": {},
+    "packages/sdk/cloudflare": {
+      "bump-minor-pre-major": true,
+      "release-as": "2.2.0-alpha"
+    },
     "packages/sdk/vercel": {
-      "extra-files": ["src/createPlatformInfo.ts"]
+      "extra-files": ["src/createPlatformInfo.ts"],
+      "bump-minor-pre-major": true,
+      "release-as": "1.2.0-alpha"
```
yusinto, line 17: @kinyoklion are these correct? Is there anymore I need to do for an alpha release?

packages/sdk/cloudflare/example/src/index.ts
```diff
@@ -1,24 +1,30 @@
...
-    const client = initLD(sdkKey, env.LD_KV);
+    const client = initLD(clientSideID, env.LD_KV, { sendEvents: true });
     await client.waitForInitialization();
     const flagValue = await client.variation(flagKey, context, false);
     const flagDetail = await client.variationDetail(flagKey, context, false);
     const allFlags = await client.allFlagsState(context);
 
+    // Gotcha: you must call flush otherwise events will not be sent to LD servers
+    // due to the ephemeral nature of edge workers.
+    await client.flush((err, res) => {
```
ldhenry, line 19: Let's put this inside of a [`context.waitUntil`](https://developers.cloudflare.com/workers/runtime-apis/fetch-event/#waituntil) to show customers how to flush events without impacting the handler's response time.

packages/sdk/cloudflare/example/src/index.ts
```diff
@@ -1,24 +1,30 @@
+/* eslint-disable no-console */
 import { init as initLD } from '@launchdarkly/cloudflare-server-sdk';
 
 export default {
   async fetch(request: Request, env: Bindings): Promise<Response> {
```
ldhenry, line 5: We'll need to add the context argument to leverage the [`waitUntil` method](https://developers.cloudflare.com/workers/runtime-apis/fetch-event/#waituntil) (more detail below).
```suggestion
  async fetch(request: Request, env: Bindings, context: ExecutionContext): Promise<Response> {
```

ldhenry (changes requested):

yusinto:
@ldhenry pls re-review when able. Thank you

ldhenry:
> @ldhenry pls re-review when able. Thank you

Hey @yusinto, I was hoping to get some sort of evidence that this is working from our alpha customers before making this more widely available. However, I can't see any usage in [honeycomb](https://ui.honeycomb.io/launchdarkly/datasets/production/result/koC6r4UCrvH?hideMarkers). Do you know if there is a better way to see adoption?

yusinto:
> Do you know if there is a better way to see adoption?

Yep we can try querying cmetrics directly following [these instructions](https://launchdarkly.atlassian.net/wiki/spaces/ENG/pages/2192638480/SDK+Notifications+Phase+1+Tech+Spec).

samstokes (commented):
Nothing to add, just commenting so GitHub will stop reminding me about this PR. If there's anything that does need my feedback please feel free to "re-request review".

louis-launchdarkly (approved):

packages/sdk/cloudflare/example/src/index.ts
mmrj, line 1: To the Vercel and/or Akamai examples also need to be updated? I initially thought this change only applied to Cloudflare because only the Cloudflare example was changed.

yusinto: Those examples can be updated, but not necessary. I'll leave that to @ldhenry .

yusinto:
@ldhenry please approve so this can be merged. Thank you.

ldhenry (approved):

PR #258, opened 2023-08-28
feat: Add invoked measurement.

kinyoklion:
The migration_op event now has a measurement for which "origins" were invoked. Basically was the 'new' method, the 'old' method, or both executed.

This is because from other data you cannot safely infer what was executed. We need to know total executions in order to properly analyze results.

### DIFF
--- a/contract-tests/index.js
+++ b/contract-tests/index.js
@@ -29,6 +29,7 @@ app.get('/', (req, res) => {
       'big-segments',
       'user-type',
       'migrations',
+      'event-sampling'
     ],
   });
 });
--- a/contract-tests/index.js
+++ b/contract-tests/index.js
@@ -29,7 +29,7 @@ app.get('/', (req, res) => {
       'big-segments',
       'user-type',
       'migrations',
-      'event-sampling'
+      'event-sampling',
     ],
   });
 });
--- a/contract-tests/sdkClientEntity.js
+++ b/contract-tests/sdkClientEntity.js
@@ -64,7 +64,7 @@ export function makeSdkConfig(options, tag) {
 }
 
 function getExecution(order) {
-  switch(order) {
+  switch (order) {
     case 'serial': {
       return new LDSerialExecution(LDExecutionOrdering.Fixed);
     }
@@ -164,7 +164,7 @@ export async function newSdkClientEntity(options) {
       case 'migrationOperation':
         const migrationOperation = params.migrationOperation;
         const readExecutionOrder = migrationOperation.readExecutionOrder;
-        
+
         const migration = new Migration(client, {
           execution: getExecution(readExecutionOrder),
           latencyTracking: migrationOperation.trackLatency,
--- a/packages/shared/sdk-server/__tests__/MigrationOpEvent.test.ts
+++ b/packages/shared/sdk-server/__tests__/MigrationOpEvent.test.ts
@@ -191,6 +191,44 @@ describe('given an LDClient with test data', () => {
         });
       });
 
+      it.each([
+        [LDMigrationStage.Off, { old: true }],
+        [LDMigrationStage.DualWrite, { old: true }],
+        [LDMigrationStage.Shadow, { old: true, new: true }],
+        [LDMigrationStage.RampDown, { new: true }],
+        [LDMigrationStage.Complete, { new: true }],
+      ])('tracks the invoked methods for reads', async (stage, values) => {
+        const flagKey = 'migration';
+        td.update(td.flag(flagKey).valueForAll(stage));
+
+        await migration.read(flagKey, { key: 'test' }, stage);
+        // Feature event.
+        await events.take();
+        // Migration event.
+        const migrationEvent = (await events.take()) as internal.InputMigrationEvent;
+        expect(migrationEvent.measurements[0].key).toEqual('invoked');
+        expect(migrationEvent.measurements[0].values).toEqual(values);
+      });
+
+      it.each([
+        [LDMigrationStage.Off, { old: true }],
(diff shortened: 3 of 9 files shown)


contract-tests/index.js
```diff
@@ -29,6 +29,7 @@ app.get('/', (req, res) => {
       'big-segments',
       'user-type',
       'migrations',
+      'event-sampling',
```
kinyoklion, line 32: To opt-in to the event-sampling tests in the contract test.

contract-tests/sdkClientEntity.js
```diff
@@ -64,7 +64,7 @@ export function makeSdkConfig(options, tag) {
 }
```
kinyoklion, line 64: Prettier attacked this file.

packages/shared/sdk-server/__tests__/MigrationOpEvent.test.ts
```diff
@@ -91,9 +91,9 @@ describe('given an LDClient with test data', () => {
           // Migration event.
           const migrationEvent = (await events.take()) as internal.InputMigrationEvent;
           // Only check the measurements component of the event.
-          expect(migrationEvent.measurements[0].key).toEqual('consistent');
+          expect(migrationEvent.measurements[1].key).toEqual('consistent');
```
kinyoklion, line 94: Adding the new measurement consistently shifted things.

packages/shared/sdk-server/src/MigrationOpTracker.ts
```diff
@@ -56,31 +63,118 @@ export default class MigrationOpTracker implements LDMigrationTracker {
...
+        default: this.defaultStage,
+        reason: this.reason,
+        variation: this.variation,
+      },
+      measurements,
+      samplingRatio: this.samplingRatio ?? 1,
+    };
+  }
+
+  private logTag(): string {
```
kinyoklion, line 116: I added various log messages for things that an be wrong with an event. These basically only apply if you are using the blocks but not the kit.

kinyoklion:
Tests are fixed in the PR which adds the new data kinds to edge SDKs.

packages/shared/sdk-server/src/Migration.ts
```diff
@@ -302,6 +302,7 @@ export default class Migration<
     origin: LDMigrationOrigin,
     method: (payload?: TInput) => Promise<LDMethodResult<TOutput>>,
   ): Promise<LDMigrationResult<TOutput>> {
+    context.tracker.invoked(origin);
```
kinyoklion, line 305: This is the call that actually causes the tracker method to be invoked in all cases.

packages/shared/sdk-server/src/MigrationOpTracker.ts
```diff
@@ -56,31 +63,118 @@ export default class MigrationOpTracker implements LDMigrationTracker {
...
+    }
+
+    if (Object.keys(this.contextKeys).length === 0) {
+      this.logger?.error(
+        'The migration was not done against a valid context and cannot generate an event.',
+      );
+      return undefined;
+    }
+
+    if (!this.wasInvoked.old && !this.wasInvoked.new) {
```
kinyoklion, line 83: If neither the old or the new implementation was invoked, then that isn't valid, and we will not create the event.

packages/shared/sdk-server/src/MigrationOpTracker.ts
```diff
@@ -20,6 +21,11 @@ export default class MigrationOpTracker implements LDMigrationTracker {
     new: false,
   };
 
+  private wasInvoked = {
```
kinyoklion, line 24: Used to track which implementations were executed.

packages/shared/sdk-server/src/MigrationOpTracker.ts
```diff
@@ -38,6 +44,7 @@ export default class MigrationOpTracker implements LDMigrationTracker {
     private readonly checkRatio?: number,
     private readonly variation?: number,
     private readonly samplingRatio?: number,
+    private readonly logger?: LDLogger,
```
kinyoklion, line 47: Added the logger so we can provide some useful information for invalid cases. Tracking what method was invoked allows us to perform some consistency checks on the data and log when there is an issue.

packages/shared/sdk-server/src/MigrationOpTracker.ts
```diff
@@ -56,31 +63,118 @@ export default class MigrationOpTracker implements LDMigrationTracker {
...
-          key: this.flagKey,
-          value: this.stage,
-          default: this.defaultStage,
-          reason: this.reason,
-          variation: this.variation,
-        },
-        measurements,
-        samplingRatio: this.samplingRatio ?? 1,
-      };
+    if (!this.operation) {
```
kinyoklion, line 71: Inverted the logic here to have early returns.

yusinto (approved):

packages/shared/sdk-server/src/MigrationOpEventConversion.ts
```diff
@@ -28,6 +29,12 @@ function isErrorMeasurement(value: LDMigrationMeasurement): value is LDMigration
   return (value as any).kind === undefined && value.key === 'error';
 }
 
+function isInvokedMeasurement(
+  value: LDMigrationMeasurement,
+): value is LDMigrationInvokedMeasurement {
+  return (value as any).kind === undefined && value.key === 'invoked';
```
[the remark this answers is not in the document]
kinyoklion: I can remove that. Previously we were going to support `custom` measurements. Which had a kind, which created an unpleasant difference in shape. 

They have been tabled for now though, so this isn't needed.

PR #259, opened 2023-08-28
chore: better support for event sender analytics paths

yusinto:


packages/shared/sdk-server-edge/src/api/LDClient.ts
```diff
@@ -16,7 +17,12 @@ export class LDClient extends LDClientImpl {
   constructor(clientSideID: string, platformInfo: Info, options: LDOptions) {
     const em = new EventEmitter();
     const platform = new EdgePlatform(platformInfo);
-    super(clientSideID, platform, createOptions(options), createCallbacks(em));
+    const internalOptions: internal.LDInternalOptions = {
```
kinyoklion, line 20: Yeah, I like this much better.

kinyoklion (approved):

packages/store/node-server-sdk-dynamodb/README.md
yusinto, line 1: Prettier auto fixes this

packages/shared/common/src/options/ServiceEndpoints.ts
```diff
@@ -12,9 +12,36 @@ export default class ServiceEndpoints {
...
+   * /mobile
+   */
+  public readonly analyticsEventPath: string;
+
+  /** Valid paths are:
+   * /diagnostic
+   * /events/diagnostic/envId
+   * /mobile/events/diagnostic
+   */
+  public readonly diagnosticEventPath: string;
```
yusinto, line 27: We can be more pedantic and make these literal types instead of freeform strings. Any opinions?

PR #261, opened 2023-08-29
chore: Change typedoc setup to use monorepo.

kinyoklion:
Updated our typedoc to be based on this example: https://github.com/Gerrit0/typedoc-packages-example/tree/master

I ran docs for each package and they were generated.

This should make things more canonical and less likely to break.


packages/shared/akamai-edgeworker-sdk/typedoc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "entryPoints": ["src/index.ts"],
+  "out": "docs",
+  "skipErrorChecking": true
```
kinyoklion, line 4: Typedoc cannot handle the package.json inclusion. This makes it ignore it.

PR #262, opened 2023-08-30
chore: Support migration payload contract tests.

kinyoklion:


contract-tests/sdkClientEntity.js
```diff
@@ -172,31 +172,47 @@ export async function newSdkClientEntity(options) {
           check: migrationOperation.trackConsistency ? (a, b) => a === b : undefined,
           readNew: async () => {
             try {
-              const res = await got.post(migrationOperation.newEndpoint, {});
+              const options = {};
+              if (migrationOperation.payload) {
+                options.body = migrationOperation.payload;
+              }
+              const res = await got.post(migrationOperation.newEndpoint, options);
```
yusinto, line 179: Looks like we can dry this logic with a function:

```ts
        const createMigrationOptions = ({ payload }) => {
          return payload ? { body: payload } : {};
        };
```

```suggestion
              const res = await got.post(
                migrationOperation.newEndpoint,
                createMigrationOptions(migrationOperation),
              );
```

yusinto (approved):

contract-tests/sdkClientEntity.js
kinyoklion: I actually realized that I messed this up.

louis-launchdarkly (approved):

PR #263, opened 2023-08-30
chore: Add initial documentation.

kinyoklion:


packages/shared/sdk-server/src/Migration.ts
```diff
@@ -39,13 +39,45 @@ async function safeCall<TResult>(
...
     success: true,
     result,
   };
 }
 
+/**
+ * Report a failed migration operation from `readNew`, `readOld`, `writeNew` or `writeOld`.
+ *
+ * ```
+ * readNew: async () => {
```
kinyoklion, line 69: This is the same sample, but I think it makes things easier for customers.

packages/shared/sdk-server/src/api/options/LDMigrationOptions.ts
```diff
@@ -90,27 +90,69 @@ export interface LDMigrationOptions<
...
+   * @param payload An optional payload. The payload is provided when calling the  `read` method on the migration.
+   * @returns The result of the operation. Use {@link LDMigrationSuccess} or {@link LDMigrationError} to create a suitable return value.
    */
   readNew: (payload?: TMigrationReadInput) => Promise<LDMethodResult<TMigrationRead>>;
 
   /**
-   * TKTK
+   * Implementation which provides a write to the "new" source.
+   *
+   * Users are required to provide two different write methods -- one to write to the old migration source, and one to
```
kinyoklion, line 111: There is a bit of duplication here, again because it makes finding the information a bit easier.

louis-launchdarkly (approved):

louis-launchdarkly:
Exciting that this is coming together!

PR #264, opened 2023-08-31
feat: Refactor variation method and consistency tracking.

kinyoklion:
Was reviewing my code and Mathew's and noticed some inconsistencies in how we approach the consistency check. Node correctly only did the check when sampled, but exposed more information than we want. So we have changed the two SDKs to be consistent with each other. So go now only samples when we want, and node will expose less information,

### DIFF
--- a/packages/shared/sdk-server/src/LDClientImpl.ts
+++ b/packages/shared/sdk-server/src/LDClientImpl.ts
@@ -17,9 +17,9 @@ import {
   LDClient,
   LDFlagsState,
   LDFlagsStateOptions,
-  LDMigrationDetail,
   LDMigrationOpEvent,
   LDMigrationStage,
+  LDMigrationVariation,
   LDOptions,
   LDStreamProcessor,
 } from './api';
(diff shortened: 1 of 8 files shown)


packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -307,13 +307,13 @@ export default class LDClientImpl implements LDClient {
...
     defaultValue: LDMigrationStage,
-  ): Promise<LDMigrationDetail> {
+  ): Promise<LDMigrationVariation> {
     const convertedContext = Context.fromLDContext(context);
     const [{ detail }, flag] = await this.evaluateIfPossible(
       key,
       context,
       defaultValue,
-      this.eventFactoryWithReasons,
+      this.eventFactoryDefault,
```
kinyoklion, line 316: Should not include reasons by default for the evaluation event associated with the migration. It will be in the migration_op event.

louis-launchdarkly (approved):

PR #270, opened 2023-09-06
fix: Use clientSideAvailability instead of clientSide for filtering client side flags.

kinyoklion:
The node-server-sdk was not using the newer `clientSideAvailability`, but `clientSide`. In porting to TS that logic was retained. This updates it to use the newer field.

### DIFF
--- a/packages/shared/akamai-edgeworker-sdk/src/__tests__/testData.json
+++ b/packages/shared/akamai-edgeworker-sdk/src/__tests__/testData.json
@@ -97,7 +97,7 @@
       "variations": [true, false],
       "clientSideAvailability": {
         "usingMobileKey": true,
-        "usingEnvironmentId": true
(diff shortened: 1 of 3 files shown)


louis-launchdarkly (approved):

packages/shared/akamai-edgeworker-sdk/src/__tests__/testData.json
```diff
@@ -97,7 +97,7 @@
       "variations": [true, false],
       "clientSideAvailability": {
         "usingMobileKey": true,
-        "usingEnvironmentId": true
+        "usingEnvironmentId": false
       },
       "clientSide": false,
```
kinyoklion, line 102: clientSide had been set false, but not usingEnvironmentId, these should be consistent.

PR #275, opened 2023-09-12
chore: part 1 - share streamer

yusinto:
This is part 1 of moving the StreamingProcessor to shared/common. [Part 2](https://github.com/launchdarkly/js-core/pull/276) contains unit test fixes.

PR #276, opened 2023-09-12
chore: part 2 - share streamer

yusinto:
This is part 2 of moving the StreamingProcessor to shared/common. This pr contains only unit test fixes. [Part 1](https://github.com/launchdarkly/js-core/pull/275) contains logic changes.


package.json
yusinto, line 1: These changes are from main, nothing added, nothing deleted.

packages/shared/sdk-server/__tests__/data_sources/StreamingProcessor.test.ts
yusinto, line 1: Note: this file has been moved to shared/common/internal/stream. It is only here to faciliate review. Will delete when review is done.

packages/shared/common/src/internal/stream/StreamingProcessor.test.ts
yusinto, line 1: Please see my comments about this file [here](https://github.com/launchdarkly/js-core/pull/276/files#r1326315616).

packages/shared/sdk-server/__tests__/LDClientImpl.test.ts
```diff
@@ -1,137 +1,100 @@
...
+    ...actual,
+    ...{
+      internal: {
+        ...actual.internal,
+        StreamingProcessor: actual.internal.mocks.MockStreamingProcessor,
       },
     },
-  );
-
-  client.close();
```
yusinto, line 55: These tests are moved to a `describe` block, you can see them starting from line 43 below.

packages/shared/sdk-server/__tests__/data_sources/StreamingProcessor.test.ts
```diff
@@ -1,298 +1,211 @@
...
+      expect.any(Function),
+    );
+    expect(mockEventSource.addEventListener).toHaveBeenNthCalledWith(
+      2,
+      'patch',
+      expect.any(Function),
+    );
   });
 
-  describe('when putting a message', () => {
```
yusinto, line 113: Listener logic tests are moved to `createStreamListeners.test.ts` so putting, patching, deleting tests here are removed here.

package.json
```diff
@@ -22,7 +22,7 @@
     "clean": "yarn workspaces foreach -pt run clean",
     "build": "yarn workspaces foreach -p --topological-dev run build",
     "//": "When using build:doc you need to specify the workspace. 'yarn run build:doc packages/shared/common' for example.",
-    "build:doc": "./scripts/build-doc.sh $1",
+    "build:doc": "npx typedoc --options $1/typedoc.json",
```
kinyoklion, line 25: Looks like maybe this PR was synced with main, but the base was not? Or github is doing strange stuff, which it does do sometimes.

packages/sdk/server-node/__tests__/LDClientNode.test.ts
```diff
@@ -11,18 +26,8 @@ it('fires ready event in offline mode', (done) => {
...
-        setTimeout(() => {
-          fn(new Error('BAD THINGS'));
-        }, 0);
-      },
-      stop: () => {},
-      close: () => {},
-      sendEvents: false,
-    },
-  });
+  setupMockStreamingProcessor(true);
```
kinyoklion, line 29: It seems the config still accepts an update processor, so it makes more sense to just make a mock processor an pass it than having to use a jestmock to override the default event processor.

package.json
yusinto: You're right i merged this from main manually, but not the base. I think this should be ok, if I do things correctly, the base will get this change and that will be in sync with main.

packages/sdk/server-node/__tests__/LDClientNode.test.ts
yusinto: Done.

kinyoklion (approved):

PR #284, opened 2023-09-25
fix: refactor mocks to its own project

yusinto:
Large pr but mostly trivial changes to import paths because mocks have been moved to its own private project `@launchdarkly/private-js-mocks`:

* Refactored mocks to its own project `@launchdarkly/private-js-mocks`.
* Better eslint rule for `no-unused-vars` to avoid disabling this rule sporadically.


louis-launchdarkly:
It may not be fully ready for review, but creating a separate package and the change within it looks generally good to me.

packages/mocks/LICENSE
```diff
@@ -0,0 +1,13 @@
+Copyright 2022 Catamorphic, Co.
```
kinyoklion, line 1: Because it is a new file we should put 2023 on this one.

kinyoklion:
Is this ready for review? I was added as a reviewer, but I do see it is draft.

yusinto:
Not ready yet I need to fix a broken tests and fix a circular dependency scenario which I want to improve.

packages/shared/mocks/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "@launchdarkly/private-js-mocks",
+  "private": true,
```
yusinto, line 3: * We are not publishing `mocks` to npm because it's internal use only.
* I'm open to improving the name of this package. It's the best I can come up with right now.

packages/shared/mocks/src/index.ts
```diff
@@ -10,6 +11,7 @@ export {
   crypto,
   logger,
   hasher,
+  ContextDeduplicator,
```
yusinto, line 14: There was a bug before this PR where this export was missing. This is to fix that bug.

packages/shared/mocks/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    "declarationMap": true, // enables importers to jump to source
+    "stripInternal": true,
+    "paths": {
+      "@common": ["../common"]
+    }
+  },
+  "exclude": ["**/*.test.ts", "dist", "node_modules", "__tests__"],
+  "references": [
+    {
+      "path": "../common"
```
yusinto, line 23: This enables us to import types from `common`. Then on line 17 above, we add an alias so we can import common  without relative paths to appease eslint.

packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -25,7 +23,7 @@ import DataSourceUpdates from './data_sources/DataSourceUpdates';
 import PollingProcessor from './data_sources/PollingProcessor';
 import Requestor from './data_sources/Requestor';
 import createDiagnosticsInitConfig from './diagnostics/createDiagnosticsInitConfig';
-import { allAsync, allSeriesAsync } from './evaluation/collection';
```
yusinto, line 28: `allSeriesAsync` is removed because it was unused.

packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -485,11 +481,9 @@ export default class LDClientImpl implements LDClient {
     this.variationInternal(flagKey, context, defaultValue, eventFactory, cb);
   }
 
-  private dataSourceErrorHandler(e: LDStreamingError | LDPollingError) {
```
yusinto, line 488: Relax the error type here to allow all kinds of errors to be handled. This allows mock errors to be handled.

packages/shared/mocks/tsconfig.json
kinyoklion: So this means that mocks specifically will only work when checked out as a monorepo? Or will these things get bundled into the dist?

kinyoklion (approved):
This is probably all good, but I think we are going to need to make some packages to be sure.

packages/shared/mocks/tsconfig.json
yusinto: Mocks will only work when checked out as monorepo. There's no bundling to produce any bundles for mocks. The dist folder only contains tsc output without bundling and does not contain any common code.

Just to re-iterate, the mocks projects are only using types from common and nothing else.

kinyoklion (approved):

PR #288, opened 2023-09-27
feat: Add typed variation methods.

### LINKED ISSUE #285: Typesafe API for TypeScript consumers
**Is your feature request related to a problem? Please describe.**

An API for TypeScript users to retrieve toggle values in a typesafe manner.

**Describe the solution you'd like**

Similar to the Java SDK, provide methods like `boolVariation` and `stringVariation` on `LDClient` for TypeScript users that only accept a default value of the correct type, and throw an error if the resolved value from LaunchDarkly is not of the correct type.

kinyoklion:
This adds typed variation methods. 

This works for both JS and TS, but may be especially nice for those that want to be more type safe in typescript.

resolves #285

### DIFF
--- a/contract-tests/index.js
+++ b/contract-tests/index.js
@@ -31,7 +31,8 @@ app.get('/', (req, res) => {
       'migrations',
       'event-sampling',
       'config-override-kind',
-      'metric-kind'
+      'metric-kind',
+      'strongly-typed',
     ],
   });
 });
--- a/contract-tests/sdkClientEntity.js
+++ b/contract-tests/sdkClientEntity.js
@@ -140,8 +140,7 @@ export async function newSdkClientEntity(options) {
           switch(pe.valueType) {
             case "bool":
               return {value: await client.boolVariation(pe.flagKey, pe.context || pe.user, pe.defaultValue)};
-            case "int":
-              return {value: await client.numberVariation(pe.flagKey, pe.context || pe.user, pe.defaultValue)};
+            case "int": // Intentional fallthrough.
             case "double":
               return {value: await client.numberVariation(pe.flagKey, pe.context || pe.user, pe.defaultValue)};
             case "string":
--- a/contract-tests/sdkClientEntity.js
+++ b/contract-tests/sdkClientEntity.js
@@ -127,8 +127,7 @@ export async function newSdkClientEntity(options) {
           switch(pe.valueType) {
             case "bool":
               return await client.boolVariationDetail(pe.flagKey, pe.context || pe.user, pe.defaultValue);
-            case "int":
-              return await client.numberVariationDetail(pe.flagKey, pe.context || pe.user, pe.defaultValue);
+            case "int": // Intentional fallthrough.
             case "double":
               return await client.numberVariationDetail(pe.flagKey, pe.context || pe.user, pe.defaultValue);
             case "string":
--- a/contract-tests/sdkClientEntity.js
+++ b/contract-tests/sdkClientEntity.js
@@ -124,10 +124,32 @@ export async function newSdkClientEntity(options) {
       case 'evaluate': {
         const pe = params.evaluate;
         if (pe.detail) {
-          return await client.variationDetail(pe.flagKey, pe.context || pe.user, pe.defaultValue);
+          switch(pe.valueType) {
+            case "bool":
+              return await client.boolVariationDetail(pe.flagKey, pe.context || pe.user, pe.defaultValue);
+            case "int":
+              return await client.numberVariationDetail(pe.flagKey, pe.context || pe.user, pe.defaultValue);
+            case "double":
+              return await client.numberVariationDetail(pe.flagKey, pe.context || pe.user, pe.defaultValue);
+            case "string":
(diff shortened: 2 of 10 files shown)


contract-tests/index.js
```diff
@@ -31,7 +31,8 @@ app.get('/', (req, res) => {
       'migrations',
       'event-sampling',
       'config-override-kind',
-      'metric-kind'
+      'metric-kind',
+      'strongly-typed',
```
kinyoklion, line 35: Run the typed contract tests.

contract-tests/sdkClientEntity.js
```diff
@@ -124,10 +124,32 @@ export async function newSdkClientEntity(options) {
       case 'evaluate': {
         const pe = params.evaluate;
         if (pe.detail) {
-          return await client.variationDetail(pe.flagKey, pe.context || pe.user, pe.defaultValue);
+          switch(pe.valueType) {
+            case "bool":
+              return await client.boolVariationDetail(pe.flagKey, pe.context || pe.user, pe.defaultValue);
+            case "int":
```
kinyoklion, line 130: The contract tests support int and double, but we are using a single numberVariation.

packages/shared/sdk-server/__tests__/LDClient.evaluation.test.ts
```diff
@@ -160,6 +160,99 @@ describe('given an LDClient with test data', () => {
     const valueB = await client.variation('my-feature-flag-1', userContextObject, 'default');
     expect(valueB).toEqual(true);
   });
+
+  it('evaluates with jsonVariation', async () => {
```
kinyoklion, line 164: This just addresses the problem with `any` in Typescript. Some people require using `unknown` so they have to explicitly cast.

packages/shared/sdk-server/__tests__/LDClient.evaluation.test.ts
```diff
@@ -160,6 +160,99 @@ describe('given an LDClient with test data', () => {
...
+      defaultUser,
+      false,
+    )) as number;
+    expect(numericRes).toBe(62);
+
+    td.update(td.flag('flagkey').valueForAll('potato'));
+    const stringRes: string = (await client.jsonVariation('flagkey', defaultUser, false)) as string;
+    expect(stringRes).toBe('potato');
+  });
+
```
kinyoklion, line 181: Below are all the typed method calls.

packages/shared/sdk-server/__tests__/LDClient.migrations.test.ts
```diff
@@ -57,7 +57,7 @@ describe('given an LDClient with test data', () => {
       const defaultValue = Object.values(LDMigrationStage).find((item) => item !== value);
       // Verify the pre-condition that the default value is not the value under test.
       expect(defaultValue).not.toEqual(value);
-      const res = await client.variationMigration(
+      const res = await client.migrationVariation(
```
kinyoklion, line 60: I decided to make this the same as the other SDKs with this change. This has not been released yet (for Yus).

packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -301,7 +303,105 @@ export default class LDClientImpl implements LDClient {
...
+          };
+          resolve(typedRes);
+        },
+        typeChecker,
+      );
+    });
+  }
+
+  async boolVariation(key: string, context: LDContext, defaultValue: boolean): Promise<boolean> {
+    return (
```
kinyoklion, line 333: The approach taken with typedEval is to have it take a validator that checks the type and also indicates what the type was, for error messages. This matched nicely with the TypeValidators implementation. 

If the type is wrong, then it generates the correct reason and evaluation events.

contract-tests/sdkClientEntity.js
```diff
@@ -124,10 +124,32 @@ export async function newSdkClientEntity(options) {
...
+              return await client.numberVariationDetail(pe.flagKey, pe.context || pe.user, pe.defaultValue);
+            case "double":
+              return await client.numberVariationDetail(pe.flagKey, pe.context || pe.user, pe.defaultValue);
```
yusinto, line 133:
```suggestion
            case "int":
            case "double":
              return await client.numberVariationDetail(pe.flagKey, pe.context || pe.user, pe.defaultValue);
```

contract-tests/sdkClientEntity.js
```diff
@@ -124,10 +124,32 @@ export async function newSdkClientEntity(options) {
...
         } else {
-          const value = await client.variation(pe.flagKey, pe.context || pe.user, pe.defaultValue);
-          return { value };
+          switch(pe.valueType) {
+            case "bool":
+              return {value: await client.boolVariation(pe.flagKey, pe.context || pe.user, pe.defaultValue)};
+            case "int":
+              return {value: await client.numberVariation(pe.flagKey, pe.context || pe.user, pe.defaultValue)};
+            case "double":
+              return {value: await client.numberVariation(pe.flagKey, pe.context || pe.user, pe.defaultValue)};
```
yusinto, line 147: You can waterfall these too like above.

yusinto (approved):

PR #289, opened 2023-10-02
feat: Remove index and custom event sampling.

kinyoklion:


packages/shared/common/src/internal/events/EventProcessor.ts
```diff
@@ -251,14 +251,14 @@ export default class EventProcessor implements LDEventProcessor {
...
 
-    if (addIndexEvent && shouldSample(inputEvent.indexSamplingRatio)) {
+    if (addIndexEvent) {
       this.enqueue(
         this.makeOutputEvent(
           {
             kind: 'index',
             creationDate: inputEvent.creationDate,
             context: inputEvent.context,
-            samplingRatio: inputEvent.indexSamplingRatio,
```
kinyoklion, line 261: I kept the sampling ratio around, but it only gets output if it isn't 1, so it will not get output.

keelerm84 (approved):

PR #291, opened 2023-10-02
fix: yarn topological dev build

yusinto:
CI failed because `yarn workspaces foreach -pR --topological-dev --from '@launchdarkly/xxx' run build`. This is because `topological-dev` instructs yarn to build all deps (prod & dev) prior to running build in the current workspace. Since `mocks` need types from `common` this fails.

This pr adds a `build-types` command to common which gets run prior to mocks `build` command to ensure common types exist for mocks to be built successfully.

### DIFF
--- a/.github/workflows/mocks.yml
+++ b/.github/workflows/mocks.yml
@@ -1,4 +1,4 @@
-name: sdk/cloudflare
+name: shared/mocks
 
 on:
   push:
--- a/.github/workflows/mocks.yml
+++ b/.github/workflows/mocks.yml
@@ -22,3 +22,4 @@ jobs:
         with:
           workspace_name: '@launchdarkly/private-js-mocks'
           workspace_path: packages/shared/mocks
+          should_build_docs: false
--- a/actions/ci/action.yml
+++ b/actions/ci/action.yml
@@ -40,4 +40,5 @@ runs:
 
     - name: Build Docs
       shell: bash
+      if: ${{inputs.workspace_name != '@launchdarkly/private-js-mocks'}}
       run: yarn build:doc -- ${{ inputs.workspace_path }}
--- a/actions/ci/action.yml
+++ b/actions/ci/action.yml
@@ -11,7 +11,9 @@ inputs:
   workspace_path:
     description: 'Path to the package to release.'
     required: true
-
+  should_build_docs:
+    description: 'Whether docs should be built. It will be by default.'
+    default: true
 runs:
   using: composite
   steps:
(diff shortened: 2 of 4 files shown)


louis-launchdarkly:
@yusinto, the change looks fine to me. Do you expect the build to still have issues?

kinyoklion:
I think that the name for build test mocks must have a problem? 
image
Showing under cloudflare?

kinyoklion (changes requested):
Should build once it has a typedoc file. Currently all packages expect to build docs.

.github/workflows/mocks.yml
```diff
@@ -1,4 +1,4 @@
-name: sdk/cloudflare
```
yusinto, line 1: Sorry this was incorrectly set as cloudflare @louis-launchdarkly . I have fixed this now.

actions/ci/action.yml
```diff
@@ -40,4 +40,5 @@ runs:
 
     - name: Build Docs
       shell: bash
+      if: ${{inputs.workspace_name != '@launchdarkly/private-js-mocks'}}
```
kinyoklion, line 43: If we are going to do this, then we should parameterize it. An optional input to disable docs.

kinyoklion (changes requested):

actions/ci/action.yml
yusinto, line 43: I didn't want to add a redundant `mocks/typedoc.json` so I'm doing this instead. I think this is ok and the `publish-docs` step should be ok too. @kinyoklion Are you able to confirm please?

kinyoklion: Yeah, I think we should not encode information about packages into the shared code. Adding an optional input, like `disable_docs` or `docs_enabled` with the correct default would be fine. Publish will be fine as long as you aren't trying to release this package.

kinyoklion (changes requested):
I had re-requested changes already, but it looks like my review was re-requested afterward.

actions/ci/action.yml
```diff
@@ -11,7 +11,9 @@ inputs:
   workspace_path:
     description: 'Path to the package to release.'
     required: true
-
+  should_build_docs:
+    description: 'Whether docs should be built. It will be by default.'
+    default: true
```
yusinto, line 16: As requested @kinyoklion

kinyoklion: I really wish the CI was running for these PRs so I could be sure. 

I added a rule that any branch that starts with `feat/` will run all checks when PR'd against. When we have cumulative ones like this, we will probably want to leverage that.

As is, I will approve this PR, and then we will need to check that the docs steps are running correctly on the other.

My main apprehension is about the types. I think that it will always be a 'string' in this case and 'true' is what we want. But the builds will tell us for sure.

kinyoklion (approved):

PR #292, opened 2023-10-03
chore: Disable diagnostic events when running contract tests.

kinyoklion:
I have done some investigation and found that the intermittent issues with contract tests seem to be related to how the test harness handles them. We do not actually test them using the test harness, so I am disabling them now to improve CI and release robustness.

I've basically been running the contract tests in a loop. Usually they fail in a few hours, but with diagnostic opt out they ran overnight.

I was testing them in combination with some other robustness improvements, but I am re-testing now with just this change. Those changes can be independent. 

Once we know what is going on the test harness side we should be able to re-enable these, but it likely isn't worth doing unless there are some tests for them.

### DIFF
--- a/contract-tests/sdkClientEntity.js
+++ b/contract-tests/sdkClientEntity.js
@@ -9,6 +9,7 @@ export { badCommandError };
 export function makeSdkConfig(options, tag) {
   const cf = {
     logger: sdkLogger(tag),
+    diagnosticOptOut: true
   };
   const maybeTime = (seconds) =>
     seconds === undefined || seconds === null ? undefined : seconds / 1000;


contract-tests/sdkClientEntity.js
```diff
@@ -9,6 +9,7 @@ export { badCommandError };
 export function makeSdkConfig(options, tag) {
   const cf = {
     logger: sdkLogger(tag),
+    diagnosticOptOut: true
   };
   const maybeTime = (seconds) =>
     seconds === undefined || seconds === null ? undefined : seconds / 1000;
```
yusinto, line 15: Unrelated, but this can be simplified:

```suggestion
    !seconds ? undefined : seconds / 1000;
```

yusinto (approved):

contract-tests/sdkClientEntity.js
kinyoklion: 0 seconds is falsey.

kinyoklion:
```
> let seconds = 0;
undefined
> !seconds ? undefined : seconds/1000
undefined
> seconds === undefined || seconds === null ? undefined : seconds/1000
0
```

I do not know if our test execution does any tests where it sets to 0, but if it does, then the results would be substantially different. As undefined will get default times, and 0 may or may not depending on the check.

kinyoklion: Now that we are post 8.0 I should convert this to typescript and I could make things like this at least look a little cleaner. This is still the original implementation from Eli, because I wanted to validate compatibility originally.

PR #294, opened 2023-10-05
chore: implement initial flag fetch

yusinto:
First attempt to implement an initial flag fetch followed by emitting events. I also added comments like this:

```tsx
Dom api usage: xxx
```

There are three right now: fetch, btoa and EventTarget. I left comments in the code for react native how to deal with these.


packages/shared/sdk-client/src/LDClientDomImpl.ts
```diff
@@ -2,37 +2,72 @@
...
+
+  async start() {
+    try {
+      this.flags = await fetchFlags(this.sdkKey, this.context, this.config, this.platform);
+      this.emitter.emit('ready');
+    } catch (error: any) {
+      this.logger.error(error);
+      this.emitter.emit('error', error);
+      this.emitter.emit('failed', error);
+    }
```
yusinto, line 70: This is new. This decouples the construction of the ldClient object and the initial poll. The decoupling gives consuming sdks more flexibility how they want initialization to proceed.

packages/shared/sdk-client/src/api/LDEmitter.ts
```diff
@@ -11,15 +17,19 @@ export type EventName = 'change' | 'internal-change' | 'ready' | 'initialized' |
...
    * @param name string event name
-   * @param listener function to handle the event
+   * @param originalListener pointer to the original function as specified by
+   * the consumer
+   * @param customListener pointer to the custom function based on original
+   * listener. This is needed to allow for CustomEvents.
    * @private
    */
-  private saveListener(name: EventName, listener: EventListener) {
+  private saveListener(name: EventName, originalListener: Function, customListener: Function) {
```
yusinto, line 31: We need to save both the original listener and the custom listener in order to remove them both correctly from our own internal map and the javascript EventTarget object respectively.

packages/shared/sdk-client/src/api/LDOptions.ts
```diff
@@ -98,7 +98,7 @@ export default interface LDOptions {
    * {@link LDClientDom.variationDetail} method. Since this increases the size of network requests,
    * such information is not sent unless you set this option to true.
    */
-  evaluationReasons?: boolean;
+  withReasons?: boolean;
```
yusinto, line 101: This is a breaking api change. I'm making this consistent with the server sdks which I think is a good idea.

packages/shared/sdk-client/src/configuration/Configuration.ts
```diff
@@ -12,7 +13,7 @@ import type LDOptions from '../api/LDOptions';
 import validators from './validators';
 
 export default class Configuration {
-  public static DEFAULT_POLLING = 'https://clientsdk.launchdarkly.com';
+  public static DEFAULT_POLLING = 'https://sdk.launchdarkly.com';
```
yusinto, line 16: According to this [confluence page](https://launchdarkly.atlassian.net/wiki/spaces/PD/pages/2129461878/Service+endpoints+used+by+SDKs), this is added as part of u2c and so should be the one used. I think either will work.

packages/shared/sdk-client/src/evaluation/fetchUtils.test.ts
yusinto, line 1: Apologies, this is not done yet, but most of the functionality in this utils class is tested by its consumer, which is fetchFlags. I will get to this soon, in another PR.

packages/shared/sdk-client/src/platform/PlatformDom.ts
yusinto, line 1: I don't see the need for a dom specific platform for now, so I have deleted this.

kinyoklion (changes requested):
Ok. We need to determine a different base for you do use, because that base you have chosen has already been merged together with my features. So if you merge there, then you will have the mother of all merge conflicts when you want to eventually merge to main.

`feat/merge-client-and-migrations` contains all of our changes combined, and if what I was going to merge when we are ready to release migrations.

If these changes don't hurt anything, then you could base it on `feat/merge-client-and-migration`, and then it will get merged to main when we release migrations.

If it needs to be held back, then we should make a new `feat/` branch.

I'll look at other things, but I wanted to get the base addressed.

packages/shared/sdk-client/package.json
```diff
@@ -49,6 +50,7 @@
     "jest": "^29.6.1",
     "jest-diff": "^29.6.1",
     "jest-environment-jsdom": "^29.6.1",
+    "jest-fetch-mock": "^3.0.3",
```
kinyoklion, line 53: This has not been published in 4 years, which makes me pretty nervous.

packages/shared/sdk-client/src/platform/PlatformDom.ts
kinyoklion: I am not ok with not using a platform. Fetch, response, crypto need to go through a platform. It is extra work, but it is important work that makes this code useful and flexible. 

What if we need to do fetch different in node client or electron? What if we need a bun client SDK? What if we need some strange container environment client SDK? What if we need to use filesystem APIs for a custom platform?

The client for DOM is really easy to implement because you just call actual fetch, for instance. Make a client SDK, make it take a platform, now we have set ourselves up to not need polyfills and to have a flexible and extensible base client SDK.

kinyoklion: For insance, node 16 is still supported, the client SDK can use the current server platform with a little movement of code. Node 16 only has experimental fetch, so we would need to polyfill. The platform is the most fundamental part of the typescript re-write. Node 16 is also the basis of electron 22, the oldest currently supported electron.

kinyoklion: Ah, yeah, platform dom needs to exist, but we should never use it in shared code. I misread this the first time and made a different comment.

sdk-client should work for any possible client platform. So it cannot have knowledge of platform dom.

But platform dom, or something, would need to exist for the browser SDK. We need to hide `fetch` for instance, behind a platform. For me this is an immutable requirement.

For instance when we make the node client SDK, we should hoist the node platform out of server and share it. But it probably should be a different platform than browser. If we support node 16, which is currently an LTS, then we shouldn't use actual browser fetch as it was experimental.

The sdk-server only uses Platform, and then implementors can use more specific platforms.

packages/shared/sdk-client/src/evaluation/fetchFlags.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+const fetchFlags = async (
+  sdkKey: string,
+  context: LDContext,
+  config: Configuration,
+  platform: Platform,
+): Promise<Flags> => {
+  const fetchUrl = createFetchUrl(sdkKey, context, config);
+  const fetchOptions: RequestInit = createFetchOptions(sdkKey, context, config, platform.info);
+  const response = await fetch(fetchUrl, fetchOptions);
```
kinyoklion, line 30: This needs to come from the platform.

kinyoklion (changes requested):

packages/shared/sdk-client/src/platform/PlatformDom.ts
kinyoklion: A concrete real example is that electron and node client SDKs should support proxies using the proxy agent. We would not want this code to be in `sdk-client` it would be handled by the platform fetch and not affect browser at all.

yusinto:
Ok I've merged changes from `feat/merge-client-and-migrations` and changed the PR root. Everything seems to be fine.

packages/shared/sdk-client/src/platform/PlatformDom.ts
yusinto: Ok I'll refactor fetch to use Platform.Requests.

packages/shared/sdk-client/src/LDClientDomImpl.ts
kinyoklion, line 1: We should remove Dom from the name of this one. Can it just be LDClientImpl?

yusinto: Done

kinyoklion (approved):

PR #296, opened 2023-10-09
chore: re-add PlatformDom and refactor btoa to platform

yusinto:


contract-tests/sdkClientEntity.js
yusinto, line 1: Prettier did this, no functionality changes.

packages/shared/mocks/src/platform.ts
```diff
@@ -1,7 +1,11 @@
-import type { Info, Platform, PlatformData, Requests, SdkData } from '@common';
+import type { Base64, Info, Platform, PlatformData, PlatformDom, Requests, SdkData } from '@common';
 
 import { crypto } from './hasher';
 
+const base64: Base64 = {
+  btoa: (s: string) => Buffer.from(s).toString('base64'),
```
yusinto, line 6: Using nodejs `Buffer` api which should be ok because this should only be run in unit tests.

packages/shared/common/src/api/platform/Platform.ts
```diff
@@ -26,3 +27,10 @@ export interface Platform {
    */
   requests: Requests;
 }
+
+export interface PlatformDom extends Platform {
```
kinyoklion, line 31: This PlatformDom is still problematic.

Leaf nodes know about their platform, but the trunk shouldn't. So if we need btoa, then we would need it in `Platform`. 
In this case it doesn't matter at all that this might be used in DOM. We might need btoa in any client SDK. 

I do think there is a question about what we do for client/server or APIs that aren't needed everywhere. So, currently server SDKs don't need this functionality. So we have a couple options, we could make it optional, we could make it required and then just implement it for server SDKs. Or we could keep this, but call it `ClientPlatform` or something. `Platform` for APIs used by all SDKs, `PlatformClient/ClientPlatform` for client-side only required APIs.

Generally speaking I prefer just `Platform` and then we either make it optional or required. The reason being is I would prefer to re-use the same implementation for the node client SDK as we currently use in the node server SDK. 

Something called PlatformDom could potentially be in the js-client-sdk, assuming that is the SDK that replaces the current js-client-sdk.

yusinto: ok ill refactor again.

packages/shared/common/src/api/platform/Base64.ts
```diff
@@ -0,0 +1,3 @@
+export interface Base64 {
```
kinyoklion, line 1: Maybe `Encoding` instead of `Base64`?

packages/shared/common/src/api/platform/Platform.ts
kinyoklion: Given where we are right now, and the timing of things, I would just recommend it be optional. Then you just check that it is present in client common ("Client SDKs require the encoding platform" or something like that). That way we don't have to figure out more implementations.

packages/shared/mocks/src/platform.ts
kinyoklion: Sounds good.

packages/shared/common/src/api/platform/Base64.ts
yusinto: Done.

packages/shared/common/src/api/platform/Platform.ts
```diff
@@ -1,9 +1,15 @@
...
+import { Encoding } from './Encoding';
 import { Filesystem } from './Filesystem';
 import { Info } from './Info';
 import { Requests } from './Requests';
 
 export interface Platform {
+  /**
+   * The interface for performing base64 operations.
+   */
+  encoding: Encoding;
```
kinyoklion, line 11:
```suggestion
  encoding?: Encoding;
```

So we don't have to implement it for server side right now.

packages/shared/common/src/api/platform/Platform.ts
```diff
@@ -1,9 +1,15 @@
 import { Crypto } from './Crypto';
+import { Encoding } from './Encoding';
 import { Filesystem } from './Filesystem';
 import { Info } from './Info';
 import { Requests } from './Requests';
 
 export interface Platform {
+  /**
+   * The interface for performing base64 operations.
```
kinyoklion, line 9:
```suggestion
   * The interface for performing encoding operations.
```

kinyoklion (approved):

PR #298, opened 2023-10-12
chore: implement variation functions

yusinto:
* Share EventFactory and friends
* Implement variation, typedVariation, allFlags and other functions for sdk-client


packages/shared/common/src/internal/evaluation/ErrorKinds.ts
yusinto, line 1: One of 4 files moved from sdk-server:
* ErrorKinds
* EvalResult
* Reason
* ClientMessages

packages/shared/common/src/internal/evaluation/EventFactoryBase.ts
yusinto, line 1: This is mostly ripped from the previous `EventFactory` implementation. All functions are identical as previously, except `evalEvent`. Each sdk-server and sdk-client implements their own copy of evalEvent, and then calling this base function. There maybe a better way to achieve this in typescript, and I'm open for suggestions.

packages/shared/sdk-client/src/LDClientImpl.test.ts
yusinto, line 1: Just rudimentary tests so not complete coverage yet. I'll add more in future prs.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -75,26 +86,36 @@ export default class LDClientImpl implements LDClient {
   }
 
   allFlags(): LDFlagSet {
-    return {};
+    const result: LDFlagSet = {};
+    Object.entries(this.rawFlags).forEach(([k, r]) => {
+      result[k] = r.value;
+    });
+    return result;
```
yusinto, line 93: This is different from the existing js-sdk allFlags. Here I am following the server sdk to not send events on allFlags, thus getting rid of `sendEventsOnlyForVariation`. I don't understand the use case for sending events in allFlags, and I am pushing for a consistent behavior with the server sdk.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -75,26 +86,36 @@ export default class LDClientImpl implements LDClient {
...
   }
 
-  close(onDone?: () => void): Promise<void> {
-    return Promise.resolve(undefined);
+  close(): void {
+    this.eventProcessor.close();
   }
```
yusinto, line 98: Breaking api change here, again I am pushing for consistency with the server sdk by getting rid of the callback arg and returning void instead of a promise.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -75,26 +86,36 @@ export default class LDClientImpl implements LDClient {
...
-  flush(onDone?: () => void): Promise<void> {
-    return Promise.resolve(undefined);
+  async flush(callback?: (err: Error | null, res: boolean) => void): Promise<void> {
+    try {
+      await this.eventProcessor.flush();
+    } catch (err) {
+      callback?.(err as Error, false);
+    }
+    callback?.(null, true);
   }
```
yusinto, line 107: This is a duplicate of sdk-server. We may be able to refactor this plus other similar functions to an LDClientImplBase class (a mouthful).

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -106,28 +127,115 @@ export default class LDClientImpl implements LDClient {
...
+    const checkedContext = Context.fromLDContext(this.context);
+    if (!checkedContext.valid) {
+      this.logger?.warn(ClientMessages.missingContextKeyNoEvent);
+      return;
+    }
+
+    this.eventProcessor.sendEvent(
+      this.eventFactoryDefault.customEvent(key, checkedContext!, data, metricValue),
+    );
+  }
```
yusinto, line 144: Another function which is almost a carbon copy of sdk-server, differing only in the LDContext argument.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -106,28 +127,115 @@ export default class LDClientImpl implements LDClient {
...
+
+  private sendEvalEvent(
```
yusinto, line 146: And another

packages/shared/common/src/internal/evaluation/ErrorKinds.ts
kinyoklion: Guess it is good I haven't released things yet. I would have had to major version everything again for these.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -106,28 +127,115 @@ export default class LDClientImpl implements LDClient {
...
+  ) {
+    evalRes.events?.forEach((event) => {
+      this.eventProcessor.sendEvent({ ...event });
+    });
+    this.eventProcessor.sendEvent(
+      eventFactory.evalEventClient(flagKey, rawFlag, evalContext, evalRes.detail, defaultValue),
+    );
+  }
+
+  private variationInternal(
```
yusinto, line 162: The logic here is copied from sdk-server, however this version is synchronous so does not have callbacks like the server counterpart. Still a lot of opportunities for dry, but I think this is good enough for now.

packages/shared/sdk-client/src/evaluation/fetchFlags.ts
```diff
@@ -3,27 +3,31 @@ import { LDContext, LDEvaluationReason, LDFlagValue, Platform } from '@launchdar
...
   trackEvents: boolean;
+  trackReason?: boolean;
   reason?: LDEvaluationReason;
+  debugEventsUntilDate?: number;
 };
 
-export type Flags = {
-  [k: string]: Flag;
+export type RawFlags = {
+  [k: string]: RawFlag;
```
yusinto, line 18: There maybe better names than `RawFlag`. I'm open.

Also, I am contemplating implementing this as a Map rather than an object. There's advantages to using Map in terms of performance and better support for removing/updating/iterating.

packages/shared/common/src/internal/evaluation/Reasons.ts
```diff
@@ -1,9 +1,7 @@
-import { LDEvaluationReason } from '@launchdarkly/js-sdk-common';
+import { LDEvaluationReason } from '../../api';
 
 /**
  * A set of static evaluation reasons and methods for creating specific reason instances.
- *
- * @internal
  */
 export default class Reasons {
```
kinyoklion, line 6: So, a server SDK is creating reasons. It does an eval and can have a rule match or a fallthrough, etc. A client SDK does not do these things.

How is this going to be used?

packages/shared/common/src/internal/evaluation/EvalResult.ts
```diff
@@ -1,16 +1,14 @@
-import { internal, LDEvaluationDetail, LDEvaluationReason } from '@launchdarkly/js-sdk-common';
-
+import { LDEvaluationDetail, LDEvaluationReason } from '../../api';
```
kinyoklion, line 1: I am a little hesitant on this file. Server SDKs do an eval, so they always produce details and reasons, but client SDKs either have reasons or do not have reasons based on how they were initialized and maybe some other factors. So, for typing I would prefer they remain required in server, but tney are going to need to be optional for clients.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -75,26 +86,36 @@ export default class LDClientImpl implements LDClient {
...
 
-  close(onDone?: () => void): Promise<void> {
-    return Promise.resolve(undefined);
+  close(): void {
+    this.eventProcessor.close();
   }
 
-  flush(onDone?: () => void): Promise<void> {
-    return Promise.resolve(undefined);
+  async flush(callback?: (err: Error | null, res: boolean) => void): Promise<void> {
```
kinyoklion, line 100: Do we need to keep the callback, or can we just use a promise instead?

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -75,26 +86,36 @@ export default class LDClientImpl implements LDClient {
...
+      await this.eventProcessor.flush();
+    } catch (err) {
+      callback?.(err as Error, false);
+    }
+    callback?.(null, true);
   }
 
   getContext(): LDContext {
-    return { kind: 'user', key: 'test-context-1' };
+    return { ...this.context };
```
kinyoklion, line 110: This is problematic. Destructing here only decouples the top level entries. The nested items within the context would still be shared with the original. 

It looks like the old implementation did a proper clone (js-sdk-common).

You could clone using `cloneWithRedactions` and passing an empty private attribute list.

packages/shared/sdk-client/src/api/LDClient.ts
```diff
@@ -263,14 +263,6 @@ export interface LDClient {
...
-   * @param onDone
-   *   A function which will be called when the operation completes. If omitted, you
-   *   will receive a Promise instead.
-   *
-   * @returns
-   *   If you provided a callback, then nothing. Otherwise, a Promise which resolves once
-   *   closing is finished. It will never be rejected.
    */
-  close(onDone?: () => void): Promise<void>;
+  close(): void;
```
kinyoklion, line 267: So, most SDKs flush inside close and then the close resolves once the flush is complete. I am not sure why it was not originally done on JS SDKs. It causes a lot of problems though. 

I still have not done anything about it in node v9 though.

packages/shared/sdk-client/src/evaluation/fetchFlags.ts
kinyoklion: What motivated changing from `Flags` to `RawFlags`?

In regards to performance using a map will probably be better. But one thing to consider is if that performance improvement exceeds the performance cost of initially creating a map. 

For long running things it will for sure, but for ephemeral web pages is may be slower.

kinyoklion: It is internal though, so it can easily change.

yusinto: The term Flag is overloaded. For the sdk-client, I want to be able to understand that this is the server response object, pristine. Any ideas? I'll revert to Flag if that's preferred.

packages/shared/sdk-client/src/api/LDClient.ts
yusinto: Good point. I'll change this to async with the flush call.

packages/shared/sdk-client/src/LDClientImpl.ts
yusinto: Ok good catch object spread only does first level clone. I'll fix this.

packages/shared/sdk-client/src/evaluation/fetchFlags.ts
kinyoklion: I kind of view the thing from LD as the "Flag". Then subsequently everything else isn't. 

Down stream we have:
- Flag values
- Evaluation details

I could see "FlagConfig" potentially as well, though it is maybe even less config on the client side.

packages/shared/common/src/internal/evaluation/EvalResult.ts
yusinto: Ok I understand your concerns. I re-visited sdk-client's variationInternal and I think I can accommodate your request. I'll put refactor and submit a secondary PR. This should address your other comment re Reasons as well.

packages/shared/common/src/internal/evaluation/Reasons.ts
yusinto: I will [refactor](https://github.com/launchdarkly/js-core/pull/298#discussion_r1357528663).

packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -636,7 +633,7 @@ export default class LDClientImpl implements LDClient {
                 `Did not receive expected type (${type}) evaluating feature flag "${flagKey}"`,
                 defaultValue,
               );
-              this.sendEvalEvent(evalRes, eventFactory, flag, evalContext, defaultValue);
+              this.sendEvalEvent(errorRes, eventFactory, flag, evalContext, defaultValue);
```
yusinto, line 636: This seems to be a bug. @kinyoklion can you confirm pls?

kinyoklion: Yes, we should change this. 

Originally I was not sure if the event should contain the original eval result, or the one with the WRONG_TYPE error. After some auditing we decided on using the error result, but then I totally forgot about it.

kinyoklion: Thank you for catching it. This is a new feature, so thankfully not released.

packages/shared/common/src/internal/events/InputEvalEvent.ts
```diff
@@ -40,9 +40,8 @@ export default class InputEvalEvent {
     public readonly samplingRatio: number = 1,
   ) {
     this.creationDate = Date.now();
+    this.value = value;
     this.default = defValue;
-    this.variation = detail.variationIndex ?? undefined;
-    this.value = detail.value;
```
yusinto, line 45: I removed `detail` because it seems we are only

packages/shared/common/src/internal/events/InputEvalEvent.ts
```diff
@@ -28,8 +28,8 @@ export default class InputEvalEvent {
     public readonly withReasons: boolean,
     public readonly context: Context,
     public readonly key: string,
+    value: LDFlagValue,
     defValue: any, // default is a reserved keyword in this context.
-    detail: LDEvaluationDetail,
     version?: number,
     variation?: number,
```
yusinto, line 34: I removed detail because we are only using it for variationIndex and value. We already have `variation` as argument online 34 so it seems redundant to have both variaionIndex and variation. I think it's safe to replace `detail` with just `value` and then use `variation` as is.

packages/shared/sdk-client/src/events/EventFactory.ts
```diff
@@ -0,0 +1,33 @@
...
+  evalEventClient(
+    flagKey: string,
+    value: LDFlagValue,
+    defaultVal: any,
+    flag: RawFlag,
+    context: Context,
+    reason?: LDEvaluationReason,
+  ): internal.InputEvalEvent {
+    const { trackEvents, debugEventsUntilDate, version, variation } = flag;
+    const addExperimentData = flag.trackReason || !!reason?.inExperiment;
```
yusinto, line 18: I'm not sure if `inExperiment` exists in client sdk's reason object. @kinyoklion @louis-launchdarkly do you know?

packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -661,7 +659,7 @@ export default class LDClientImpl implements LDClient {
       this.eventProcessor.sendEvent({ ...event });
     });
```
yusinto, line 660: @kinyoklion are you able to explain what populates the `evalRes.events` array please? I can't find any references to this. I need to understand this and apply to sdk-client however I don't see anything that pushes to this array.

packages/shared/sdk-client/src/events/EventFactory.ts
kinyoklion: It will conditionally exist. We didn't update all the client SDKs when it was added, because they can use `trackReason`. I did add it for inspections I thought, or maybe I just added the ticket.

packages/shared/sdk-server/src/LDClientImpl.ts
kinyoklion: The server SDK populates it with prerequisite events during evaluation.

If you evaluate a flag A that depends on B and C, then events need generated for B and C.

```
      this.evaluateInternal(
        flag,
        context,
        state,
        [],
        (res) => {
          if (state.bigSegmentsStatus) {
            res.detail.reason = {
              ...res.detail.reason,
              bigSegmentsStatus: state.bigSegmentsStatus,
            };
          }
          res.events = state.events; //  <---- ASSIGNED HERE
          resolve(res);
        },
        eventFactory,
      );
    });
```

This part handles prerequisite evaluations
```
          this.evaluateInternal(
            prereqFlag,
            context,
            state,
            updatedVisitedFlags,
            (res) => {
              // eslint-disable-next-line no-param-reassign
              state.events = state.events ?? [];

              if (eventFactory) {
                state.events.push(
                  eventFactory.evalEvent(prereqFlag, context, res.detail, null, flag), // <-- HERE
                );
              }
```

packages/shared/sdk-client/src/events/EventFactory.ts
kinyoklion: To clarify server SDKs put this in bootstrap data and the LD back-end will include it as well. It will only exist when it is true, and it will be undefined if it would have been false.

Logically it doesn't matter to client SDKs, because trackEvents will also be true. But it can be useful for application developers to know that it was in an experiment.

packages/shared/sdk-client/src/evaluation/fetchFlags.ts
yusinto: ok i'll revert back to Flags. Thank you.

packages/shared/sdk-client/src/events/EventFactory.ts
yusinto: I understand now. I'm going to remove this logical or because it only complicates reading the code. If trackReason is logically equivalent and sufficient, I prefer a terse simple code.

kinyoklion: Yeah, that should be good for client. (In server it does matter)

packages/shared/sdk-server/src/LDClientImpl.ts
yusinto: This is irrelevant to sdk-client so I'll omit this.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -75,26 +88,37 @@ export default class LDClientImpl implements LDClient {
...
+      await this.eventProcessor.flush();
+    } catch (e) {
+      return { error: e as Error, result: false };
+    }
+    return { result: true };
   }
 
   getContext(): LDContext {
-    return { kind: 'user', key: 'test-context-1' };
+    return clone(this.context);
```
yusinto, line 113: I copied the legacy function from js-common-sdk because `cloneWithRedactions` takes LDContextCommon and not LDContext.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -75,26 +88,37 @@ export default class LDClientImpl implements LDClient {
...
 
-  flush(onDone?: () => void): Promise<void> {
-    return Promise.resolve(undefined);
+  async flush(): Promise<{ error?: Error; result: boolean }> {
+    try {
+      await this.eventProcessor.flush();
+    } catch (e) {
+      return { error: e as Error, result: false };
+    }
+    return { result: true };
```
yusinto, line 109: @kinyoklion I rewrote this to return an object. I think is ok because the previous code's callback takes the error and result as args. Any comments?

packages/shared/common/src/utils/clone.ts
```diff
@@ -0,0 +1,3 @@
+export default function clone(obj: any) {
+  return JSON.parse(JSON.stringify(obj));
```
kinyoklion, line 2: So, this will be fine for now, but it does contain some weaknesses compared to how we do it for filtering.

Primarily this will throw an exception for circular contents and some other situations.

packages/shared/sdk-client/src/LDClientImpl.ts
kinyoklion: Sure. Hopefully most people never use the result in general.

packages/shared/common/src/utils/clone.ts
yusinto: Ok I understand. in the interest of time i will defer improvements to this to a later date.

PR #304, opened 2023-10-18
chore: implement identify

yusinto:
Initial implementation of identify, good enough for an alpha release.


packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -78,12 +73,11 @@ export default class LDClientImpl implements LDClient {
 
   async start() {
     try {
-      this.flags = await fetchFlags(this.sdkKey, this.context, this.config, this.platform);
+      await this.identify(this.context);
```
louis-launchdarkly, line 76: I think the idea makes sense in general - I am wondering if would this change the default behavior (or am I misunderstanding the default behavior?)

The current initialization behavior is to use the bootstrapped flags, and then look at the locally stored flags while the SDK makes the connection.

louis-launchdarkly (approved):

PR #305, opened 2023-10-21
chore: add react-native workspace

yusinto:


packages/sdk/react-native/LICENSE
```diff
@@ -0,0 +1,13 @@
+Copyright 2022 Catamorphic, Co.
```
kinyoklion, line 1: 2023

packages/sdk/react-native/README.md
```diff
@@ -0,0 +1,66 @@
+# LaunchDarkly React Native SDK
```
kinyoklion, line 1: Because the repo is public we should probably add a callout.

```
# :warning: UNSUPPORTED

This SDK is in pre-release development and is not supported. 

```

Or something.

kinyoklion (approved):
Couple minor inline comments.

PR #306, opened 2023-10-23
chore: Provider and hooks

yusinto:
* Added LDProvider
* Added hooks
* Added typed variation[Detail] functions

Please try and run the app, following the README under sdk/react-native to see if this is better than the existing legacy version in terms of brokenness.

### DIFF
--- a/packages/sdk/react-native/src/hooks/useBoolVariation.ts
+++ b/packages/sdk/react-native/src/hooks/useBoolVariation.ts
@@ -1,0 +1,13 @@
+import useLDClient from './useLDClient';
+
+export const useBoolVariation = (flagKey: string, defaultValue: boolean) => {
+  const ldClient = useLDClient();
+  return ldClient?.boolVariation(flagKey, defaultValue) ?? defaultValue;
+};
+
+export const useBoolVariationDetail = (flagKey: string, defaultValue: boolean) => {
+  const ldClient = useLDClient();
+  return ldClient?.boolVariationDetail(flagKey, defaultValue) ?? defaultValue;
+};
+
+export default useBoolVariation;
--- a/packages/sdk/react-native/src/provider/reactSdkContext.ts
+++ b/packages/sdk/react-native/src/provider/reactSdkContext.ts
@@ -1,0 +1,14 @@
+import { createContext } from 'react';
+
+import { LDClient, LDFlagSet } from '@launchdarkly/js-client-sdk-common';
+
+type ReactSdkContext = {
+  allFlags?: LDFlagSet;
+  ldClient?: LDClient;
+};
+
+const context = createContext<ReactSdkContext>({});
+
+const { Provider, Consumer } = context;
+
+export { context, Provider, Consumer, ReactSdkContext };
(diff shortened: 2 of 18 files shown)


packages/shared/sdk-client/src/api/LDClient.ts
yusinto, line 1: I ordered the props in this interface alphabetically automatically resulting in an ugly diff. The only significant changes are the additions of typed variation functions.

packages/sdk/react-native/src/provider/reactSdkContext.ts
```diff
@@ -0,0 +1,14 @@
+import { createContext } from 'react';
+
+import { LDClient, LDFlagSet } from '@launchdarkly/js-client-sdk-common';
+
+type ReactSdkContext = {
+  allFlags?: LDFlagSet;
+  ldClient?: LDClient;
+};
```
yusinto, line 8: This is still in flux and will change to have more/less things exposed.

packages/sdk/react-native/src/hooks/useBoolVariation.ts
```diff
@@ -0,0 +1,13 @@
+import useLDClient from './useLDClient';
+
+export const useBoolVariation = (flagKey: string, defaultValue: boolean) => {
```
kinyoklion, line 3: I think these will be very nice for consistent experiment results. (Aside from also allow type safety, which is nice.)

packages/sdk/react-native/src/hooks/useBoolVariation.ts
```diff
@@ -0,0 +1,13 @@
...
+  const ldClient = useLDClient();
+  return ldClient?.boolVariation(flagKey, defaultValue) ?? defaultValue;
+};
+
+export const useBoolVariationDetail = (flagKey: string, defaultValue: boolean) => {
+  const ldClient = useLDClient();
+  return ldClient?.boolVariationDetail(flagKey, defaultValue) ?? defaultValue;
```
kinyoklion, line 10: Because detailed variations return objects we need to ensure we keep stable object references. Which may happen for flags that exist, because we might be returning a value we have stored.

I am concerned that the variation for a default value will not be stable. Which could lead to the hook triggering more re-renders than desired.

packages/sdk/react-native/src/hooks/useBoolVariation.ts
yusinto: I agree. Thank you for pushing for these!

packages/sdk/react-native/src/hooks/useBoolVariation.ts
yusinto: Ok this is definitely a bug. What do you think of a fix like this:
```suggestion
  const def = {
    value: defaultValue,
    variationIndex: null,
    reason: null,
  };
  return ldClient?.boolVariationDetail(flagKey, defaultValue) ?? def;
```

yusinto:
I'm closing this PR to make improvements to the provider, hook and error apis.

Issue #312, opened 2023-11-10
`waitForInitialization` hanging when declaring multiple clients outside scope


jterstall:
**Describe the bug**
Client `waitForInitialization` function seems to hang for a long time when another client is initialized before it. Also the timeout parameter does not seem to work.

**To reproduce**
I used the following jest function to reproduce it locally:

```
describe('Launch Darkly tests', () => {
  const outsideLdCLient = init('sdk-key', { timeout: 1 });
  const outsideLdClient2 = init('sdk-key-3', { timeout: 1 });

  it('Outside LD client does not timeout here', async () => {
    try {
      await outsideLdClient2.waitForInitialization();
    } catch (e) {
      console.log((e as Error).message);
    }
  });
  it('Outside LD client does timeout here', async () => {
    try {
      await outsideLdCLient.waitForInitialization();
    } catch (e) {
      console.log((e as Error).message);
    }
  });
  it('Inside LD client does not timeout', async () => {
    const insideLdClient = init('sdk-key-2');
    try {
      await insideLdClient.waitForInitialization();
    } catch (e) {
      console.log((e as Error).message);
    }
  });
});
```

**Expected behavior**
All clients to fail and not timeout.

**Logs**
If applicable, add any log output related to your problem.

**SDK version**
9.0.2

**Language version, developer tools**
Node 18.15.0

**OS/platform**
WIndows 10

**Additional context**
I originally created this test because the same behaviour is happening when calling our lambda that retrieves a feature flag. The lambda function is timing out everytime.

kinyoklion:
Hello @jterstall,

The timeout parameter is not for the timeout of the SDK initialization. If is the connection timeout for individual requests made within the SDK. So, for instance, a single HTTP request is made this would be the timeout time to start receiving data. 

I am not 100% sure what you are specifically trying to illustrate with `waitForInitialization` independent of that.

The SDK will keep trying to reconnect in most cases, regardless of that timeout being specified. In these specific instances it should fail initialization in a permanent manner after the invalid SDK is receied.

If you wanted to timeout on overall initialization you would do that with `Promise.race`

```
    await Promise.race([
      client.waitForInitialization(),
      new Promise((resolve) => setTimeout(resolve, 3000)),
    ]);
```

I suspect the problem here may be related to it having been rejected before the wait starts. It seems if we have already rejected before the wait, then the logic for already failed initialization is missing. 

We will look into it.

Thanks,
Ryan

_Filed internally as 224103_

[...]

PR #314, opened 2023-11-14
fix: Better handle waiting for initialization for failure cases.

kinyoklion:
Affects #312

### DIFF
--- a/packages/shared/sdk-server/__tests__/LDClientImpl.test.ts
+++ b/packages/shared/sdk-server/__tests__/LDClientImpl.test.ts
@@ -51,6 +51,17 @@ describe('LDClientImpl', () => {
     expect(callbacks.onError).not.toBeCalled();
   });
 
+  it('wait for initialization completes even if initialization completes before it is called', (done) => {
+    setupMockStreamingProcessor();
+    client = createClient();
+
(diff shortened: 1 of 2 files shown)


packages/shared/sdk-server/__tests__/LDClientImpl.test.ts
```diff
@@ -74,6 +93,29 @@ describe('LDClientImpl', () => {
     expect(callbacks.onError).toBeCalled();
   });
 
+  it('initialization promise is rejected even if the failure happens before wait is called', (done) => {
```
kinyoklion, line 96: This test fails without the associated fix.

packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -231,9 +233,30 @@ export default class LDClientImpl implements LDClient {
   }
 
   waitForInitialization(): Promise<LDClient> {
```
kinyoklion, line 235: So, the associated github ticket looks to have just allowed the task system to execute things in different orders resulting in the undesirable behavior. It provided the opportunity for a completion to happen before the wait method was called.

keelerm84 (approved):

Issue #310, opened 2023-11-05
FileDataSource no longer emits flag changes on file updates


fhawkes:
**Describe the bug**
The FileDataSource no longer correctly tracks streaming (realtime) changes to flags when updating the json file on disk.

**To reproduce**
Create a LD client utilizing the FileDataSource with autoUpdate turned on and pointing to one json file on disk that looks similar to this

``` json
{
    "flagValues": {
        "coolFlag": false
    }
}
```

While creating the client subscribe to flag changes like so
``` typescript
this.client.on('update', (flagChange) => {
	// Expecting the flag change to be emitted, it is not
    logger.trace(`LaunchDarkly flag changed ${flagChange.key}.`);
});
```

**Expected behavior**
When changing the value of an existing flag in the json file, I expect events to be emitted by the LD client informing us about flag changes.

**Logs**
If applicable, add any log output related to your problem.

**SDK version**
@launchdarkly/node-server-sdk 9.0.2

**Language version, developer tools**
Node v18.17.1

**OS/platform**
macOS 13.5.2

**Additional context**
I did some digging and it appears that in the most recent version of this package we hardcode the flag version to always be 1 here
https://github.com/launchdarkly/js-core/blob/388c2874ba8fbb64683449aa848965548feda81a/packages/shared/sdk-server/src/data_sources/FileDataSource.ts#L24

This prevents the DataSourceUpdates code here from ever detecting a change with the flag because is specifically looks at the version which will always be equal. 
https://github.com/launchdarkly/js-core/blob/388c2874ba8fbb64683449aa848965548feda81a/packages/shared/sdk-server/src/data_sources/DataSourceUpdates.ts#L162

In previous versions of the sdk that were no typescript the version flag was not included. We could pull the version in from the the existing flag in the feature store if it exists, but I didn't see a clean way to fetch that value at a glance.

kinyoklion:
Hello @fhawkes,

Thank you for pointing this out. We will look into it.

As a work around the more defined flag format, with `flags:` should work if you include and increment the version. I know this is not ideal.

```
{
  "flags": {
    "coolFlag": {
      "key": "coolFlag",
      "on": true,
      "fallthrough": {
        "variation": 0
      },
      "variations": [
        false,
        true
      ],
      "version": 0
    }
  }
}
```

Thank you,
Ryan

_Filed internally as 224098_

fhawkes:
[...]

PR #317, opened 2023-11-16
fix: Increment version for changing flagValues

kinyoklion:
When using FileData you have the option of just providing values. The pre-typescript SDK was liberal with what it considered an update and it would emit updates as things changed.

Some improvements to prevent spurious update events were added in typscript and these interfered with the basic way the FileData implementation works.

This PR makes a couple updates:
1.) If the value of a "flagValue" changes, then we will increment the version associated with it.
2.) If the value of a "flagValue" does not change, then we will not increment the version.

This should mean when values change the version will change. Standard flag/segment JSON must update their versions manually.

This will add some performance overhead.

This addresses #310

### DIFF
--- a/packages/shared/sdk-server/src/data_sources/FileDataSource.ts
+++ b/packages/shared/sdk-server/src/data_sources/FileDataSource.ts
@@ -6,8 +6,9 @@ import {
   VoidFunction,
 } from '@launchdarkly/js-sdk-common';
 
-import { DataKind, LDFeatureStore, LDFeatureStoreDataStorage } from '../api';
+import { DataKind, LDFeatureStoreDataStorage } from '../api';
 import { FileDataSourceOptions } from '../api/integrations';
+import { LDDataSourceUpdates } from '../api/subsystems';
 import { Flag } from '../evaluation/data/Flag';
 import { processFlag, processSegment } from '../store/serialization';
 import VersionedDataKinds from '../store/VersionedDataKinds';
@@ -15,13 +16,13 @@ import FileLoader from './FileLoader';
 
 export type FileDataSourceErrorHandler = (err: LDFileDataSourceError) => void;
 
-function makeFlagWithValue(key: string, value: any): Flag {
+function makeFlagWithValue(key: string, value: any, version: number): Flag {
   return {
     key,
     on: true,
     fallthrough: { variation: 0 },
     variations: [value],
-    version: 1,
+    version: version ?? 1,
   };
 }
 
@@ -42,7 +43,7 @@ export default class FileDataSource implements subsystem.LDStreamProcessor {
   constructor(
     options: FileDataSourceOptions,
     filesystem: Filesystem,
-    private readonly featureStore: LDFeatureStore,
+    private readonly featureStore: LDDataSourceUpdates,
     private initSuccessHandler: VoidFunction = () => {},
     private readonly errorHandler?: FileDataSourceErrorHandler,
   ) {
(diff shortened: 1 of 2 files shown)


kinyoklion:
@fhawkes If you are setup for development could you give this a try and see if it solves your problems?

kinyoklion:
There are some additional mismatches in typing that should be addressed. In specifically the FileDataSourceFactory. The old implementation had access to the config directly and was able to access the featureStore. This typing remained, which means it isn't completely correct. We don't have an external type for an update only feature store abstraction, but we 100% should. 

This would be a breaking change, so I am not incorporating it as part of this change. Generally speaking it isn't an interface people implement often unless they are attempting to implement a custom update processor, which is a leaky abstraction we have maintained for backward compatibility.

packages/shared/sdk-server/src/data_sources/FileDataSource.ts
```diff
@@ -128,13 +130,22 @@ export default class FileDataSource implements subsystem.LDStreamProcessor {
...
       this.addItem(VersionedDataKinds.Features, parsed.flags[key]);
     });
     Object.keys(parsed.flagValues || {}).forEach((key) => {
-      const flag = makeFlagWithValue(key, parsed.flagValues[key]);
+      const previousInstance = oldData[VersionedDataKinds.Features.namespace]?.[key];
+      let { version } = previousInstance ?? { version: 1 };
+      // If the data is different, then we want to increment the version.
+      if (
+        previousInstance &&
+        JSON.stringify(parsed.flagValues[key]) !== JSON.stringify(previousInstance?.variations?.[0])
```
kinyoklion, line 144: This isn't deterministic. So we may want to do something different here.

kinyoklion: We don't recommend using FileDataSource in production, so this may be sufficient.

packages/shared/sdk-server/src/data_sources/FileDataSource.ts
```diff
@@ -6,22 +6,23 @@ import {
...
 
-function makeFlagWithValue(key: string, value: any): Flag {
+function makeFlagWithValue(key: string, value: any, version: number): Flag {
   return {
     key,
     on: true,
     fallthrough: { variation: 0 },
     variations: [value],
-    version: 1,
+    version: version ?? 1,
```
yusinto, line 25: Is null coalescing needed here? `version` is a number and should not be null or undefined.

```suggestion
    version,
```

yusinto (changes requested):
One small [comment](https://github.com/launchdarkly/js-core/pull/317/files#r1399726920).

fhawkes:
> @fhawkes If you are setup for development could you give this a try and see if it solves your problems?

Hey @kinyoklion  I was able to test these changes out and they do solve the issue for us! I'm now seeing flag changes get emitted as expected. Thanks!!

packages/shared/sdk-server/src/data_sources/FileDataSource.ts
kinyoklion: Interesting that linting doesn't catch this. It was leftover from an interim version. Thank you for noticing it.

fhawkes:
Any movement on this one?

packages/shared/sdk-server/src/data_sources/FileDataSource.ts
kinyoklion: I think we can do with this, and then change when we can. Most flags aren't structured as well.

kinyoklion:
Hey @fhawkes,

No movement. I will see if I can get to it soon.

Thank you,
Ryan

yusinto (approved):

PR #319, opened 2023-11-22
build: modfying publish action to use npm cli to include provenance in npm publish

rsoberano-ld:
**Requirements**

**Related issues**

Yarn publish doesn't currently support NPM's publish package with provenance functionality. As a workaround until this is supported, we'll pack the workspace with yarn in order to guarantee we get the same package as before, and then use the npm cli to do the final publish with provenance. This also involves passing the workspace path as input parameters to the publish actions/script as npm's workspace functionality doesn't work exactly the same as yarns. 

While npm's generated provenance isn't the most robust provenance attestation, it results in a verified checkmark on the npm package page, which brings the provenance closest to the consumer and makes it most useful. 

Ideally we'd want yarn to support this natively, but tracking the yarn repo issues for the past couple months has shown no movement here. 

**Additional context**

Add any other context about the pull request here.

### DIFF
--- a/.github/workflows/manual-publish.yml
+++ b/.github/workflows/manual-publish.yml
@@ -41,7 +41,9 @@ jobs:
         with:
           node-version: 16.x
           registry-url: 'https://registry.npmjs.org'
-      - uses: ./actions/install-latest-npm
+      - uses: ./actions/install-npm-version
+        with: 
+          npm_version: 9.5.0
       - name: 'Setup Redis'
         if: ${{ inputs.workspace_path == 'packages/store/node-server-sdk-redis' }}
         run: |
--- a/.github/workflows/manual-publish.yml
+++ b/.github/workflows/manual-publish.yml
@@ -41,9 +41,7 @@ jobs:
         with:
           node-version: 16.x
           registry-url: 'https://registry.npmjs.org'
-      - name: 'Install latest npm to support provenance publishing'
-        run: |
-          npm install -g npm
+      - uses: ./actions/install-latest-npm
       - name: 'Setup Redis'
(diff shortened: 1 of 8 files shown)


rsoberano-ld:
Compare with changes made to learn-release-please/npm-provenance-test repo during npm provenance testing: https://github.com/launchdarkly/learn-release-please/compare/7e9e9b348398ea077d789b535ef4d7b937f2aeae...rsoberano/SEC-4384/js-core-sigstore-updates

.github/workflows/manual-publish.yml
```diff
@@ -41,6 +41,9 @@ jobs:
         with:
           node-version: 16.x
           registry-url: 'https://registry.npmjs.org'
+      - name: 'Install latest npm to support provenance publishing'
+        run: |
+          npm install -g npm
```
rsoberano-ld, line 46: The npm that comes with `node-version: 16.x` is an older version that doesn't support the `--provenance` option, so adding a step here to update it to latest per https://docs.npmjs.com/downloading-and-installing-node-js-and-npm/. 

Open to other suggestions of how to update npm here, though

actions/install-latest-npm/action.yml
```diff
@@ -0,0 +1,8 @@
+name: Install latest npm
+description: Install the latest version of the npm CLI utility
+runs:
+  using: composite
+  steps:
+    - name: 'Install latest npm to support provenance publishing'
+      run: |
+        npm install -g npm
```
kinyoklion, line 8: I think you are going to need a `shell: bash` entry. If you have your IDE setup with the github actions schema it should show you an error.

Could we add a newline to the end of the file as well.

kinyoklion (approved):

PR #320, opened 2023-11-23
chore: Exclude js-eventsource logic from StreamingProcessor

yusinto:
This PR modifies the StreamingProcessor so it is compatible with  any EventSource implementation, not just js-eventsource. The react native sdk has its own event source implementation, so this is necessary. This is also a precursor for future attempts to replace the js-eventsource completely.


packages/shared/common/src/api/platform/EventSource.ts
```diff
@@ -7,7 +9,7 @@ export type ProcessStreamResponse = {
 
 export interface EventSource {
   onclose: (() => void) | undefined;
-  onerror: (() => void) | undefined;
+  onerror: ((err: HttpErrorResponse) => void) | undefined;
   onopen: (() => void) | undefined;
   onretrying: ((e: { delayMillis: number }) => void) | undefined;
```
yusinto, line 14: callout: I understand this is a public api so this is a breaking change(?)

packages/shared/common/src/api/platform/Requests.ts
```diff
@@ -81,7 +81,7 @@ export interface Options {
 export interface Requests {
   fetch(url: string, options?: Options): Promise<Response>;
 
-  createEventSource(url: string, eventSourceInitDict: EventSourceInitDict): EventSource;
+  createEventSource(url: string, eventSourceInitDict?: EventSourceInitDict | any): EventSource;
```
yusinto, line 84: Relax es creation options so other implementations can work.

packages/shared/common/src/internal/stream/StreamingProcessor.ts
```diff
@@ -64,52 +50,52 @@ class StreamingProcessor implements LDStreamProcessor {
...
+    this.eventSource = this.createEventSource();
 
-    const eventSource = this.requests.createEventSource(this.streamUri, {
-      headers: this.headers,
-      errorFilter,
-      initialRetryDelayMillis: 1000 * this.streamInitialReconnectDelay,
-      readTimeoutMillis: STREAM_READ_TIMEOUT_MS,
-      retryResetIntervalMillis: RETRY_RESET_INTERVAL_MS,
-    });
-    this.eventSource = eventSource;
```
yusinto, line 93: The logic for creating an EventSource now the responsibility of the consumer in order to keep StreamingProcessor as compatible as possible.

packages/shared/common/src/internal/stream/StreamingProcessor.ts
```diff
@@ -64,52 +50,52 @@ class StreamingProcessor implements LDStreamProcessor {
...
-
-    eventSource.onclose = () => {
+    this.eventSource.onclose = () => {
       this.logger?.info('Closed LaunchDarkly stream connection');
     };
 
-    eventSource.onerror = () => {
-      // The work is done by `errorFilter`.
```
yusinto, line 100: I checked [js-eventsource](https://github.com/launchdarkly/js-eventsource/blob/44e77f3f6debb322d5233fe2d3f9484cd1ded55e/lib/eventsource.js#L157) and it uses `errorFilter` for two purposes: to handle errors and to determine if a retry should be made. With this change, `errorFilter`'s only responsibility is the latter i.e. determining if a retry should be made. 

The error handling will be done by `logAndHandleErrors`. The net effect is there
s no change in functionality because `js-eventsource` emits the `error` event after invoking `errorFilter`.

packages/shared/common/src/internal/stream/StreamingProcessor.ts
```diff
@@ -64,52 +50,52 @@ class StreamingProcessor implements LDStreamProcessor {
...
+    this.logger?.warn(httpErrorMessage(err, 'streaming request', 'will retry'));
+    this.logConnectionResult(false);
+    this.logConnectionStarted();
+  }
 
-      this.logger?.warn(httpErrorMessage(err, 'streaming request', 'will retry'));
-      this.logConnectionResult(false);
-      this.logConnectionStarted();
-      return true;
-    };
```
yusinto, line 82: `errorFilter` is now declared as part of `createEventSource`.  See my comments [below](https://github.com/launchdarkly/js-core/pull/320/files#r1402818569).

packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -187,16 +189,31 @@ export default class LDClientImpl implements LDClient {
...
+      const streamUri = `${serviceEndpoints.streaming}/all`;
+
+      return requests.createEventSource(streamUri, {
+        headers: defaultHeaders(sdkKey, info, tags),
+        errorFilter: shouldRetry,
+        initialRetryDelayMillis: 1000 * this.config.streamInitialReconnectDelay,
+        readTimeoutMillis: 5 * 60 * 1000,
+        retryResetIntervalMillis: 60 * 1000,
+      });
+    };
```
yusinto, line 207: Creates js-eventsource. `errorFilter` is now only responsible for determining if retries should be made.

packages/shared/common/src/internal/stream/StreamingProcessor.test.ts
```diff
@@ -94,50 +82,13 @@ describe('given a stream processor with mock event source', () => {
...
-      `${serviceEndpoints.streaming}/all`,
-      {
-        errorFilter: expect.any(Function),
-        headers: defaultHeaders(sdkKey, info, tags),
-        initialRetryDelayMillis: 22000,
-        readTimeoutMillis: 300000,
-        retryResetIntervalMillis: 60000,
-      },
-    );
-  });
```
yusinto, line 132: Moved to LDClientImpl.test.ts because `createEventSource` logic is part of LDClientImpl now.

kinyoklion:
So, other SDKs have traditionally already had other event sources. This API is abstract enough to work with the polyfill or the browser, for instance. It does require writing some small amount of glue code for compatibility with the interface.

Is there some specific reason you could not encapsulate it? I don't see why the streaming level would need access to implementation detail related to this event source.

As we just did two major versions of the node SDK I am against further changes to the public interface.

packages/shared/common/src/api/platform/Requests.ts
kinyoklion: What is the problem with this? This provides a stable way to make different event sources. It isn't intended to attack to the specific event source. Allowing any removes the value of having some common abstract settings.

All event sources may not need to use them, but we should have a superset of data. For instance the browser event source wouldn't use anything but the error filter (any it may only result in reporting.)

packages/shared/common/src/api/platform/Requests.ts
```diff
@@ -93,3 +93,8 @@ export interface Requests {
    */
   usingProxyAuth?(): boolean;
 }
+
+export interface HttpErrorResponse {
```
kinyoklion, line 97: This is probably good/fine.

kinyoklion (changes requested):
I need to understand why these changes are needed. I don't see anything that would make or break compatibility with any event source that we may want to implement.

I think the changes themselves are improvements.

packages/shared/common/src/utils/http.ts
```diff
@@ -46,3 +47,7 @@ export function httpErrorMessage(
   const action = retryMessage ?? 'giving up permanently';
   return `Received ${desc} for ${context} - ${action}`;
 }
+
+export function shouldRetry(err: HttpErrorResponse) {
+  return !(err.status && !isHttpRecoverable(err.status));
```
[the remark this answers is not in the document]
kinyoklion: (And theoretically status is not optional based on typing, but is based on reality.)

packages/shared/common/src/api/platform/EventSource.ts
```diff
@@ -16,7 +18,7 @@ export interface EventSource {
 }
 
 export interface EventSourceInitDict {
-  errorFilter: (err: { status: number; message: string }) => boolean;
```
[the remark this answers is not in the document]
kinyoklion: Theoretically breaking, but not really a breaking change being as it could have actually been undefined already. :shrug:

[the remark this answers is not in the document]
yusinto: Ok I've made status optional and made the err argument optional as well for backwards compat.

packages/shared/common/src/api/platform/Requests.ts
kinyoklion: @yusinto This is the one I was talking about. We should have standard parameteters and the event source decides what to do with them. Versus `any`.

kinyoklion (approved):

PR #321, opened 2023-11-30
chore: RN provider and hooks attempt 2

yusinto:
This is a second attempt of #306 using the newly added event source.

- use StreamingProcessor and event source to fetch flags
- added LDProvider
- added hooks
- added typed variation[Detail] functions

Please read the [example README](https://github.com/launchdarkly/js-core/blob/9d458f28625c12169df614c8d2289f9c0952cc45/packages/sdk/react-native/example/README.md#L1) and try running the example app.


packages/sdk/react-native/package.json
```diff
@@ -40,7 +40,7 @@
     "ios": "yarn ./example && yarn build && yarn ./example ios"
   },
   "peerDependencies": {
-    "react-native": "*"
+    "react": "*"
```
yusinto, line 43: We need react for `useMemo`.

packages/shared/sdk-client/src/api/LDClient.ts
yusinto, line 1: I only alphabetically re-ordered the members of this interface, no change in functionality here.

packages/shared/sdk-client/src/evaluation/fetchUtils.ts
```diff
@@ -1,19 +1,14 @@
...
- * because the react-native repo uses it too. Set the global.btoa to the encode
- * function of base64-js.
- * https://github.com/beatgammit/base64-js
- * https://github.com/axios/axios/issues/2235#issuecomment-512204616
- *
- * Ripped from https://thewoods.blog/base64url/
- */
-export const base64UrlEncode = (s: string, encoding: Encoding): string =>
-  encoding.btoa(s).replace(/\+/g, '-').replace(/\//g, '_').replace(/=+$/, '');
-
```
yusinto, line 16: Moved to common so this can be shared.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -66,18 +73,6 @@ export default class LDClientImpl implements LDClient {
...
-
-  async start() {
-    try {
-      await this.identify(this.context);
-      this.emitter.emit('ready');
-    } catch (error: any) {
-      this.emitter.emit('failed', error);
-      throw error;
-    }
   }
```
yusinto, line 76: This seems redundant because identify does the same work. It is a breaking change so I am calling it out here.

packages/sdk/react-native/src/ReactNativeLDClient.test.ts
yusinto, line 1: These are just basic tests for now. I will add more soon in a separate PR.

kinyoklion (changes requested):
Your tests seem like maybe they have an open handles problem and are failing.

It would be nice to have some amount of documentation on each of the hooks. To provide some intellisense so developers no what to expect.

I am continuing to look through other parts, but looks reasonable so far.

yusinto:
> tests seem like maybe they have an open handles

Yep. The issue was EventProcessor was actually trying to send events. I have set `sendEvents: false` for unit tests to fix this. Thank you for checking.

packages/sdk/react-native/src/hooks/variation/useTypedVariation.ts
```diff
@@ -0,0 +1,76 @@
...
+import useLDDataSourceStatus from '../useLDDataSourceStatus';
+
+export const useTypedVariation = <T extends boolean | number | string | unknown>(
+  flagKey: string,
+  defaultValue: T,
+): T => {
+  const ldClient = useLDClient();
+  const { status } = useLDDataSourceStatus();
+
+  if (status === 'ready') {
```
kinyoklion, line 15: Is error a terminal state? This is concerning. 

Typically we always want the client to be used, we generate unknown flag events. If it goes from good to bad we also have cached events.

Or maybe the status needs explained. My reflex is this is problemati.

packages/sdk/react-native/src/hooks/variation/useTypedVariation.ts
```diff
@@ -0,0 +1,76 @@
...
+  const ldClient = useLDClient();
+  const { status } = useLDDataSourceStatus();
+  const unsupportedType = useMemo(
+    () => ({
+      value: defaultValue,
+      variationIndex: null,
+      reason: { kind: 'ERROR', errorKind: 'UNSUPPORTED_VARIATION_TYPE' },
+    }),
+    [defaultValue],
+  );
```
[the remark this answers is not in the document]
kinyoklion: Yeah, I think this will do it. :+1:

packages/sdk/react-native/src/hooks/variation/useTypedVariation.ts
yusinto: You are right. I did this originally to avoid valuation errors caused by a null/undefined ldClient. This was early in the iteration when using polling. The Provider now requires an ldClient instance so that means we can fallback to the ldClient by default. I have made this change now.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -103,27 +100,112 @@ export default class LDClientImpl implements LDClient {
...
+    );
+  }
+
+  private createIdentifyPromise() {
+    return new Promise<void>((resolve, reject) => {
+      if (this.identifyReadyListener) {
+        this.emitter.off('ready', this.identifyReadyListener);
+      }
+      if (this.identifyErrorListener) {
+        this.emitter.off('ready', this.identifyErrorListener);
```
kinyoklion, line 167:
```suggestion
        this.emitter.off('error', this.identifyErrorListener);
```

kinyoklion (changes requested):

packages/shared/sdk-client/src/LDClientImpl.ts
kinyoklion: It seems the failed functionality was not replaced.

I think the fundamental flow is also different.

So a ready/failed event was a singular activity. The SDK wouldn't emit ready more than once. You people put code there that they want to execute once the SDK is ready to go. Or when they know it will never be ready, `failed`.

kinyoklion:
> > documentation on each of the hooks
> 
> Good idea. These hooks call the ldClient variation methods, which are fully documented [there](https://github.com/launchdarkly/js-core/blob/cbfd5b13dd0935be371664b936780163a436e143/packages/shared/sdk-client/src/api/LDClient.ts#L29). I think we should link them somehow so we only maintain 1 copy? I'm not sure how to link them though.

Well, to make actual links you can do `{@link LDClient.variation}` for example. It isn't ideal because it will literally link you to the item.

I am not sure, cross package, if there is a way to insert the docs inline. With inheritance you can inherit them, but doesn't help for hooks.

yusinto:
> documentation on each of the hooks

Ok I duplicated the comments and also added comments for the base hooks. There were mistakes in the LDClient docs as well and I corrected those too.

kinyoklion (approved):

PR #326, opened 2023-12-18
chore: rn sdk use localstorage for bootstrapping

yusinto:
Please run the example app as well if you can by running `yarn ios` from the `react-native` folder.  Note that [expo 49 only supports rn 0.72](https://docs.expo.dev/versions/latest/#each-expo-sdk-version-depends-on-a-react-native-version). Expo 50 will work with 0.73 and it's in beta with full release coming in about month. Thank you for reviewing.

* Added bootstrap from storage.
* Added local copy of fast-deep-equal (see my comments [here](https://github.com/launchdarkly/js-core/pull/326/files#r1430690374)).
* Cleaned up and improved React Native SDK api and example.


packages/sdk/react-native/package.json
```diff
@@ -44,6 +44,7 @@
   },
   "dependencies": {
     "@launchdarkly/js-client-sdk-common": "0.0.1",
+    "@react-native-async-storage/async-storage": "^1.21.0",
```
yusinto, line 47: We need this now since AsyncStorage is removed in 0.73.

packages/sdk/react-native/src/hooks/useLDDataSourceStatus.ts
yusinto, line 1: I used this for debugging but now no longer required.

packages/shared/common/src/utils/fast-deep-equal/index.ts
yusinto, line 1: I add this locally because there are typescript es6 issues with the original package from npm. I tried my best to solve that but could not get it to work in the timebox. I think it's ok to have a copy of this so we can share it and be independent.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -191,22 +254,32 @@ export default class LDClientImpl implements LDClient {
...
+    const p = this.createPromiseWithListeners();
+    this.emitter.emit('initializing', context);
+
+    const flagsStorage = await this.getFlagsFromStorage(checkedContext.canonicalKey);
+    if (flagsStorage) {
+      this.logger.debug('Initializing all data from storage');
+      this.context = context;
+      this.flags = flagsStorage;
+      this.emitter.emit('ready', context);
+    }
```
yusinto, line 266: Callout here to highlight the new bootstrap from storage implementation.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -99,41 +99,99 @@ export default class LDClientImpl implements LDClient {
...
       deserializeData: JSON.parse,
-      processJson: (dataJson) => {
-        this.logger.debug('Initializing all data');
-        this.context = context;
-        this.flags = {};
-        Object.keys(dataJson).forEach((key) => {
-          this.flags[key] = dataJson[key];
-        });
-        this.emitter.emit('ready', context);
+      processJson: async (dataJson: Flags) => {
```
yusinto, line 115: Callout here to highlight the new synchronization logic between storage and streamer.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -99,41 +99,99 @@ export default class LDClientImpl implements LDClient {
...
+      processJson: async (dataJson: DeleteFlag) => {
+        this.logger.debug(`Streamer DELETE ${JSON.stringify(dataJson, null, 2)}`);
+        const existing = this.flags[dataJson.key];
+
+        if (existing && existing.version <= dataJson.version) {
+          const changeset: LDFlagChangeset = {};
+          changeset[dataJson.key] = {
+            previous: this.flags[dataJson.key].value,
+          };
+          delete this.flags[dataJson.key];
```
yusinto, line 191: I perform a hard delete here but the legacy js common performs a soft-delete (tombstoning). I don't see the need for soft-deleting. Is this ok?

packages/sdk/react-native/example/package.json
```diff
@@ -17,6 +17,7 @@
     "clean": "expo prebuild --clean && yarn cache clean && rm -rf node_modules && rm -rf .expo"
   },
   "dependencies": {
+    "@react-native-async-storage/async-storage": "1.18.2",
```
yusinto, line 20: I had to add this as a dep of the example app to get auto-linking to work. I'm not sure if this is the best way?

yusinto: Additionally, [expo 49 only supports rn 0.72](https://docs.expo.dev/versions/latest/#each-expo-sdk-version-depends-on-a-react-native-version). Expo 50 will work with 0.73 and it's in beta with full release coming in about month.

packages/sdk/react-native/src/ReactNativeLDClient.ts
```diff
@@ -1,26 +1,33 @@
...
-import platform from './platform';
+import createPlatform from './platform';
 
 export default class ReactNativeLDClient extends LDClientImpl {
   constructor(sdkKey: string, options: LDOptions = {}) {
     const logger =
       options.logger ??
       new BasicLogger({
-        level: 'info',
+        level: 'debug',
```
yusinto, line 17: I intentionally left this as `debug` for the alpha.

packages/shared/sdk-client/src/api/LDEmitter.ts
```diff
@@ -1,4 +1,4 @@
-export type EventName = 'connecting' | 'ready' | 'error' | 'change';
```
kinyoklion, line 1: Are you keeping a list of breaking changes for when we do a js-client release? Things like this seem like they could be easy to slip through the cracks.

packages/shared/sdk-client/src/LDClientImpl.ts
kinyoklion: I don't have a good picture of what the external picture is for using the SDK.

When do I first know the SDK is ready?
When do I know that a specific identify is complete?

Can I split those two things?
Can I use async as well as callbacks?

(This looks like it emits ready any time we get a put, but if we have bootrap we should already be ready. Or if you got more than 1 put message, not something we currently do, but I don't think we specifically disallow it.)

kinyoklion: Reconnecting to a stream is probably in this case as well.

yusinto: > When do I first know the SDK is ready?

It's like you said. `ready` is emitted at 2 places: 

* bootstrap from [storage is complete](https://github.com/launchdarkly/js-core/blob/87d09ba00cd8aa15ea605cd7d479128b6c307e67/packages/shared/sdk-client/src/LDClientImpl.ts#L266)
* if no bootstrap when the [put is complete](https://github.com/launchdarkly/js-core/blob/87d09ba00cd8aa15ea605cd7d479128b6c307e67/packages/shared/sdk-client/src/LDClientImpl.ts#L151).

> When do I know that a specific identify is complete?

There are two ways:
* awaiting `identify` like [this](https://github.com/launchdarkly/js-core/blob/80cc003407bdbd4aa46b9a1d0043dc8528f6adbe/packages/sdk/react-native/example/src/welcome.tsx#L14)
* listening to the `ready` event which includes the context identified as a callback param.

> Can I split those two things?

Can you please clarify what you want to split? In the clean-slate start-from-scratch scenario, you call `identify` and either await that or listen to `ready` for completion.

> Can I use async as well as callbacks?

`identify` is async only. `ready` is an event so it needs an event listener with a callback.

packages/shared/sdk-client/src/api/LDEmitter.ts
yusinto: I plan to submit a follow up PR to update all comments, readme and docs including breaking changes. I want to get the code reviewed asap first. The `connecting` event is updated to `initializing` though to follow @cwaldren-ld 's [spec](https://github.com/launchdarkly/sdk-specs/tree/cw/sc-215574/persistent-spec/specs/SDKC-SDK-core/DS-data-system#141-overview).

packages/sdk/react-native/README.md
```diff
@@ -32,7 +32,13 @@ See the full [example app](https://github.com/launchdarkly/js-core/tree/main/pac
 
 ## Developing this SDK
 
-:information_source: See the example [README](https://github.com/launchdarkly/js-core/blob/main/packages/sdk/react-native/example/README.md#L1).
+```shell
+# ios
```
kinyoklion, line 36: The readme in the example seems to have instructions, but these instructions here have pre-requisites, like setting the mobile key, which makes this misleading.

packages/shared/mocks/src/streamingProcessor.ts
```diff
@@ -33,8 +35,16 @@ export const setupMockStreamingProcessor = (
             errorHandler(unauthorized);
           });
         } else {
-          // execute put which will resolve the init promise
+          // execute put which will resolve the identify promise
           process.nextTick(() => listeners.get('put')?.processJson(putResponseJson));
```
kinyoklion, line 39: I wonder if these were a `Promise.resolve` or `setTimeout` with 0 if you wouldn't need to process the ticks? (I know this is a test, but `process.nextTick` always alarms me for a moment, being as it isn't portable.

packages/shared/sdk-client/src/LDClientImpl.ts
kinyoklion: Technically yes, our algorithm currently requires a tombstone. It may not in the future. If you get a delete and a patch you could get them in the wrong order. Without the tombstone you wouldn't detect it.

Imagine flag A has a value of "a" in version 1.
We patch flag A to "b" in version 2.
We delete flag A in version 3.

We get version 1 on a put, we get version 3 on a delete, and then we get version 2.

With a tombstone we know we got the update out of order. Without the tombstone we do not, so the flag will come back into existence.

kinyoklion (changes requested):

packages/shared/mocks/src/streamingProcessor.ts
yusinto: I will attempt to replace nextTick with what you suggested and respond.

packages/sdk/react-native/README.md
yusinto: Good point. I will improve this to be clearer and independent.

packages/shared/sdk-client/src/LDClientImpl.ts
yusinto: Thanks for the clarification. I will do this.

packages/sdk/react-native/README.md
```diff
@@ -20,24 +21,64 @@ For more information, see the [complete reference guide for this SDK](https://do
...
+    </View>
+  );
+}
 ```
 
 See the full [example app](https://github.com/launchdarkly/js-core/tree/main/packages/sdk/react-native/example).
 
 ## Developing this SDK
 
 ```shell
```
kinyoklion, line 76: Need to add `MOBILE_KEY` to the `.env` as well.

yusinto: Done.

packages/shared/mocks/src/streamingProcessor.ts
```diff
@@ -24,7 +26,7 @@ export const setupMockStreamingProcessor = (
     ) => ({
       start: jest.fn(async () => {
         if (shouldError) {
-          process.nextTick(() => {
+          setTimeout(() => {
```
yusinto, line 29: Using `setTimeout` actually resolves the "weird" async issue in one of the storage tests related to ticks. @kinyoklion thanks for the suggestion!

packages/shared/sdk-client/src/LDClientImpl.ts
yusinto: Done in #328

kinyoklion (approved):

PR #340, opened 2023-12-29
fix: Add Detox e2e tests.

yusinto:
This adds e2e tests for ios using Detox. Please try running it following the [example/README](https://github.com/launchdarkly/js-core/blob/c81c90002d7db4c5b2ccdc059fd7b76a908deac3/packages/sdk/react-native/example/README.md#running-detox-e2e-tests). 

Additionally I also included some fixes for release-please to publish sdk-client and rn.

https://github.com/launchdarkly/js-core/assets/1593077/2f32316f-59a9-4601-ae07-5450ba29fe33


.github/workflows/release-please.yml
```diff
@@ -64,9 +89,9 @@ jobs:
           node-version: 16.x
           registry-url: 'https://registry.npmjs.org'
       - uses: ./actions/install-npm-version
-        with: 
+        with:
           npm_version: 9.5.0
-      - id: release-common
```
yusinto, line 69: This is a bug in the existing config. This step id should say release-sdk-server, but it says release-common instead. I think it's an artifact of copy pasting. The other packages suffer from the same bug and I've fixed them all too.

release-please-config.json
yusinto, line 1: I'm adding preMajor back in for sdk-client and rn now that I've included them in the publishing config.

yusinto, line 1: Adding `bump-minor-pre-major` now that the packages have been published as 0.1.x.

packages/sdk/react-native/example/README.md
```diff
@@ -28,3 +28,24 @@ yarn && yarn ios-go
...
+
+```shell
+brew tap wix/brew
+brew install applesimutils
+```
+
+2. On a terminal:
+
+```shell
+yarn && yarn start
```
kinyoklion, line 44: Need something like `cd e2e` or `From the e2e directory:`

yusinto: This needs to be run at the `example` level. I will update this and also include the extra step to run the xcode build for ios to generate the necessary ios folder and install cocoapods.

kinyoklion (approved):

Issue #345, opened 2024-01-11
Improve compatibility with TypeScript 4.x


ismaelgt:
**Is your feature request related to a problem? Please describe.**

Due to the usage of [type modifiers on import names](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#type-modifiers-on-import-names), the package can't be compiled as a dependency on projects using TypeScript <4.5.

**Describe the solution you'd like**

Remove the usage of type modifiers on import names to improve backwards compatibility with early versions of TypeScript.

**Describe alternatives you've considered**

The obvious solution would be upgrading the TypeScript version in the consumer, but this is not always easy (it's not in our case). [More info here](https://github.com/microsoft/TypeScript/issues/25778).

ismaelgt:
This is now fixed.

PR #346, opened 2024-01-11
fix: remove type modifiers on imports for better TS compatibility

ismaelgt:
**Requirements**

- [ ] ~I have added test coverage for new or changed functionality~
- [ ] ~I have validated my changes against all supported platform versions~

No new functionality has been added (only types import/export refactoring).

**Related issues**

https://github.com/launchdarkly/js-core/issues/345 https://github.com/launchdarkly/js-core/issues/347

Remove the usage of [type modifiers on import names](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#type-modifiers-on-import-names) to improve backwards compatibility with early versions of TypeScript.

The obvious solution would be upgrading the TypeScript version in the consumer, but this is not always easy (it's not in our case). [More info here](https://github.com/microsoft/TypeScript/issues/25778).

### DIFF
--- a/packages/shared/common/src/internal/stream/index.ts
+++ b/packages/shared/common/src/internal/stream/index.ts
@@ -1,4 +1,5 @@
 import StreamingProcessor from './StreamingProcessor';
-import { type StreamingErrorHandler } from './types';
+import type { StreamingErrorHandler } from './types';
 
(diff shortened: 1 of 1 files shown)


kinyoklion (approved):
This seems reasonable to me. Sorry for the delay in response.

PR #348, opened 2024-01-16
fix: Treat 413 HTTP status as recoverable for events.

kinyoklion:


packages/shared/common/src/internal/events/EventProcessor.ts
```diff
@@ -168,7 +168,9 @@ export default class EventProcessor implements LDEventProcessor {
 
   async flush(): Promise<void> {
     if (this.shutdown) {
-      throw new LDInvalidSDKKeyError('Events cannot be posted because SDK key is invalid');
+      throw new LDInvalidSDKKeyError(
```
kinyoklion, line 171: This type isn't ideal. I am hesitant to change the actual type though.

packages/shared/common/src/internal/events/EventProcessor.ts
```diff
@@ -168,7 +168,9 @@ export default class EventProcessor implements LDEventProcessor {
...
+        'Events cannot be posted because a permanent error has been encountered.',
+      );
```
yusinto, line 173: Maybe we should create a different error kind rather than use `LDInvalidSDKKeyError`? It is misleading and confusing.

yusinto (approved):
One minor comment about the error kind.

packages/shared/common/src/internal/events/EventProcessor.ts
kinyoklion: I did comment on this before the review. Do you have something to assuage my concern? I could alternatively throw this exception type specifically for only that condition. 

Adding different exception types could be considered a breaking change.

kinyoklion: I've extended the message at least to be less confusing.

yusinto: > Do you have something to assuage my concern

Apologies I missed your earlier comment about this. I checked the extended message and I think this is fine for now. Thank you.

PR #350, opened 2024-01-18
fix: Implement anonymous context processing

yusinto:
Apologies, I missed implementing this when moving the project to js-core.

### DIFF
--- a/packages/shared/common/src/Context.ts
+++ b/packages/shared/common/src/Context.ts
@@ -1,8 +1,14 @@
 /* eslint-disable no-underscore-dangle */
 // eslint-disable-next-line max-classes-per-file
-import { LDContextCommon, LDMultiKindContext, LDSingleKindContext, LDUser } from './api/context';
-import { LDContext } from './api/context/LDContext';
+import type {
+  LDContext,
+  LDContextCommon,
+  LDMultiKindContext,
+  LDSingleKindContext,
+  LDUser,
+} from './api';
 import AttributeReference from './AttributeReference';
+import { isLegacyUser, isMultiKind, isSingleKind } from './internal/context';
 import { TypeValidators } from './validators';
 
 // The general strategy for the context is to transform the passed in context
@@ -36,39 +42,6 @@ function encodeKey(key: string): string {
     return key.replace(/%/g, '%25').replace(/:/g, '%3A');
   }
   return key;
-}
-
-/**
- * Check if a context is a single kind context.
- * @param context
- * @returns true if the context is a single kind context.
- */
-export function isSingleKind(context: LDContext): context is LDSingleKindContext {
-  if ('kind' in context) {
-    return TypeValidators.String.is(context.kind) && context.kind !== 'multi';
-  }
-  return false;
-}
-
-/**
- * Check if a context is a multi-kind context.
- * @param context
- * @returns true if it is a multi-kind context.
- */
-export function isMultiKind(context: LDContext): context is LDMultiKindContext {
-  if ('kind' in context) {
-    return TypeValidators.String.is(context.kind) && context.kind === 'multi';
-  }
-  return false;
-}
-
-/**
- * Check if a context is a legacy user context.
- * @param context
- * @returns true if it is a legacy user context.
- */
-export function isLegacyUser(context: LDContext): context is LDUser {
-  return !('kind' in context) || context.kind === null || context.kind === undefined;
 }
 
 /**
--- a/packages/shared/common/src/Context.ts
+++ b/packages/shared/common/src/Context.ts
@@ -43,7 +43,7 @@ function encodeKey(key: string): string {
  * @param context
  * @returns true if the context is a single kind context.
  */
-function isSingleKind(context: LDContext): context is LDSingleKindContext {
+export function isSingleKind(context: LDContext): context is LDSingleKindContext {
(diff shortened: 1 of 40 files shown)


packages/shared/sdk-client/src/utils/calculateFlagChanges.ts
yusinto, line 1: This was in `index.ts` but I moved it now to its own explicit file.

packages/shared/sdk-client/src/utils/ensureKey.ts
```diff
@@ -0,0 +1,79 @@
...
+  LDContext,
+  LDContextCommon,
+  LDMultiKindContext,
+  LDSingleKindContext,
+  LDUser,
+  Platform,
+} from '@launchdarkly/js-sdk-common';
+import { isLegacyUser, isMultiKind, isSingleKind } from '@launchdarkly/js-sdk-common';
+
+export const ns = (s: string) => `LaunchDarkly_AnonKeys_${s}`;
```
yusinto, line 11: I'm using `LaunchDarkly_AnonKeys_` which is different from the existing js common. We may have to worry about migrating existing keys. Although for the react native sdk, since it's a brand new sdk using a new storage package, migration maybe difficult if not impossible.

packages/shared/sdk-client/src/LDClientImpl.test.ts
```diff
@@ -133,6 +135,20 @@ describe('sdk-client object', () => {
     });
   });
 
+  test('identify anonymous', async () => {
+    defaultPutResponse['dev-test-flag'].value = false;
+    const carContext: LDContext = { kind: 'car', anonymous: true, key: '' };
```
yusinto, line 140: Even anonymous, `key` is a required field so it has to be set to an empty string. I think this is fine, but calling this out for concerns/discussion.

packages/shared/common/src/Context.ts
```diff
@@ -43,7 +43,7 @@ function encodeKey(key: string): string {
  * @param context
  * @returns true if the context is a single kind context.
  */
-function isSingleKind(context: LDContext): context is LDSingleKindContext {
+export function isSingleKind(context: LDContext): context is LDSingleKindContext {
```
kinyoklion, line 46: I have serious reservations about exporting these functions. We don't want them exported from the final package and becoming part of the stable API of all of the packages.

packages/shared/sdk-client/src/LDClientImpl.test.ts
kinyoklion: If you are ok with that requirement. I was originally planning on exporting a different type from client SDKs that did not require a key. Then internal to the client implementation decorate it (generate keys, verify anonymous is set)  and create a `Context`.

Because I think that the pattern we had before, where you don't include a key and that makes the context anonymous is very intuitive.

packages/shared/sdk-client/src/utils/ensureKey.ts
kinyoklion: Could we name this function `namespaced` or something more explicit. I read the later code before this and was like what is an `nsKind`?

packages/shared/common/src/Context.ts
yusinto: Will you be open to exporting these from `internal` please? They are useful internally and the common sdks can re-use them for internal logic.

kinyoklion: I am ok with them being in internal.

I am really concerned about them being useful outside these two cases. They really are not supposed to be.

We are provided an `LDContext` and we then validate that into a `Context`. Once we have a `Context` these functions are not useful. 

So, if we are careful to try not to use LDContext, aside from adding keys to it and then constructing a context, then that is probably fine.

yusinto: Agreed. I'll add comments in this file reflecting what you said.

packages/shared/sdk-client/src/LDClientImpl.test.ts
yusinto: I'm fine with this for now.

packages/shared/sdk-client/src/utils/ensureKey.ts
```diff
@@ -0,0 +1,81 @@
...
+      ensureKeyCommon(k, c as LDContextCommon, platform),
+    ),
+  );
+};
+
+const ensureKeyLegacy = async (c: LDUser, platform: Platform) => {
+  await ensureKeyCommon('user', c, platform);
+};
+
+const ensureKey = async (c: LDContext, platform: Platform) => {
```
kinyoklion, line 67: I think I missed that we are modifying the context which is provided, versus modifying a copy. Typically we would never directly change things which are provided to the SDK. 

We could either always clone, which is probably fine because identify isn't a high frequency situation.

Or only clone when we determine that we need to.

kinyoklion (changes requested):

packages/shared/sdk-client/src/utils/ensureKey.ts
yusinto: Ok I'll clone and return the new object from ensureKey.

yusinto: Done.

kinyoklion (approved):

PR #351, opened 2024-01-19
feat: Always inline contexts for feature events

keelerm84:


kinyoklion:
Do we need to change what harness version we are using? Or we will just be holding this PR until some time where we can re-run that?

keelerm84:
> Do we need to change what harness version we are using? Or we will just be holding this PR until some time where we can re-run that?

We will be able to use the v2 release for the test harness. But I'm still making changes as I work through the different types of SDKs so I've just been checking things locally. And since I've been making these PRs against a separate feature branch, I haven't been too worried about the failure in CI. 

I'm far enough along at this point that once the current PRs are merged, I can probably just cut a v2 release.

kinyoklion (approved):

PR #356, opened 2024-01-25
feat: Implement common client side support for auto environment attributes.

yusinto:
This is a follow up to #355. This PR implements client-sdk support for auto env. In addition, the mocks internal api have been updated to work better with crypto and mock resets. Server sdks are impacted by the mocks api changes so these are fixed in this PR as well.

### DIFF
--- a/packages/shared/sdk-client/src/api/LDOptions.ts
+++ b/packages/shared/sdk-client/src/api/LDOptions.ts
@@ -24,7 +24,6 @@ export interface LDOptions {
      * Example: `authentication-service`
      */
     id?: string;
-
     /**
      * A unique identifier representing the version of the application where the LaunchDarkly SDK is running.
      *
@@ -35,6 +34,18 @@ export interface LDOptions {
      */
     version?: string;
   };
+
+  /**
+   * Enable / disable Auto environment attributes. When enabled, the SDK will automatically
+   * provide data about the mobile environment where the application is running. This data makes it simpler to target
+   * your mobile customers based on application name or version, or on device characteristics including manufacturer,
+   * model, operating system, locale, and so on. We recommend enabling this when you configure the SDK. To learn more,
+   * read [Automatic environment attributes](https://docs.launchdarkly.com/sdk/features/environment-attributes).
+   * for more documentation.
+   *
+   * The default is true.
+   */
+  autoEnvAttributes?: boolean;
 
   /**
    * The base uri for the LaunchDarkly server.
--- a/packages/shared/sdk-client/src/utils/addAutoEnv.ts
+++ b/packages/shared/sdk-client/src/utils/addAutoEnv.ts
@@ -1,0 +1,87 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+import {
+  clone,
+  internal,
+  LDApplication,
+  LDContext,
+  LDDevice,
+  LDMultiKindContext,
+  LDSingleKindContext,
+  LDUser,
+  Platform,
+} from '@launchdarkly/js-sdk-common';
+
+import Configuration from '../configuration';
+import { getOrGenerateKey } from './getOrGenerateKey';
+
+const { isLegacyUser, isSingleKind } = internal;
+const defaultAutoEnvSchemaVersion = '1.0';
+
+export const toMulti = (c: LDSingleKindContext) => {
+  const { kind, ...contextCommon } = c;
+
+  return {
+    kind: 'multi',
+    [kind]: contextCommon,
+  };
+};
+
+/**
+ * Clones the LDApplication object and populates the key, id and version fields.
+ *
+ * @param crypto
+ * @param info
+ * @param config
+ * @return An LDApplication object with populated key, id and version.
+ */
+export const addApplicationInfo = ({ crypto, info }: Platform, config: Configuration) => {
+  const { name, version, wrapperName, wrapperVersion } = info.sdkData();
+  const { ld_application } = info.platformData();
+
+  const ldApplication = clone<LDApplication>(ld_application) ?? {};
+  ldApplication.id = config.application?.id || ldApplication?.id || name || wrapperName;
+  ldApplication.name = ldApplication?.name || name || wrapperName;
+  ldApplication.version =
+    config.application?.version || ldApplication.version || version || wrapperVersion;
+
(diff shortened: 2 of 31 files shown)


packages/shared/sdk-client/src/api/LDOptions.ts
```diff
@@ -36,6 +35,18 @@ export interface LDOptions {
...
+   * Enable / disable Auto environment attributes. When enabled, the SDK will automatically
+   * provide data about the mobile environment where the application is running. This data makes it simpler to target
+   * your mobile customers based on application name or version, or on device characteristics including manufacturer,
+   * model, operating system, locale, and so on. We recommend enabling this when you configure the SDK. To learn more,
+   * read [Automatic environment attributes](https://docs.launchdarkly.com/sdk/features/environment-attributes).
+   * for more documentation.
+   *
+   * The default is true.
+   */
+  autoEnvAttributes?: boolean;
```
kinyoklion, line 48: This needs to be explicitly opt-in and cannot default to true.

So, usually the constructor or init methods have been updated with an additional required field.

So, above this line is a requirement, below it is a suggestion.

Also, I would generally recommend it not be a boolean, for forward compatibility.

Example from flutter:
`LDClient('sdk-key', AutoEnvAttributes.enabled)`

Then, if we want to have some new setting, we don't have to major version.

For instance `AutoEnvAttributes.appOnly` or something.

packages/shared/sdk-client/src/utils/addAutoEnv.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  ldDevice.os.name = os?.name || ldDevice.os.name;
+  ldDevice.os.version = os?.version || ldDevice.os.version;
+  ldDevice.key = await getOrGenerateKey('ld_device', platform);
+  ldDevice.envAttributesVersion = ldDevice.envAttributesVersion || defaultAutoEnvSchemaVersion;
+
+  return ldDevice;
+};
+
+export const addAutoEnv = async (context: LDContext, platform: Platform, config: Configuration) => {
```
kinyoklion, line 74: I think this is missing the logic of not replacing a user-provided context.

So, if for whatever reason, someone makes an `ld_device` context, then we just leave that context alone.

yusinto: Fixed. Please see line 106 onwards below.

packages/shared/sdk-client/src/api/LDOptions.ts
yusinto: [Done](https://github.com/launchdarkly/js-core/blob/75bb880f7986902d725b43d8122b7c9015916a11/packages/shared/sdk-client/src/api/AutoEnvAttributes.ts#L12).

```tsx
export enum AutoEnvAttributes {
  Disabled,
  Enabled,
}
```

yusinto:
Note react native tests are broken, but will be fixed in #357 to avoid bloating this PR even more.

packages/shared/sdk-client/src/api/LDOptions.ts
yusinto: [Done](https://github.com/launchdarkly/js-core/blob/5243f3fe225d991167d296b9f0da2fd1c336e994/packages/shared/common/src/api/platform/AutoEnv.ts#L12).

```tsx
export enum AutoEnvAttributes {
  Disabled,
  Enabled,
}
```

packages/shared/sdk-client/src/utils/getOrGenerateKey.ts
```diff
@@ -0,0 +1,15 @@
+import { Platform } from '@launchdarkly/js-sdk-common';
+
+export const addNamespace = (s: string) => `LaunchDarkly_AnonKeys_${s}`;
+
+export const getOrGenerateKey = async (kind: string, { crypto, storage }: Platform) => {
```
kinyoklion, line 5: I thought we were going to use a different namespace for these generated contexts. So anon keys would be in `LaunchDarkly_AnonKeys_` and the generated device keys would be in something like `LaunchDarkly_GenContextKeys_` or something.

kinyoklion: I think everything else looks good now.

yusinto: You are right I missed this. It's bug, I'll fix it. Thank you.

packages/shared/sdk-client/src/utils/getOrGenerateKey.ts
```diff
@@ -0,0 +1,38 @@
+import { Platform } from '@launchdarkly/js-sdk-common';
+
+export type Namespace = 'anonymous' | 'context';
```
yusinto, line 3: Separate namespaces for anonymous and non-anonymous. @kinyoklion

packages/shared/sdk-client/src/utils/getOrGenerateKey.ts
yusinto: This is now [done](https://github.com/launchdarkly/js-core/pull/356/files#r1468633109).

yusinto:
I'm not sure why the builds all failed. The error is quite strange. I can't reproduce locally and so this is frustrating. It maybe because of typescript versions.

```sh
'crypto' is referenced directly or indirectly in its own type annotation.
```

package.json
```diff
@@ -49,5 +49,9 @@
     "typedoc": "0.25.0",
     "typescript": "5.1.6"
   },
-  "packageManager": "yarn@3.4.1"
+  "packageManager": "yarn@3.4.1",
+  "//": "There is a bug with @types/node: 20.11.8 where crypto types are erroneous. See https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/68351",
+  "resolutions": {
+    "@types/node": "20.11.7"
+  }
```
yusinto, line 56: Please see my comments [here](https://github.com/launchdarkly/js-core/pull/356#issuecomment-1913323946).

kinyoklion: We will have to be sure that we remove this in a timely manner after the fix is available.

kinyoklion (approved):

package.json
yusinto: I have removed this because the DT team has released 20.11.10 which works.

PR #357, opened 2024-01-25
feat: React-native support for auto-env attributes.

yusinto:
This PR adds auto env support for the react-native sdk.


packages/shared/sdk-server/__tests__/store/serialization.test.ts
yusinto, line 1: This fixes `Test title is used multiple times in the same describe block  jest/no-identical-title`.

kinyoklion (approved):

PR #358, opened 2024-01-29
chore: Add client sdk application name and versionname.

yusinto:


packages/shared/common/__tests__/options/ApplicationTags.test.ts
```diff
@@ -1,82 +1,90 @@
...
-            throw new Error(`Did not find expectation for ${call[0]}`);
-          }
-          expected.splice(index, 1);
-        });
-        if (expected.length) {
-          throw new Error(
-            `Did not find expected messages: ${expected.map((item) => item.matches.toString())}`,
-          );
-        }
-      }
```
yusinto, line 79: This has been simplified to use the built-in `toHaveBeenCalledTimes` and `toHaveBeenCalledWith`.

packages/shared/common/__tests__/options/ApplicationTags.test.ts
yusinto, line 1: * Replaced inline logger with `logger` from the mocks package.
* Added name and versionName tests.
* Improved jest mock assertions using native jest mock assertions.

yusinto:
Note react native tests are broken, but will be fixed in #357 to avoid bloating #356 and this PR.

packages/shared/sdk-client/src/api/LDOptions.ts
```diff
@@ -33,6 +33,22 @@ export interface LDOptions {
      * Example: `1.0.0` (standard version string) or `abcdef` (sha prefix)
      */
     version?: string;
+
+    /**
```
kinyoklion, line 37: We may want to add these to the config for server as well. (Though we cannot rename `application` there.)

yusinto: Done.

louis-launchdarkly (approved):
We have reviewed this together on zoom and the change looks correct to me.

PR #359, opened 2024-01-31
fix: Implement RN SDK EventSource jitter backoff.

yusinto:
This PR adds jitter and backoff logic to the RNEventSource.

Screenshot 2024-01-30 at 2 52 01 PM


packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.ts
```diff
@@ -12,16 +12,27 @@ const XMLReadyStateMap = ['UNSENT', 'OPENED', 'HEADERS_RECEIVED', 'LOADING', 'DO
...
+const jitterRatio = 0.5; // Delay should be 50%-100% of calculated time.
+
+export function backoff(base: number, retryCount: number) {
+  const delay = base * Math.pow(2, retryCount);
+  return delay > maxRetryDelay ? maxRetryDelay : delay;
+}
+
+export function jitter(computedDelayMillis: number) {
+  return computedDelayMillis - Math.trunc(Math.random() * jitterRatio * computedDelayMillis);
+}
```
yusinto, line 34: Duplicated from legacy js-common-sdk.

packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.ts
```diff
@@ -234,7 +244,8 @@ export default class EventSource<E extends string = never> {
       } else if (line.indexOf('retry') === 0) {
         retry = parseInt(line.replace(/retry:?\s*/, ''), 10);
         if (!Number.isNaN(retry)) {
-          this.pollingInterval = retry;
+          // GOTCHA: Ignore the server retry recommendation. Use our own custom getNextRetryDelay logic.
+          // this.pollingInterval = retry;
```
yusinto, line 248: The original source code uses the server recommended retry seconds but we don't need that for our purpose.

packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.ts
```diff
@@ -128,16 +143,18 @@ export default class EventSource<E extends string = never> {
 
         if (this.xhr.status >= 200 && this.xhr.status < 400) {
           if (this.status === this.CONNECTING) {
+            this.retryCount = 0;
```
kinyoklion, line 146: Typically we would only reset this after having been active for 1 minute.

kinyoklion (approved):

PR #360, opened 2024-02-01
fix: RN streamer connection in background and foreground.

yusinto:
This PR adds a new hook `useAppState` to manage streamer connections on AppState transitions.

### DIFF
--- a/packages/sdk/react-native/src/provider/useAppState.ts
+++ b/packages/sdk/react-native/src/provider/useAppState.ts
@@ -3,17 +3,24 @@ import { AppState, AppStateStatus } from 'react-native';
 
 import { debounce } from '@launchdarkly/js-client-sdk-common';
 
+import { PlatformRequests } from '../platform';
 import ReactNativeLDClient from '../ReactNativeLDClient';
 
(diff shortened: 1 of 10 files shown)


packages/sdk/react-native/src/provider/useAppState.ts
```diff
@@ -0,0 +1,58 @@
...
+      client.logger.debug('App state background stopping streamer.');
+      client.streamer?.stop();
+    } else {
+      client.logger.debug('No action needed.');
+    }
+
+    appState.current = nextAppState;
+  };
+
+  // debounce with a default delay of 5 seconds.
```
kinyoklion, line 46: This will likely not run sometimes on iOS, or maybe at all, when transitioning to the background. But that is probably fine, because restoring entering the foreground will be more important. My only concern will be if we end up basically superseding the transition with the retry.

Theoretically, on iOS, you only are certain to be able to execute during the transition, and not after. Which is the reason flutter doesn't debounce it, even though I would generally like to.

kinyoklion (approved):

louis-launchdarkly (approved):

PR #362, opened 2024-02-06
fix: Improvements and fixes from docs review.

yusinto:
These are discovered after the docs PR review:

* Client sdk default baseUri should be clientsdk.launchdarkly instead of sdk.launchdarkly.
* Add getConnectionMode.
* Remove identify hash arg.
* Change rn default logging to info.
* Add debug config option.
* Add comments about autoEnv applicationInfo.
* Remove LDProvider context prop.

### DIFF
--- a/packages/sdk/react-native/example/App.tsx
+++ b/packages/sdk/react-native/example/App.tsx
@@ -8,7 +8,13 @@ import {
 
 import Welcome from './src/welcome';
 
-const featureClient = new ReactNativeLDClient(MOBILE_KEY, AutoEnvAttributes.Enabled);
+const featureClient = new ReactNativeLDClient(MOBILE_KEY, AutoEnvAttributes.Enabled, {
+  debug: true,
+  applicationInfo: {
+    id: 'ld-rn-test-app',
+    version: '0.0.1',
+  },
+});
 
 const App = () => {
   return (
--- a/packages/sdk/react-native/example/tsconfig.json
+++ b/packages/sdk/react-native/example/tsconfig.json
@@ -1,9 +1,9 @@
 {
   "extends": "expo/tsconfig.base",
   "compilerOptions": {
-    "jsx": "react",
+    "jsx": "react-jsx",
     "strict": true,
-    "typeRoots": ["./types"]
+    "typeRoots": ["./types"],
   },
-  "exclude": ["e2e"]
+  "exclude": ["e2e"],
 }
--- a/packages/sdk/react-native/src/provider/LDProvider.tsx
+++ b/packages/sdk/react-native/src/provider/LDProvider.tsx
@@ -1,7 +1,5 @@
 import React, { PropsWithChildren, useEffect, useState } from 'react';
 
-import { type LDContext } from '@launchdarkly/js-client-sdk-common';
-
 import ReactNativeLDClient from '../ReactNativeLDClient';
 import { Provider, ReactContext } from './reactContext';
 import setupListeners from './setupListeners';
@@ -9,7 +7,6 @@ import useAppState from './useAppState';
 
 type LDProps = {
   client: ReactNativeLDClient;
-  context?: LDContext;
 };
 
 /**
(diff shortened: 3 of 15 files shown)


packages/sdk/react-native/example/tsconfig.json
```diff
@@ -1,9 +1,9 @@
 {
   "extends": "expo/tsconfig.base",
   "compilerOptions": {
-    "jsx": "react",
+    "jsx": "react-jsx",
```
yusinto, line 4: Fixes typescript complaint: `'React' refers to a UMD global, but the current file is a module.`

packages/sdk/react-native/example/App.tsx
```diff
@@ -8,7 +8,13 @@ import {
 
 import Welcome from './src/welcome';
 
-const featureClient = new ReactNativeLDClient(MOBILE_KEY, AutoEnvAttributes.Enabled);
+const featureClient = new ReactNativeLDClient(MOBILE_KEY, AutoEnvAttributes.Enabled, {
+  debug: true,
```
kinyoklion, line 12: Not sure the example should enable debug.

packages/sdk/react-native/example/tsconfig.json
```diff
@@ -1,9 +1,9 @@
...
     "strict": true,
-    "typeRoots": ["./types"]
+    "typeRoots": ["./types"],
   },
-  "exclude": ["e2e"]
+  "exclude": ["e2e"],
```
yusinto, line 8: Prettier insists that tsconfig.json is a `JSONC` file and so trailing commas should be used.

packages/shared/sdk-client/src/api/LDOptions.ts
```diff
@@ -80,6 +80,13 @@ export interface LDOptions {
    */
   capacity?: number;
 
+  /**
+   * Enables logging for debugging.
```
kinyoklion, line 84:
```suggestion
   * Enables debug logging.
```

Or "Enables the debug log level."

Probably also needs "Has no effect when a custom logger is provided."

packages/sdk/react-native/src/provider/LDProvider.tsx
```diff
@@ -18,25 +15,15 @@ type LDProps = {
  *
  * @param client The ReactNativeLDClient object. Initialize this object separately
  * and then set this prop when declaring the LDProvider.
- * @param context Optional. The LDContext object. If set, the LDProvider will
```
yusinto, line 21: Removing this because this seems extraneous and unnecessary. Customers can already construct and then identify through LDClient which can be awaited so that should be the primary usage.

Secondly supplying the context this way and then having the LDProvider identifying it under the hood hides logic and creates a mysterious magic effect which we have to document and support. The identify call also can't be awaited which is a big problem.

The example app has been setup and tested with a manual identify thus far and everything works good.

Due to these I have decided its best to remove this for the initial v10 release. We can revisit if this is needed in the future.

packages/shared/sdk-client/src/api/LDClient.ts
```diff
@@ -100,12 +100,10 @@ export interface LDClient {
    *
    * @param context
    *   The LDContext object.
-   * @param hash
```
yusinto, line 103: Removing this for now for RN v10 since it is unused. This will be revisited when we do the react sdk next project.

packages/shared/sdk-client/src/api/LDOptions.ts
```diff
@@ -80,6 +80,13 @@ export interface LDOptions {
...
+   *
+   * The default value is false.
+   */
+  debug?: boolean;
```
yusinto, line 88: Callout: this is new. It's an easy way to debug the sdk.

packages/sdk/react-native/src/provider/LDProvider.tsx
mmrj: watching this, because removing this will mean minor docs corrections -- I'll update when this PR is merged

kinyoklion (approved):

PR #369, opened 2024-02-08
feat: React Native SDK major release.

yusinto:
This prepares for the React Native SDK major release as well as the release of common/client-sdk.

### DIFF
--- a/packages/shared/common/README.md
+++ b/packages/shared/common/README.md
@@ -6,15 +6,13 @@
 
 This project contains Typescript classes and interfaces that are applicable to client-side and server-side JavaScript SDKs.
 
-This library is a beta version and should not be considered ready for production use while this message is visible.
-
 ## Contributing
 
 See [Contributing](../CONTRIBUTING.md).
 
 ## Verifying SDK build provenance with the SLSA framework
 
-LaunchDarkly uses the [SLSA framework](https://slsa.dev/spec/v1.0/about) (Supply-chain Levels for Software Artifacts) to help developers make their supply chain more secure by ensuring the authenticity and build integrity of our published SDK packages. To learn more, see the [provenance guide](PROVENANCE.md). 
+LaunchDarkly uses the [SLSA framework](https://slsa.dev/spec/v1.0/about) (Supply-chain Levels for Software Artifacts) to help developers make their supply chain more secure by ensuring the authenticity and build integrity of our published SDK packages. To learn more, see the [provenance guide](PROVENANCE.md).
 
 ## About LaunchDarkly
 
--- a/packages/shared/common/README.md
+++ b/packages/shared/common/README.md
@@ -6,13 +6,15 @@
 
 This project contains Typescript classes and interfaces that are applicable to client-side and server-side JavaScript SDKs.
 
+This library is a beta version and should not be considered ready for production use while this message is visible.
+
 ## Contributing
 
 See [Contributing](../CONTRIBUTING.md).
 
 ## Verifying SDK build provenance with the SLSA framework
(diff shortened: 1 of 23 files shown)


packages/shared/sdk-client/package.json
```diff
@@ -4,7 +4,7 @@
   "type": "commonjs",
   "main": "./dist/index.js",
   "types": "./dist/index.d.ts",
-  "homepage": "https://github.com/launchdarkly/js-core/tree/main/packages/shared/sdk-server",
```
yusinto, line 7: This was a mistake and is fixed now.

packages/shared/sdk-client/src/configuration/Configuration.ts
```diff
@@ -25,7 +25,7 @@ export default class Configuration {
 
   public readonly capacity = 100;
   public readonly diagnosticRecordingInterval = 900;
-  public readonly flushInterval = 2;
```
yusinto, line 28: Callout: I updated this to 30s because 2s seems too frequent. I double checked the iOS and Flutter SDKs also defaults to 30s.

packages/shared/sdk-server-edge/README.md
```diff
@@ -6,15 +6,13 @@
 
 This project contains Typescript classes and interfaces that are applicable to server-side edge SDKs.
 
-This library is a beta version and should not be considered ready for production use while this message is visible.
```
kinyoklion, line 9: You should revert this file and we can address it in another PR.

packages/shared/common/README.md
```diff
@@ -6,15 +6,13 @@
 
 This project contains Typescript classes and interfaces that are applicable to client-side and server-side JavaScript SDKs.
 
-This library is a beta version and should not be considered ready for production use while this message is visible.
```
kinyoklion, line 9: You should revert this file, and we can address it in another PR. (Both these readme notices that need removed from this PR could be done in a `chore:` and not trigger a specific release.

packages/shared/sdk-client/src/configuration/Configuration.ts
kinyoklion: In the past web SDKs have been 2 seconds and mobile SDKs have been 30 seconds. Web SDKs are 2 seconds to hope to catch events due to the ephemeral nature of the browser. 

This reminds me of something, which is flushing as the page closes. Which should be added to flutter. I will make a ticket.

packages/shared/sdk-client/src/api/LDOptions.ts
```diff
@@ -53,14 +56,14 @@ export interface LDOptions {
...
+   * The base uri for the LaunchDarkly server. Most users should use the default value.
    *
-   * Most users should use the default value.
+   * @defaultValue https://clientsdk.launchdarkly.com.
    */
   baseUri?: string;
 
   /**
-   * TODO: bootstrap
+   * TODO: bootstrap. This is not implemented and should not be used.
```
kinyoklion, line 66: Should we remove all the stuff that shouldn't be used and then add it when it can be used?

packages/shared/sdk-client/src/configuration/Configuration.ts
yusinto: Ok. I'll leave this as 30s now. When we implement the react sdk, I can override to 2s in the React SDK itself. Does that sound good?

kinyoklion: Yes.

kinyoklion (approved):

PR #371, opened 2024-02-08
fix: Fix an issue where failed http requests could cause an unhandled promise rejection.

### LINKED ISSUE #370: LaunchDarkly node-server-sdk is crashing node process due to unhandled promise
**Describe the bug**
Currently, if incomingmessage errors due to intermittent network issues such as ECONNRESET, as the promise in the constructor in NodeResponse is unhandled, this bubbles up as `unhandledRejection`, crashing the process.

https://github.com/launchdarkly/js-core/blob/6349b98e70554d8240f0e8d6b1090e4c37bde6eb/packages/sdk/server-node/src/platform/NodeResponse.ts#L31

This is happening quite frequently, affecting our production environment. 

**To reproduce**
This should be reproducible with a sample application that talks to the server but terminates the connection. I did not spend too much effort in getting a repro ready as the code fault and error location is clear.

**Expected behavior**
Rejection should not be unhandled and if request fails, LaunchDarkly SDK should serve the default value gracefully.

**Logs**
Note: promise has been monkey patched to get the stack trace of the promise init.

`unhandledRejection`
```
Error: aborted
    at connResetException (node:internal/errors:720:14)
    at TLSSocket.socketCloseListener (node:_http_client:461:19)
    at TLSSocket.emit (node:events:529:35)
    at TLSSocket.emit (node:domain:489:12)
    at node:net:350:12
    at TCP.done (node:_tls_wrap:657:7)
```

`call site stack trace`
```
at new MonkeyPromise (/var/service/out/lib/app.js:13:19)
    at new NodeResponse (/var/service/node_modules/@launchdarkly/node-server-sdk/dist/src/platform/NodeResponse.js:12:24)
    at ClientRequest.<anonymous> (/var/service/node_modules/@launchdarkly/node-server-sdk/dist/src/platform/NodeRequests.js:86:33)
    at Object.onceWrapper (node:events:632:26)
    at ClientRequest.emit (node:events:517:28)
    at ClientRequest.emit (node:domain:489:12)
    at HTTPParser.parserOnIncomingClient (node:_http_client:700:27)
    at HTTPParser.parserOnHeadersComplete (node:_http_common:119:17)
    at TLSSocket.socketOnData (node:_http_client:541:22)
    at TLSSocket.emit (node:events:517:28)
```

**SDK version**
@launchdarkly/node-server-sdk@8.2.4

node v18

**OS/platform**
ubuntu:jammy@sha256:6042500cf4b44023ea1894effe7890666b0c5c7871ed83a97c36c76ae560bb9b

kinyoklion:
Should fix #370

### DIFF
--- a/packages/sdk/server-node/__tests__/platform/NodeRequests.test.ts
+++ b/packages/sdk/server-node/__tests__/platform/NodeRequests.test.ts
@@ -16,8 +16,14 @@ describe('given a default instance of NodeRequests', () => {
   let resolve: (value: TestRequestData | PromiseLike<TestRequestData>) => void;
   let promise: Promise<TestRequestData>;
   let server: http.Server;
+  let resetResolve: () => void;
+  let resetPromise: Promise<void>;
 
   beforeEach(() => {
+    resetPromise = new Promise((res) => {
+      resetResolve = res;
+    });
+
     promise = new Promise<TestRequestData>((res) => {
       resolve = res;
     });
@@ -43,6 +49,14 @@ describe('given a default instance of NodeRequests', () => {
       } else if ((req.url?.indexOf('404') || -1) >= 0) {
         res.statusCode = 404;
         res.end();
+      } else if ((req.url?.indexOf('reset') || -1) >= 0) {
+        res.statusCode = 200;
+        res.flushHeaders();
+        res.write('potato');
+        setTimeout(() => {
+          res.destroy();
+          resetResolve();
+        }, 0);
       } else {
         res.end(TEXT_RESPONSE);
       }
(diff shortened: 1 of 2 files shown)


packages/sdk/server-node/__tests__/platform/NodeRequests.test.ts
```diff
@@ -115,4 +129,19 @@ describe('given a default instance of NodeRequests', () => {
     expect(serverResult.body).toEqual('BODY TEXT');
     expect(serverResult.headers['sample-header']).toEqual('Some header value');
   });
+
+  it('rejection is handled for response even if not awaited', async () => {
```
kinyoklion, line 133: Running this test against the original code will fail, against this code it works.

packages/sdk/server-node/src/platform/NodeResponse.ts
```diff
@@ -15,6 +15,9 @@ export default class NodeResponse implements platform.Response {
 
   status: number;
 
+  listened: boolean = false;
```
kinyoklion, line 18: Basically we don't attempt to read the body from all HTTP requests we make. Event posting doesn't return a body, so we discard the inner response. 

Unfortunately that inner response may still be rejected.

So we track if the promise has been listened to. If it has been, then we reject like normal. If not we cache the rejection error and throw it when someone listens to the promise.

yusinto (approved):

PR #373, opened 2024-02-08
chore: Implement maintenance release config.

kinyoklion:
This implements support for a maintenance branch for creating node-server-sdk releases.

It removes all builds related to other packages and can only build/release node-server-sdk.


scripts/publish.sh
```diff
@@ -7,6 +7,8 @@ else
     echo "Publishing with prerelease tag."
     yarn workspace $WORKSPACE npm publish --tag prerelease || { echo "npm publish failed" >&2; exit 1; }
   else
-    yarn workspace $WORKSPACE npm publish || { echo "npm publish failed" >&2; exit 1; }
+    # Using a maintenance tag prevents this release from getting tagged with "latest".
+    echo "Publishing with maintenance tag."
+    yarn workspace $WORKSPACE npm publish --tag maintenance || { echo "npm publish failed" >&2; exit 1; }
```
kinyoklion, line 12: If you do not include a tag, then it defaults to latest. This results in `npm i @launcharkly/node-server-sdk` installing the maintenance release. (This was a problem with releaser).

yusinto (approved):
I'm approving, but a little disappointed RP does not support this in a more elegant fashion.

kinyoklion:
> I'm approving, but a little disappointed RP does not support this in a more elegant fashion.

Well, it may, but it depends what is meant by elegant. This version is careful. I could just branch and change the branch name.

PR #374, opened 2024-02-08
fix: Fix an issue where failed http requests could cause an unhandled promise rejection.

kinyoklion:
This is a cherry-pick from the 9.x release.


packages/sdk/server-node/__tests__/platform/NodeRequests.test.ts
```diff
@@ -43,6 +49,14 @@ describe('given a default instance of NodeRequests', () => {
       } else if ((req.url?.indexOf('404') || -1) >= 0) {
         res.statusCode = 404;
         res.end();
+      } else if ((req.url?.indexOf('reset') || -1) >= 0) {
```
yusinto, line 52:
```suggestion
      } else if (req.url?.includes('reset')) {
```

yusinto (approved):
Minor comment on string includes.

PR #387, opened 2024-03-05
fix: Bug in sdk-client where withReasons was not passed to streamer.

### LINKED ISSUE #383: LDEvaluationReason is null, if a flag is not linked to an experiment
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `@launchdarkly/js-sdk-common@2.2.0` for the project I'm working on.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/@launchdarkly/js-sdk-common/dist/api/data/LDEvaluationDetail.d.ts b/node_modules/@launchdarkly/js-sdk-common/dist/api/data/LDEvaluationDetail.d.ts
index c8d689d..fe994e6 100644
--- a/node_modules/@launchdarkly/js-sdk-common/dist/api/data/LDEvaluationDetail.d.ts
+++ b/node_modules/@launchdarkly/js-sdk-common/dist/api/data/LDEvaluationDetail.d.ts
@@ -37,7 +37,8 @@ export interface LDEvaluationDetailTyped<TFlag> {
     variationIndex?: number | null;
     /**
      * An object describing the main factor that influenced the flag evaluation value.
+     * This is null if the flag is not linked to a running experiment.
      */
-    reason: LDEvaluationReason;
+    reason?: LDEvaluationReason | null;
 }
 //# sourceMappingURL=LDEvaluationDetail.d.ts.map
\ No newline at end of file
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

yusinto:
Fixes #383.

### DIFF
--- a/packages/shared/sdk-client/src/LDClientImpl.test.ts
+++ b/packages/shared/sdk-client/src/LDClientImpl.test.ts
@@ -148,7 +148,7 @@ describe('sdk-client object', () => {
 
   test('identify success', async () => {
     defaultPutResponse['dev-test-flag'].value = false;
-    const carContext: LDContext = { kind: 'car', key: 'mazda-cx7' };
+    const carContext: LDContext = { kind: 'car', key: 'test-car' };
 
     await ldc.identify(carContext);
     const c = ldc.getContext();
@@ -156,7 +156,7 @@ describe('sdk-client object', () => {
 
     expect(c).toEqual({
       kind: 'multi',
-      car: { key: 'mazda-cx7' },
+      car: { key: 'test-car' },
       ...autoEnv,
     });
     expect(all).toMatchObject({
@@ -173,7 +173,7 @@ describe('sdk-client object', () => {
   });
 
   test('identify success withReasons', async () => {
-    const carContext: LDContext = { kind: 'car', key: 'mazda-cx7' };
+    const carContext: LDContext = { kind: 'car', key: 'test-car' };
     ldc = new LDClientImpl(testSdkKey, AutoEnvAttributes.Enabled, basicPlatform, {
       logger,
       sendEvents: false,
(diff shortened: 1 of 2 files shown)


packages/shared/sdk-client/src/LDClientImpl.test.ts
```diff
@@ -161,6 +162,34 @@ describe('sdk-client object', () => {
...
+      expect.anything(),
+      '/stream/path',
+      expect.anything(),
+      undefined,
+      expect.anything(),
+    );
+  });
+
+  test('identify success withReasons', async () => {
+    const carContext: LDContext = { kind: 'car', key: 'mazda-cx7' };
```
kinyoklion, line 176: Please use a fictional car brand here.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -321,10 +321,14 @@ export default class LDClientImpl implements LDClient {
       }
     } else {
       this.streamer?.close();
+      let streamUri = this.createStreamUriPath(context);
+      if (this.config.withReasons) {
+        streamUri = `${streamUri}?withReasons=true`;
```
kinyoklion, line 326: I am not sure about this. Generally we default to not including reasons, and only including them if the user configured the SDK to request them. (Which most people do not need, so we don't have to spend the money sending them.) That configuration then still needs to make its way to the streaming and polling configuration.

The reason does have to be optional on the evaluation detail for a client SDK though.

kinyoklion (changes requested):

packages/shared/sdk-client/src/LDClientImpl.ts
yusinto: I understand this. Hence the check on line 325 to only include it if the user explicitly config'd the sdk to do so. Is this what you are talking about?

kinyoklion: I failed at reading.

kinyoklion (approved):

packages/shared/sdk-client/src/LDClientImpl.ts
yusinto: You were right 99% the time I was wrong though. Thanks for your review!

PR #397, opened 2024-03-14
chore: add .sdk_metadata.json to repo

cwaldren-ld:
This adds a (prototype, subject to change) `.sdk_metadata.json` dotfile to the repo. 

It contains metadata about each of the SDKs present here. 

Things to think about:
- Filename, being a dotfile, ok?
- Location, root of repo, makes sense?
- Contents: how to tell the tool which releases belong to each SDK? (currently: specify the tag prefix, and the tool will filter based on that)

### DIFF
--- a/.sdk_metadata.json
+++ b/.sdk_metadata.json
@@ -2,6 +2,7 @@
   "version": 1,
   "sdks": {
     "akamai-base": {
+      "name": "Akamai SDK",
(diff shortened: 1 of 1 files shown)


cwaldren-ld:
I have no idea if the `languages` are correct here.

yusinto (approved):

PR #400, opened 2024-03-18
feat: Implement support for hooks.

kinyoklion:
Adds hook support.


kinyoklion:
Contract tests will fail until there is a release of the test harness.

packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -66,6 +67,23 @@ export interface LDClientCallbacks {
   hasEventListeners: () => boolean;
 }
 
+const BOOL_VARIATION_METHOD_NAME = 'LDClient.boolVariation';
```
kinyoklion, line 70: An alternative to this would be to do something like `LDClient${this.boolVariation.name}` within each method. Doing that would potentially produce lots of temporary variables, so instead a constant is defined for each method.

packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -74,8 +92,6 @@ export default class LDClientImpl implements LDClient {
 
   private featureStore: LDFeatureStore;
 
-  private asyncFeatureStore: AsyncStoreFacade;
```
kinyoklion, line 77: This was dead code left over after the change to callbacks.

packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -738,4 +847,81 @@ export default class LDClientImpl implements LDClient {
...
   }
+
+  private async withHooks(
+    key: string,
+    context: LDContext,
+    defaultValue: unknown,
+    methodName: string,
+    method: () => Promise<LDEvaluationDetail>,
+  ): Promise<LDEvaluationDetail> {
+    if(this.hooks.length == 0) {
```
kinyoklion, line 858: When there are no hooks we just return the non-awaited method, which should yield similar performance to the current implementation.

keelerm84 (approved):

packages/shared/sdk-server/__tests__/LDClient.hooks.test.ts
```diff
@@ -0,0 +1,603 @@
...
+    {},
+    {
+      value: true,
+      reason: Reasons.Fallthrough,
+      variationIndex: 0,
+    },
+  );
+});
+
+it('executes hook stages in the specified order', async () => {
```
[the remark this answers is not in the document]
kinyoklion: I'll add a runtime registered hook. I don't think I need another test. It is "registered order" and "reverse registered order" So if I register a hook C, it would execute after a and b for before and then before, b and a for after.

kinyoklion:
I think I can move the hook execution code out of the client, but I would like to do that in a discreet PR.

PR #393, opened 2024-03-08
fix: Update to major version of @vercel/edge-config

yusinto:
Fixes #379 . I have tested this locally and verified the type errors are fixed and the example application (complete) still works.

PR #402, opened 2024-03-19
chore: Update vercel examples to use workspace packages.

yusinto:
We updated @vercel/edge-config package to use the major version in this [pr](https://github.com/launchdarkly/js-core/pull/393) however I overlooked updating the corresponding example apps to use the workspace versions of our vercel sdk.

### DIFF
--- a/packages/sdk/vercel/examples/complete/package.json
+++ b/packages/sdk/vercel/examples/complete/package.json
@@ -4,7 +4,7 @@
   "license": "MIT",
   "private": true,
   "scripts": {
-    "build": "next build",
+    "build-prod": "next build",
     "dev": "next dev",
     "lint": "next lint",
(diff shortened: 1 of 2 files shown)


packages/sdk/vercel/examples/complete/package.json
```diff
@@ -4,13 +4,13 @@
   "license": "MIT",
   "private": true,
   "scripts": {
-    "build": "next build",
+    "build-prod": "next build",
```
yusinto, line 7: Exclude building these examples when we run `yarn build` at the repo root.

Alternatively, we can also modify the root `build` command to exclude all example*. I'm open to both ideas, but this is the easiest for now hence.

kinyoklion (approved):

PR #407, opened 2024-03-25
fix: Send identify event.

yusinto:


packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -168,20 +168,8 @@ export default class LDClientImpl implements LDClient {
...
-
-        if (changedKeys.length > 0) {
-          this.logger.debug(`Emitting changes from PUT: ${changedKeys}`);
-          // emitting change resolves identify
-          this.emitter.emit('change', context, changedKeys);
-        } else {
-          // manually resolve identify
-          this.logger.debug('Not emitting changes from PUT');
-          identifyResolve();
-        }
```
yusinto, line 184: Dry'd this to `onIdentifyResolve`.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -297,17 +284,14 @@ export default class LDClientImpl implements LDClient {
       return Promise.reject(error);
     }
 
-    const { identifyPromise, identifyResolve } = this.createPromiseWithListeners();
+    this.eventProcessor?.sendEvent(this.eventFactoryDefault.identifyEvent(checkedContext));
```
yusinto, line 287: This was the missing piece to send `identify` event.

packages/shared/sdk-client/src/events/createEventProcessor.ts
```diff
@@ -1,5 +1,4 @@
 import { ClientContext, internal, Platform } from '@launchdarkly/js-sdk-common';
-import { EventProcessor } from '@launchdarkly/js-sdk-common/dist/internal';
```
yusinto, line 2: This is an unrelated bug. I'm not sure why this wasn't caught by the linter.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -263,7 +251,6 @@ export default class LDClientImpl implements LDClient {
 
       this.identifyChangeListener = (c: LDContext, changedKeys: string[]) => {
         this.logger.debug(`change: context: ${JSON.stringify(c)}, flags: ${changedKeys}`);
-        resolve();
```
yusinto, line 266: Resolve not gets explicitly called when needed, rather than here. I think this change makes the code more readable.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -168,20 +168,8 @@ export default class LDClientImpl implements LDClient {
       deserializeData: JSON.parse,
       processJson: async (dataJson: Flags) => {
         this.logger.debug(`Streamer PUT: ${Object.keys(dataJson)}`);
-        const changedKeys = calculateFlagChanges(this.flags, dataJson);
-        this.context = context;
-        this.flags = dataJson;
+        this.onIdentifyResolve(identifyResolve, dataJson, context, 'streamer PUT');
```
kinyoklion, line 171: This looks like it needs to be after:
`await this.platform.storage?.set(canonicalKey, JSON.stringify(this.flags));`

If it is before, and you await identify and then evaluate a flag, you would get old values instead of the new values.

kinyoklion (approved):

PR #409, opened 2024-03-26
fix: Stream retry failure due to previous open connection.

### LINKED ISSUE #405: Streaming not working on react-native android even with Flipper removed
**Describe the bug**

Even with Flipper completely removed from our `react-native` project (as bullet point 4 describes here: <https://github.com/launchdarkly/js-core/tree/main/packages/sdk/react-native#known-android-identify-issue>), streaming still doesn't seem to be working properly for Android. The app doesn't hang on `identify` and initial flags are retrieved so thats good. But any changes to the LD flags through our dashboard are not being reflected live in the Android app as they should. The streaming is working as expected on iOS.

**To reproduce**

I have tried using the exact code from the quickstart section: <https://github.com/launchdarkly/js-core/tree/main/packages/sdk/react-native#quickstart> and the streaming changes are reflected live on iOS but not on android.

I think this issue should be quickly apparent, but please let me know if you need any more reproduction details!

**Expected behavior**

Updating the flags on the LD dashboard should immediately be reflected in the Android app like they are on iOS

**Logs**

On Android on app startup:

```
info: [LaunchDarkly] Opened LaunchDarkly stream connection
info: [LaunchDarkly] Will retry stream connection in 915 milliseconds
```

For comparison on iOS on app startup:
Note: there is no line about retrying the streaming connection in x milliseconds

```
info: [LaunchDarkly] Opened LaunchDarkly stream connection
```

**SDK version**

`"@launchdarkly/react-native-client-sdk": "^10.0.3"`

Not expo, just base `"react-native": "0.72.12"`

**OS/platform**

Tested on Pixel 7 Emulator and Physical Device
Android 14

yusinto:
This should fix #405 or at least provide more logging to debug the issue further if it persists.

### DIFF
--- a/packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.ts
+++ b/packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.ts
@@ -17,7 +17,7 @@ const defaultOptions: EventSourceOptions = {
   timeout: 0,
   withCredentials: false,
   retryAndHandleError: undefined,
-  initialRetryDelayMillis: 5000,
+  initialRetryDelayMillis: 1000,
   logger: undefined,
 };
 
@@ -59,7 +59,7 @@ export default class EventSource<E extends string = never> {
   private xhr: XMLHttpRequest = new XMLHttpRequest();
   private pollTimer: any;
   private retryAndHandleError?: (err: any) => boolean;
-  private initialRetryDelayMillis: number = 5000;
+  private initialRetryDelayMillis: number = 1000;
   private retryCount: number = 0;
   private logger?: any;
 
--- a/packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.ts
+++ b/packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.ts
@@ -17,7 +17,7 @@ const defaultOptions: EventSourceOptions = {
   timeout: 0,
   withCredentials: false,
   retryAndHandleError: undefined,
-  initialRetryDelayMillis: 1000,
+  initialRetryDelayMillis: 5000,
   logger: undefined,
 };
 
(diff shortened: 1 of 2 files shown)


packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.ts
```diff
@@ -17,7 +17,7 @@ const defaultOptions: EventSourceOptions = {
   timeout: 0,
   withCredentials: false,
   retryAndHandleError: undefined,
-  initialRetryDelayMillis: 1000,
+  initialRetryDelayMillis: 5000,
```
yusinto, line 20: Increasing this on a mobile device seems like a good idea so as to not drain power.

packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.ts
```diff
@@ -93,6 +93,7 @@ export default class EventSource<E extends string = never> {
     this.logger?.debug(`[EventSource] Will open new connection in ${delay} ms.`);
     this.dispatch('retry', { type: 'retry', delayMillis: delay });
     this.pollTimer = setTimeout(() => {
+      this.close();
```
yusinto, line 96: `react-native` has its own implementation of `XMLHttpRequest` which throws an error if a previous connection is still `open` and you try to re-open it. Here we are closing the connection first to solve that error.

packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.ts
```diff
@@ -315,6 +316,7 @@ export default class EventSource<E extends string = never> {
         this.onclose();
         break;
       case 'error':
+        this.logger?.debug(`[EventSource][dispatch][ERROR]: ${JSON.stringify(data)}`);
```
yusinto, line 319: This is much needed for debugging.

packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.ts
kinyoklion: I would rather change this independently, unless it contributes to the fix. Because the initial retry is 1 second, when streaming, among the SDKs.

kinyoklion (approved):

kinyoklion (changes requested):
Tests are failing.

kinyoklion (approved):

PR #413, opened 2024-03-29
fix: Fix typings for big segment store factories.

kinyoklion:
The typings were incorrect for the factories. I am not making this a major version as internally the factories were being called with this type. Additionally the integrations don't use the logger which is what they access from these options.

The feature stores had the correct options.

### DIFF
--- a/packages/store/node-server-sdk-dynamodb/__tests__/LDClient.option.test.ts
+++ b/packages/store/node-server-sdk-dynamodb/__tests__/LDClient.option.test.ts
@@ -1,0 +1,11 @@
+import { LDOptions } from '@launchdarkly/node-server-sdk';
+import DynamoDBBigSegmentStoreFactory from '../src/DynamoDBBigSegmentStoreFactory';
+
+it('can construct options with a big segment store', () => {
+  // This is just a typescript test to ensure the typings match.
+  const _: LDOptions = {
+    bigSegments: {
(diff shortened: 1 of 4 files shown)


packages/store/node-server-sdk-dynamodb/__tests__/LDClient.option.test.ts
```diff
@@ -0,0 +1,13 @@
+import { LDOptions } from '@launchdarkly/node-server-sdk';
+
+import DynamoDBBigSegmentStoreFactory from '../src/DynamoDBBigSegmentStoreFactory';
+
+it('can construct options with a big segment store', () => {
+  // This is just a typescript test to ensure the typings match.
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
yusinto, line 7: nit: We have the `varsIgnorePattern` eslint [rule](https://github.com/launchdarkly/js-core/blob/main/.eslintrc.js#L22) which we should correct to ignore unused vars:

```js
'@typescript-eslint/no-unused-vars': [
      'error',
      { ignoreRestSiblings: true, argsIgnorePattern: '^_', varsIgnorePattern: '^__' },
    ],
```

Although vars with dangling underscores violate other eslint rules, which means we should use other patterns, like `xx` maybe.

yusinto (approved):

PR #420, opened 2024-04-03
feat: Add identify timeout to client-sdk.

yusinto:


packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -260,16 +266,30 @@ export default class LDClientImpl implements LDClient {
       this.emitter.on('change', this.identifyChangeListener);
       this.emitter.on('error', this.identifyErrorListener);
     });
+    const timed = timeout(timeoutSeconds, 'identify timed out.');
+    const raced = Promise.race([timed, slow]);
```
yusinto, line 270: Do we want to logger.debug a timeout here ourselves?

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -260,16 +266,33 @@ export default class LDClientImpl implements LDClient {
...
+   * @param pristineContext The LDContext object to be identified.
+   * @param identifyOptions Optional configuration. See {@link LDIdentifyOptions}.
+   *
+   * @returns {Promise<void>}.
+   */
+  async identify(pristineContext: LDContext, identifyOptions?: LDIdentifyOptions): Promise<void> {
+    const identifyTimeout = identifyOptions?.timeoutSeconds ?? this.defaultIdentifyTimeout;
+    if (identifyTimeout > this.defaultIdentifyTimeout) {
+      this.logger.warn(
+        `** Your timeout is too high **. We recommend a max of ${this.defaultIdentifyTimeout}s to prevent blocking your application in the event of network delays.`,
```
yusinto, line 292: Please suggest a better message if you have one.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -261,15 +267,39 @@ export default class LDClientImpl implements LDClient {
       this.emitter.on('error', this.identifyErrorListener);
     });
 
-    return { identifyPromise: p, identifyResolve: res };
+    const timeoutError = 'identify timed out.';
+    const timed = timeout(timeoutSeconds, timeoutError);
+    const raced = Promise.race([timed, slow]).catch((reason: Error) => {
+      if (reason.message === timeoutError) {
+        this.logger.error(reason);
+      }
```
yusinto, line 275: Only log timeout errors because identify errors are handled by `identifyErrorListener`. We can arguably set the log level to debug rather than error here, I'm open to suggestions.

packages/shared/sdk-client/src/api/LDIdentifyOptions.ts
```diff
@@ -0,0 +1,11 @@
+export interface LDIdentifyOptions {
+  /**
+   * Determines when the identify promise resolves if no flags have been
+   * returned from the network. If you use a large timeout and await it, then
+   * any network delays will cause your application to wait a long time before
+   * continuing execution.
+   *
+   * Defaults to 5 seconds.
+   */
+  timeoutSeconds: number;
```
yusinto, line 10: Again, please suggest a better comment if you have one.

packages/shared/common/src/utils/timeout.ts
```diff
@@ -0,0 +1,19 @@
...
+ * @param t Timeout in seconds.
+ * @param taskName Name of task being timed for logging and error reporting.
+ * @param logger {@link LDLogger} object.
+ */
+const timeout = (t: number, taskName: string, logger?: LDLogger) =>
+  new Promise<void>((_res, reject) => {
+    setTimeout(() => {
+      const e = `${taskName} timed out after ${t} seconds.`;
+      logger?.error(e);
+      reject(new Error(e));
```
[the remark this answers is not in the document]
kinyoklion: (I know the node SDK does it, but I am not certain that is a good idea either.)

packages/shared/sdk-client/src/LDClientImpl.ts
louis-launchdarkly: In iOS, the warning is `"%s LDClient.start called with high timeout parameter %s"` while the in-code doc for the parameter is: `- parameter <parameter name>: A <Time data type/unit> that determines when the completion will return if no flags have been returned from the network.
If you use a large <Time data type/unit> and wait for the timeout, then any network delays will cause your application to wait a long time before continuing execution.`

So I think we can remove the recommendation from the warning (in the case the customer does need a higher timeout) as they have no easy way just to turn off this line. Because you already have the explanation in [packages/shared/sdk-client/src/api/LDIdentifyOptions.ts](https://github.com/launchdarkly/js-core/pull/420/files#diff-32377ba2fa37e279d3ef6bcec395b2e2d97f7418f5da095bf77593f528c6edb7)

packages/shared/common/src/utils/timeout.ts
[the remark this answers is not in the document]
yusinto: I don't have an issue with resolving. However the breaking side-effect is `identify` now returns `Promise<void | Error>` rather than just `Promise<void>`.
If we stay with rejecting, the identify promise should be caught by the consumer. This is an accepted safe practice. If we can this clearly I think the risk of breaking apps is low.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -34,9 +36,12 @@ export default class LDClientImpl implements LDClient {
   context?: LDContext;
   diagnosticsManager?: internal.DiagnosticsManager;
   eventProcessor?: internal.EventProcessor;
+  identifyTimeout: number = 5;
```
yusinto, line 39: Harcoded to 5s by default. This will be set in `identify` if provided.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -261,15 +266,34 @@ export default class LDClientImpl implements LDClient {
...
+   * @returns {Promise<void>}.
+   */
+  async identify(pristineContext: LDContext, identifyOptions?: LDIdentifyOptions): Promise<void> {
+    if (identifyOptions?.timeout) {
+      this.identifyTimeout = identifyOptions.timeout;
+    }
+
+    if (this.identifyTimeout > this.highTimeoutThreshold) {
+      this.logger.warn('identify called with high timeout parameter.');
+    }
```
yusinto, line 295: shout: This is the core change.

louis-launchdarkly (approved):
If I understand correctly, because JS support default parameter (so it is not a breaking signature change), and the identify method can already error for other reason, adding and defaulting the timeout doesn't count as a behavior breaking change.

If that is the case, this should be the fix and approved.

PR #425, opened 2024-04-04
chore: Improved identify timeout comments.

yusinto:


packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -282,8 +282,13 @@ export default class LDClientImpl implements LDClient {
    *
    * @param pristineContext The LDContext object to be identified.
    * @param identifyOptions Optional configuration. See {@link LDIdentifyOptions}.
+   * @returns
+   *   A Promise which resolves when the flag values for the specified context
+   *   are available.
    *
-   * @returns {Promise<void>}.
+   *   The promise rejects if the identify timeout is exceeded. In client SDKs
```
kinyoklion, line 289: Looks like it is also rejected on a bad context.
```
 if (!checkedContext.valid) {
      const error = new Error('Context was unspecified or had no key');
      this.logger.error(error);
      this.emitter.emit('error', context, error);
      return Promise.reject(error);
    }
```

I thought there was another case as well.

kinyoklion:
```
     this.identifyErrorListener = (c: LDContext, err: Error) => {
        this.logger.debug(`error: ${err}, context: ${JSON.stringify(c)}`);
        reject(err);
      };
```

yusinto: I found other bugs related to identify promise rejections when investigating this. I'll report back with an update once I resolve those bugs. Thank you for your patience.

yusinto:
Closing this in favor of #426 .

PR #426, opened 2024-04-08
fix: Identify incorrectly rejected in client-sdk

yusinto:
This is a bugfix. Previously, any `error` event from the emitter rejects the identify promise, which is incorrect.


packages/shared/common/src/utils/timedPromise.ts
```diff
@@ -1,17 +1,13 @@
...
  * @param t Timeout in seconds.
  * @param taskName Name of task being timed for logging and error reporting.
- * @param logger {@link LDLogger} object.
  */
-const timedPromise = (t: number, taskName: string, logger?: LDLogger) =>
+const timedPromise = (t: number, taskName: string) =>
   new Promise<void>((_res, reject) => {
     setTimeout(() => {
       const e = `${taskName} timed out after ${t} seconds.`;
-      logger?.error(e);
```
yusinto, line 14: We should not log here because a Promise once created will always run so this will always run, which is futile.

packages/shared/mocks/src/streamingProcessor.ts
```diff
@@ -28,11 +28,9 @@ export const setupMockStreamingProcessor = (
       start: jest.fn(async () => {
         if (shouldError) {
           setTimeout(() => {
-            const unauthorized: LDStreamingError = {
-              code: 401,
-              name: 'LaunchDarklyStreamingError',
-              message: 'test-error',
-            };
```
yusinto, line 35: This error mock should be an error object and not a plain js object.

packages/shared/sdk-client/src/LDClientImpl.test.ts
```diff
@@ -115,37 +115,6 @@ describe('sdk-client object', () => {
...
-    ldc.flags['dev-test-flag'].deleted = true;
-    const flag = ldc.variationDetail('dev-test-flag', 'deleted');
-
-    expect(flag).toEqual({
-      reason: { errorKind: 'FLAG_NOT_FOUND', kind: 'ERROR' },
-      value: 'deleted',
-      variationIndex: null,
-    });
-  });
-
```
yusinto, line 148: Variation tests are moved to its own test file to make this file size manageable.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -244,32 +248,20 @@ export default class LDClientImpl implements LDClient {
...
-      if (this.identifyErrorListener) {
-        this.emitter.off('error', this.identifyErrorListener);
-      }
-
-      this.identifyChangeListener = (c: LDContext, changedKeys: string[]) => {
-        this.logger.debug(`change: context: ${JSON.stringify(c)}, flags: ${changedKeys}`);
-      };
-      this.identifyErrorListener = (c: LDContext, err: Error) => {
-        this.logger.debug(`error: ${err}, context: ${JSON.stringify(c)}`);
-        reject(err);
```
yusinto, line 262: This was the bug where any error will reject the identify promise.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -244,32 +248,20 @@ export default class LDClientImpl implements LDClient {
...
-    const timed = timedPromise(timeout, 'identify', this.logger);
-    const raced = Promise.race([timed, slow]);
+    const timed = timedPromise(timeout, 'identify');
+    const raced = Promise.race([timed, slow]).catch((e) => {
+      this.logger.error(`identify error: ${e}`);
+      throw e;
+    });
 
-    return { identifyPromise: raced, identifyResolve: res };
+    return { identifyPromise: raced, identifyResolve: res, identifyReject: rej };
```
yusinto, line 264: Include `reject` so we can manually reject the identify promise if needed.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -340,7 +340,7 @@ export default class LDClientImpl implements LDClient {
         this.createStreamListeners(context, checkedContext.canonicalKey, identifyResolve),
         this.diagnosticsManager,
         (e) => {
-          this.logger.error(e);
+          identifyReject(e);
```
yusinto, line 343: If streamer errors, reject the identify promise. This maybe more relevant during the initial connection to streamer and when PUT is being processed. The other PATCH/DELETE commands come afterwards and by that time the identify promise will already be done so any errors that happens then is moot to the identify promise.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -420,7 +420,6 @@ export default class LDClientImpl implements LDClient {
       const error = new LDClientError(
         `Unknown feature flag "${flagKey}"; returning default value ${defVal}.`,
       );
-      this.logger.error(error);
       this.emitter.emit('error', this.context, error);
```
yusinto, line 423: Emit `error` now logs the error so there's no need to duplicate here.

packages/shared/sdk-client/src/api/LDClient.ts
```diff
@@ -112,7 +112,15 @@ export interface LDClient {
...
-   *   A Promise which resolves when the flag values for the specified context are available.
+   *    A Promise which resolves when the flag values for the specified
+   * context are available. It rejects when:
+   *
+   * 1. The context is unspecified or has no key.
+   *
+   * 2. The identify timeout is exceeded. In client SDKs this defaults to 5s.
+   * You can customize this timeout with {@link LDIdentifyOptions | identifyOptions}.
+   *
+   * 3. A streaming error occurs.
```
yusinto, line 123: Added these improved comments to include rejection scenarios. Will close #425 in favor of this.

kinyoklion: I would make 3 more abstract to account for polling or other initialization methods.

`A network error is encountered during initialization.`
or
`An error is encountered accessing flag values from LaunchDarkly.`

kinyoklion (approved):
Approved with comment.

PR #427, opened 2024-04-08
fix: Improved retry reset logic to wait before resetting.

yusinto:


packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.ts
```diff
@@ -98,6 +99,18 @@ export default class EventSource<E extends string = never> {
     }, delay);
   }
 
+  /**
+   * Reset after 1 minute.
+   *
+   * @private
+   */
+  private resetRetryCount() {
```
kinyoklion, line 107: This implements a different style of reset than other SDKs.

If you connect then it will reset the retry count after 1 minute regardless how many other problems happen later. Also it can have multiple concurrent timers running.

The logic would typically not use a timer, but it could be made to use a timer by:
Cancel the timer if a timer already exists and start a new timer.
Cancel the timer if there is an error.
(You shouldn't have to clear the timer in the handler, because it should have already ran.)

Generally it has been done with a timestamp. 

Set a timestamp when you connect.

When you have an error check to see if there is a timestamp and it is more more than 1 minute ago. Clear the timestamp and reset the retry count.

kinyoklion (changes requested):

packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.ts
kinyoklion: I think we need to at least clear the timeout on an error.

yusinto: No worries I'll rewrite this to use timestamps as you suggest.

kinyoklion (approved):

kinyoklion (changes requested):
Un-approving for test failure.

yusinto:
I detected some scenarios which are not being retried while fixing this. I'm investigating and fixing those, I'll close this PR for now in favor of that.

PR #435, opened 2024-04-12
feat: Support publishing to jsr.

yusinto:


scripts/publish.sh
```diff
@@ -8,6 +8,13 @@ else
     echo "Publishing with prerelease tag."
     npm publish --tag prerelease --provenance --access public "./$WORKSPACE_PATH/package.tgz" || { echo "npm publish failed" >&2; exit 1; }
   else
+    if [ -f "./$WORKSPACE_PATH/jsr.json" ]; then
+      echo "Publishing to jsr."
+      cd $WORKSPACE_PATH
+      npx jsr publish || { echo "jsr publish failed" >&2; exit 1; }
+    fi
```
yusinto, line 15: This is repeatable. If the version specified in `jsr.json` is already published to JSR, this will [skip publishing](https://jsr.io/docs/publishing-packages#publishing-from-github-actions) without failure and do nothing.

Screenshot_2024-04-11_at_7_56_22 PM

scripts/publish.sh
```diff
@@ -8,6 +8,13 @@ else
     echo "Publishing with prerelease tag."
     npm publish --tag prerelease --provenance --access public "./$WORKSPACE_PATH/package.tgz" || { echo "npm publish failed" >&2; exit 1; }
```
yusinto, line 9: Jsr does not seem to support tagging.

scripts/publish.sh
```diff
@@ -8,6 +8,13 @@ else
...
   else
+    if [ -f "./$WORKSPACE_PATH/jsr.json" ]; then
+      echo "Publishing to jsr."
+      cd $WORKSPACE_PATH
+      npx jsr publish || { echo "jsr publish failed" >&2; exit 1; }
+    fi
+        
+    echo "Publishing to npm."
     npm publish --provenance --access public "./$WORKSPACE_PATH/package.tgz" || { echo "npm publish failed" >&2; exit 1; }
```
yusinto, line 18: According to the [docs](https://jsr.io/docs/trust#provenance-and-trust), we get provenance for free if we use `npx jsr publish` and github-actions. These two conditions are required for provenance.

scripts/publish.sh
```diff
@@ -8,6 +8,13 @@ else
     echo "Publishing with prerelease tag."
     npm publish --tag prerelease --provenance --access public "./$WORKSPACE_PATH/package.tgz" || { echo "npm publish failed" >&2; exit 1; }
   else
+    if [ -f "./$WORKSPACE_PATH/jsr.json" ]; then
+      echo "Publishing to jsr."
+      cd $WORKSPACE_PATH
```
kinyoklion, line 13: I suspect this would break npm publishing and it needs to pop this directory.

PR #437, opened 2024-04-12
feat: Add support for optional timeout for waitForInitialization.

kinyoklion:


packages/shared/mocks/src/streamingProcessor.ts
```diff
@@ -14,6 +14,7 @@ export const setupMockStreamingProcessor = (
   patchResponseJson?: any,
   deleteResponseJson?: any,
   errorTimeoutSeconds: number = 0,
+  initTimeoutMs: number = 0,
```
kinyoklion, line 17: Allow an initialization delay for testing.

packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -248,27 +253,44 @@ export default class LDClientImpl implements LDClient {
...
   }
 
-  waitForInitialization(): Promise<LDClient> {
+  waitForInitialization(options?: LDWaitForInitializationOptions): Promise<LDClient> {
     // An initialization promise is only created if someone is going to use that promise.
     // If we always created an initialization promise, and there was no call waitForInitialization
     // by the time the promise was rejected, then that would result in an unhandled promise
     // rejection.
 
+    if (options?.timeout === undefined) {
```
kinyoklion, line 262: Existing users will start getting this warning.

packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -248,27 +253,51 @@ export default class LDClientImpl implements LDClient {
...
+  waitForInitialization(options?: LDWaitForInitializationOptions): Promise<LDClient> {
     // An initialization promise is only created if someone is going to use that promise.
     // If we always created an initialization promise, and there was no call waitForInitialization
     // by the time the promise was rejected, then that would result in an unhandled promise
     // rejection.
 
+    // If there is no update processor, then there is functionally no initialization
+    // so it is fine not to wait.
+
+    if (options?.timeout === undefined && this.updateProcessor !== undefined) {
```
kinyoklion, line 265: Existing users will start getting this warning.

kinyoklion:
I'll be holding off merging this until next week to sort documentation and examples.

packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -858,4 +887,33 @@ export default class LDClientImpl implements LDClient {
...
+    basePromise: Promise<LDClient>,
+    timeout?: number,
+    logger?: LDLogger,
+  ): Promise<LDClient> {
+    if (timeout) {
+      const timeoutPromise = timedPromise(timeout, 'waitForInitialization');
+      return Promise.race([basePromise, timeoutPromise.then(() => this)]).catch((reason) => {
+        if (reason instanceof LDTimeoutError) {
+          logger?.error(reason.message);
+        }
```
yusinto, line 913: It will be good to have test coverage for this.

packages/shared/sdk-server/src/LDClientImpl.ts
yusinto:
```suggestion
    if (!options?.timeout && this.updateProcessor) {
```

packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -248,27 +253,51 @@ export default class LDClientImpl implements LDClient {
...
+    }
+    if (
+      options?.timeout !== undefined &&
+      options?.timeout > HIGH_TIMEOUT_THRESHOLD &&
+      this.updateProcessor !== undefined
+    ) {
+      this.logger?.warn(
+        'The waitForInitialization function was called with a timeout greater than ' +
+          `${HIGH_TIMEOUT_THRESHOLD} seconds. We recommend a timeout of less than ` +
+          `${HIGH_TIMEOUT_THRESHOLD} seconds.`,
```
yusinto, line 279: I think we should make client and server warnings consistent? Or at least mostly consistent. Because at the moment the client sdk says this:

```ts
this.logger.warn('identify called with high timeout parameter.');
```

kinyoklion: I can change the other one, but it will need to be a PR. I didn't make this one match because I didn't think it was actionable without looking into the code to establish what "high" meant.

packages/shared/sdk-server/src/LDClientImpl.ts
kinyoklion: This would not be logically the same. The timeout can be 0, which is not the same as not being defined. Though potentially we could argue 0 isn't good.

packages/shared/sdk-server/src/LDClientImpl.ts
kinyoklion: (Combining client and server side PRs produces less than desirable change notes.)

yusinto (approved):
lgtm. Thank u!

packages/shared/sdk-server/src/LDClientImpl.ts
kinyoklion: I'll add this test.

kinyoklion: Added.

PR #440, opened 2024-04-16
chore: Update gha checkout and node versions.

yusinto:
Update checkout and setup-node actions to v4. Update node versions to 20 which uses npm v10 which supports provenance.

Screenshot_2024-04-16_at_11_41_26 AM


actions/install-npm-version/action.yml
yusinto, line 1: node v20 comes with npm v10 which supports provenance out of the box so we don't need this anymore.

.github/workflows/server-node.yml
```diff
@@ -17,14 +17,14 @@ jobs:
     strategy:
       matrix:
         # Node versions to run on.
-        version: [16, 18]
+        version: [18, 20]
```
kinyoklion, line 20:
```suggestion
        version: [18, 24]
```

I think we should be able to do this now.

kinyoklion (approved):

kinyoklion:
Is `ga` supposed to be `gh` in the title?

yusinto:
> Is `ga` supposed to be `gh` in the title?

I was trying to say github-actions. I'll re-write to say gha to make it clearer and less confusing.

.github/workflows/server-node.yml
```diff
@@ -17,14 +17,14 @@ jobs:
     strategy:
       matrix:
         # Node versions to run on.
-        version: [16, 18]
+        version: [18, 24]
```
kinyoklion, line 20:
```suggestion
        version: [18, 21]
```
I put the wrong version in my suggestion. Sorry about that. :pray:

PR #445, opened 2024-04-23
fix: Remove erroneous jsr.json check.

yusinto:
I ran and tested this locally now and I am very confident this should work!


scripts/publish-jsr.sh
```diff
@@ -6,9 +6,10 @@ if [ -f "./$WORKSPACE_PATH/jsr.json" ]; then
   if $LD_RELEASE_IS_DRYRUN ; then
     echo "Doing a dry run of jsr publishing."
     npx jsr publish --dry-run --allow-dirty || { echo "jsr publish failed" >&2; exit 1; }
-  elif [ -f "./$WORKSPACE_PATH/jsr.json" ]; then
```
yusinto, line 9: This check is erroneous and unnecessary because in line 4 we already cd'd into $WORKSPACE_PATH.

louis-launchdarkly (approved):

PR #449, opened 2024-04-26
fix: Produce a warning when track is called with a non-numeric metric value.

kinyoklion:


packages/shared/sdk-client/src/LDClientImpl.events.test.ts
```diff
@@ -73,4 +73,86 @@ describe('sdk-client object', () => {
...
+  it('produces a warning when the metric value is non-numeric', async () => {
+    // @ts-ignore
+    await ldc.identify(carContext);
+    // @ts-ignore
+    ldc.track('the-event', { the: 'data' }, '12');
+
+    expect(logger.warn).toHaveBeenCalled();
+    expect(logger.warn).toHaveBeenCalledWith(
+      expect.stringMatching(/was called with a non-numeric/),
+    );
```
yusinto, line 156: The second assertion implies the first, so we can safely remove the first assertion.

```suggestion
    expect(logger.warn).toHaveBeenCalledWith(
      expect.stringMatching(/was called with a non-numeric/),
    );
```

yusinto (approved):

PR #450, opened 2024-04-26
fix: Handle missing message with valid string.

kinyoklion:


packages/shared/common/src/utils/http.ts
```diff
@@ -49,7 +49,7 @@ export function httpErrorMessage(
   if (err.status) {
     desc = `error ${err.status}${err.status === 401 ? ' (invalid SDK key)' : ''}`;
   } else {
-    desc = `I/O error (${err.message || err})`;
```
kinyoklion, line 52: This condition should not happen after js-eventsource 2.0.2 is released. But if the condition ever happens again we can just say that we don't know what the error was. When we see this, then we know we have made a mistake in the event source implementation.

yusinto (approved):

Issue #324, opened 2023-12-05
`waitForInitialization()` returns a internal `LDClient` type than the type returned from `init()`


Velitheda:
**Describe the bug**

`waitForInitialization` is returning an internal `LDClient` type, instead of the `LDClient` that is exported, causing TypeScript compile errors if I want to use the output from the function.

**To reproduce**

```
import { init } from '@launchdarkly/node-server-sdk'
import type { LDClient } from '@launchdarkly/node-server-sdk'
import type { LDClient as LDClientCommon } from '@launchdarkly/js-server-sdk-common'

const ldClient: LDClient = init('sdk-key', {})

// Should be LDClient:
const initializedClient: LDClientCommon = await ldClient.waitForInitialization()
```

**Expected behavior**

`waitForInitialization` should return the `LDClient` type from `node-server-sdk`, not `js-server-sdk-common`

**SDK version**
`v9.0.4`

**Language version, developer tools**
Typescritp `v4.9.5` on Node `v18.16.0`

**OS/platform**
macOS Ventura Version `13.5.2`

**Additional context**
I can currently get around the issue by doing the following:

```
const ldClient: LDClient = init('sdk-key', {})

await ldClient.waitForInitialization()

if (!ldClient.initialized()) {
    logger.error('LaunchDarkly client failed to initialize')
}

return ldClient
```

keanan:
Hello @kinyoklion,

We are facing a similar issue where trying to add inspectors as per the [docs](https://docs.launchdarkly.com/sdk/features/inspectors/?q=inspector) but getting errors from the client which is returned. 

The type declarations are correctly [added](https://github.com/launchdarkly/js-core/blob/main/packages/shared/sdk-client/src/api/LDOptions.ts) but this is not what is being returned. 

Should we ignore the TS error and use the following?
```
const client = ld.init("SDK-KEY", {
      logger: log,
      inspectors: [
        {
          type: 'flag-used',
          name: 'inspector',
         ...
        },
      ],
    });
```
Thanks

PR #459, opened 2024-05-10
fix(types): LDClient.waitForInitialization resolves this

SevenOutman:
**Requirements**

**Related issues**

https://github.com/launchdarkly/js-core/issues/324

Type the return type with `this` keyword

**Additional context**

In my use case, the problem was that the return type of `waitForInitialization` is not compatible with `ld.LDClient`

```typescript
import * as ld from '@launchdarkly/node-server-sdk';

export function initLdClient(): Promise<ld.LDClient> {
  return ld.init(process.env.LAUNCH_DARKLY_SDK_KEY as string)
           .waitForInitialization(); // Type error
}
```

### DIFF
--- a/packages/shared/sdk-server/src/LDClientImpl.ts
+++ b/packages/shared/sdk-server/src/LDClientImpl.ts
@@ -103,13 +103,13 @@ export default class LDClientImpl implements LDClient {
 
   private evaluator: Evaluator;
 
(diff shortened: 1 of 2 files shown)


chadxz:
In the docs for the waitForInitialization method, it says:

> It is not recommended to use the returned client object. It will be removed in a future version.

FYI.

SevenOutman:
Understood. Thanks both! I'm closing the PR.

Issue #462, opened 2024-05-17
Incompatible with RN New Architecture


artyorsh:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `@launchdarkly/react-native-client-sdk@10.1.3` for the project I'm working on.

Enabling new architecture in react native project produces runtime errors while trying to access nullable `NativeModules.SettingsManager`.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/@launchdarkly/react-native-client-sdk/dist/src/platform/locale.js b/node_modules/@launchdarkly/react-native-client-sdk/dist/src/platform/locale.js
index 725ef98..c4cdcdd 100644
--- a/node_modules/@launchdarkly/react-native-client-sdk/dist/src/platform/locale.js
+++ b/node_modules/@launchdarkly/react-native-client-sdk/dist/src/platform/locale.js
@@ -5,7 +5,7 @@ import { NativeModules, Platform } from 'react-native';
  * https://dev.to/medaimane/localization-and-internationalization-in-react-native-reaching-global-audiences-3acj
  */
 const locale = Platform.OS === 'ios'
-    ? NativeModules.SettingsManager.settings.AppleLocale // iOS
+    ? NativeModules.SettingsManager?.settings.AppleLocale // iOS
     : (_a = NativeModules.I18nManager) === null || _a === void 0 ? void 0 : _a.localeIdentifier;
 export default locale;
 //# sourceMappingURL=locale.js.map
\ No newline at end of file
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

yusinto:
Thanks for submitting this. We are looking into it. Internally logged as 244927.

PR #471, opened 2024-05-23
fix: Added iOS SettingsManager null check.

yusinto:
Fixes this [gh issue](https://github.com/launchdarkly/js-core/issues/462).

Enabling new architecture in react native project produces runtime errors while trying to access nullable `NativeModules.SettingsManager`.

### DIFF
--- a/packages/sdk/react-native/src/platform/locale.ts
+++ b/packages/sdk/react-native/src/platform/locale.ts
@@ -6,7 +6,7 @@ import { NativeModules, Platform } from 'react-native';
  */
 const locale =
   Platform.OS === 'ios'
-    ? NativeModules.SettingsManager.settings.AppleLocale // iOS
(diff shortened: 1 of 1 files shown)


kinyoklion (approved):

quinnturner:
For `_a.settings.AppleLocale`, `settings` can seemingly also be null/undefined using the new arch, so I think we also need a check `_a.settings?.AppleLocale`

kinyoklion:
> For `_a.settings.AppleLocale`, `settings` can seemingly also be null/undefined using the new arch, so I think we also need a check `_a.settings?.AppleLocale`

Hello @quinnturner, 

Thank you for the note. Would you mind filing an issue?

Thank you,
Ryan

quinnturner:
Created an issue: https://github.com/launchdarkly/js-core/issues/757

PR #475, opened 2024-05-31
fix: LRU cache memory leak.

### LINKED ISSUE #474: Memory Leak Inside LruCache
**Describe the bug**
It appears that the [keyMap](https://github.com/launchdarkly/js-core/blob/8ff641be8b32f24884457bd506566e283cf04e40/packages/shared/sdk-server/src/cache/LruCache.ts#L29) used to store indexes for retrieving [values](https://github.com/launchdarkly/js-core/blob/8ff641be8b32f24884457bd506566e283cf04e40/packages/shared/sdk-server/src/cache/LruCache.ts#L19) is a memory leak. When [clear()](https://github.com/launchdarkly/js-core/blob/8ff641be8b32f24884457bd506566e283cf04e40/packages/shared/sdk-server/src/cache/LruCache.ts#L105-L113) runs on the cache, it clears out _most_ of the data stored except for the `keyMap` which can grow infinitely as [unique keys get added into the cache](https://github.com/launchdarkly/js-core/blob/8ff641be8b32f24884457bd506566e283cf04e40/packages/shared/sdk-server/src/cache/LruCache.ts#L68). It also seems like the indexes stored in the `keyMap` would become stale on `clear()` since they would no longer point to the correct index in the `values` array.

**To reproduce**
If you call `client.variation(key, context, defaultValue)` many times passing unique keys, the `keyMap` will continue to grow infinitely even if `clear()` were to be called.

**Expected behavior**
I would expect a call to `this.keyMap.clear()` in the [LruCache's clear()](https://github.com/launchdarkly/js-core/blob/8ff641be8b32f24884457bd506566e283cf04e40/packages/shared/sdk-server/src/cache/LruCache.ts#L105-L113) function like so:
```js
  clear() {
    ...
    this.keyMap.clear();
  }
```
Local testing of this has proven to reduce the memory as we're clearing out the `keyMap` periodically.

**SDK version**
@launchdarkly/js-server-sdk-common version 1.2.3

Node 18.19.0

**OS/platform**
macOS 14.4

kinyoklion:
Fixes #474

### DIFF
--- a/packages/shared/sdk-server/__tests__/cache/LruCache.test.ts
+++ b/packages/shared/sdk-server/__tests__/cache/LruCache.test.ts
@@ -46,14 +46,19 @@ it('when it has been cleared', () => {
     expect(lruCache.get(i.toString())).toBeUndefined();
   }
 
-  // Fill the cache with integers again.
+  // Fill the cache with integers in a different range.
   for (let i = 0; i < max; i += 1) {
-    lruCache.set(`${i}`, i);
+    lruCache.set(`${i + 100}`, i);
   }
 
   // Check they are all there.
   for (let i = 0; i < max; i += 1) {
-    expect(lruCache.get(i.toString())).toEqual(i);
+    expect(lruCache.get(`${i + 100}`)).toEqual(i);
+  }
+
+  // Check old keys don't associate with new values.
+  for (let i = 0; i < max; i += 1) {
+    expect(lruCache.get(i.toString())).toBeUndefined();
   }
 });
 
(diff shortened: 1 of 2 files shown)


packages/shared/sdk-server/__tests__/cache/LruCache.test.ts
```diff
@@ -46,14 +46,19 @@ it('when it has been cleared', () => {
...
+    lruCache.set(`${i + 100}`, i);
   }
 
   // Check they are all there.
   for (let i = 0; i < max; i += 1) {
-    expect(lruCache.get(i.toString())).toEqual(i);
+    expect(lruCache.get(`${i + 100}`)).toEqual(i);
+  }
+
+  // Check old keys don't associate with new values.
```
kinyoklion, line 59: Without clearing the key map these would reference the old values.

keelerm84 (approved):

PR #476, opened 2024-06-03
feat: Add support for a cancel-able timeout.

kinyoklion:
The current timedPromise doesn't provide access to the handle, so it cannot cancel the timer. The timer itself will not have a harmful result, but it is an open handle. Open handles often prevent runtimes from exiting. For instance initializing something like the node server with a really long timeout may keep the process open for that entire timeout even if the SDK initialized.

I am adding this as a new function, which node server will then migrate to. RN can of course migrate as well, but this change will be independent. 

I noticed this unit testing the openfeature-node-provider with a timeout and each test had a remaining open handle.


kinyoklion:
It is possible that this could be done a little more cleanly with an abort controller, but I do not want to add that API has a dependency into common layers.

PR #484, opened 2024-06-20
feat: Scaffold browser telemetry project.

kinyoklion:
This is basic prototype code for error monitoring.


louis-launchdarkly (approved):
This has the Collectors, Inspectors, and the breadcrumbs that we discussed in the plan so far.

PR #486, opened 2024-06-24
feat: Basic session replay prototype.

kinyoklion:
These are prototype changes targeting a prototype branch.

Adds the ability to capture sessions as a custom event using rrweb.


package.json
```diff
@@ -54,5 +54,8 @@
     "typedoc": "0.25.0",
     "typescript": "5.1.6"
   },
+  "resolutions": {
```
kinyoklion, line 57: This should be in main already. Unrelated to these changes.

packages/telemetry/browser-telemetry/package.json
```diff
@@ -41,6 +41,9 @@
   "bugs": {
     "url": "https://github.com/launchdarkly/js-core/issues"
   },
+  "dependencies": {
```
kinyoklion, line 44: For the real-deal this will probably be a peer dependency.

packages/telemetry/browser-telemetry/src/collectors/rrweb/SessionReplay.ts
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Experimental capture of sessions using rrweb.
+ */
+export default class SessionReplay implements Collector {
+  telemetry?: BrowserTelemetry;
+  buffer: SessionBuffer;
+
+  constructor(checkoutEveryNth: number, numBuffers: number) {
+    applyPatches();
```
kinyoklion, line 16: If this is in use then apply the monkey patching for XmlHttpRequests.

packages/telemetry/browser-telemetry/src/collectors/rrweb/patches.ts
```diff
@@ -0,0 +1,33 @@
+// This file contains temporary patches that will be accomplished
+// with SDK modifications in the future.
+
+const sessionId = crypto.randomUUID();
+const sessionTag = `session-id/${sessionId}`;
+
+class HeaderModifyingXMLHttpRequest extends XMLHttpRequest {
```
kinyoklion, line 7: For production the JS SDK would directly add this tag itself. (Or an entirely different header).

packages/telemetry/browser-telemetry/src/collectors/rrweb/SessionBuffer.ts
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * A session buffer is a circular buffer of buffers. Each buffer has a fixed size and is intended
+ * to match the "checkout" size of rrweb. Each individual buffer should contain a valid replay
+ * and then the sum of all the buffers should also be a valid replay containing each of the other
+ * session chunks.
+ *
+ * The buffer can continuously capture events while always remaining in a playable state
+ * and dropping old events.
+ */
```
kinyoklion, line 11: I need to validate that rrweb is producing these on the cadence I expect. Not for this PR though.

package.json
```diff
@@ -20,13 +20,13 @@
     "packages/store/node-server-sdk-redis",
     "packages/store/node-server-sdk-dynamodb",
     "packages/telemetry/node-server-sdk-otel",
-    "packages/telemetry/browser-telemetry",
-    "packages/tooling/jest"
+    "packages/tooling/jest",
+    "packages/telemetry/browser-telemetry"
```
karayusuf, line 24: Is the order change important here? I'm guessing no, but figured I'd ask.

karayusuf (approved):

package.json
kinyoklion: Not really, but I ended up adjusting them to match the order they were added originally. Thank you though.

PR #489, opened 2024-06-25
feat: Add continuous session replay capture.

kinyoklion:
Prototype code.

Usage:
```
  const session = new SessionReplay();
  const telemetry = initializeTelemetry({collectors: [session]});
  const client = initialize('client-side-id', {
    kind: 'user',
    key: 'example-user'
  }, {
    inspectors: telemetry.inspectors(),
  });
  telemetry.register(client);
```

### DIFF
--- a/packages/telemetry/browser-telemetry/src/collectors/rrweb/ContinuousReplay.ts
+++ b/packages/telemetry/browser-telemetry/src/collectors/rrweb/ContinuousReplay.ts
@@ -1,0 +1,80 @@
+import * as rrweb from 'rrweb';
+
+import { BrowserTelemetry } from '../../api/BrowserTelemetry';
+import { Collector } from '../../api/Collector';
+import { sessionId } from './patches';
+import { ContinuousCapture } from './SessionReplayOptions';
+
+export default class ContinuousReplay implements Collector {
+  private telemetry?: BrowserTelemetry;
+  // TODO: Use a better buffer.
+  private buffer: any[] = [];
+  private stopper?: () => void;
+  private visibilityHandler: any;
+  private timerHandle: any;
+  private index: number = 0;
+
+  constructor(private readonly options: ContinuousCapture) {
+    this.visibilityHandler = () => {
+      this.handleVisibilityChange();
+    };
+
+    document.addEventListener('visibilitychange', this.visibilityHandler, true);
+
+    this.timerHandle = setInterval(() => {
+      // If there are only 2 events, then we just have a snapshot.
+      // We can wait to capture until there is some activity.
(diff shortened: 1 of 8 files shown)


packages/telemetry/browser-telemetry/src/collectors/rrweb/SessionReplay.ts
```diff
@@ -3,41 +3,46 @@ import * as rrweb from 'rrweb';
...
 /**
  * Experimental capture of sessions using rrweb.
  */
 export default class SessionReplay implements Collector {
-  telemetry?: BrowserTelemetry;
-  buffer: SessionBuffer;
+  impl: Collector;
 
-  constructor(checkoutEveryNth: number, numBuffers: number) {
+  constructor(options?: SessionReplayOptions) {
```
kinyoklion, line 25: I refactored this to a PIMPL style where the underlying recorder is contingent on the options. Both the rolling and continuous capture mechanisms exists and it uses the continuous mechanism by default.

packages/telemetry/browser-telemetry/src/collectors/rrweb/ContinuousReplay.ts
```diff
@@ -0,0 +1,81 @@
+import * as rrweb from 'rrweb';
```
kinyoklion, line 1: This mechanism captures a payload every X milliseconds. Defaults to 5 seconds.

It attempts to avoid doing captures which only contain a snapshot.

packages/telemetry/browser-telemetry/src/collectors/rrweb/ContinuousReplay.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  unregister(): void {
+    this.stopper?.();
+    this.telemetry = undefined;
+    document.removeEventListener('visibilitychange', this.visibilityHandler);
+    clearInterval(this.timerHandle);
+  }
+
+  private handleVisibilityChange() {
```
kinyoklion, line 49: Visibility change seems inconsistent, but likely an SDK problem.

packages/telemetry/browser-telemetry/src/collectors/rrweb/RollingReplay.ts
```diff
@@ -0,0 +1,55 @@
+import * as rrweb from 'rrweb';
```
kinyoklion, line 1: This records events into a background buffer and then on demand produces a capture with the entirety of the buffer. This would be for emitting replays leaking up to an error, or when some specific situation is encountered.

cdelst (approved):

PR #492, opened 2024-06-26
chore: Scaffold universal sdk

yusinto:
As part of moonshots, I'm aiming to release an alpha version of a universal react sdk which can work with App Router.


packages/sdk/universal/package.json
```diff
@@ -0,0 +1,59 @@
+{
+  "name": "@launchdarkly/universal-sdk",
```
kinyoklion, line 2: I think we need a more constrained name for the package. `@launchdarkly/universal-react` or `@launchdarkly/react-universal` or something.

yusinto: I renamed it react-universal.

kinyoklion (approved):

PR #493, opened 2024-06-27
feat: Added react-universal code from labs.

yusinto:


.github/workflows/release-please.yml
```diff
@@ -343,8 +343,7 @@ jobs:
     permissions:
       id-token: write
       contents: write
-    # HACK: react-universal sdk is not ready for release yet.
-    if: false #${{ needs.release-please.outputs.package-react-universal-release == 'true' }}
+    if: ${{ needs.release-please.outputs.package-react-universal-release == 'true' }}
```
yusinto, line 346: I would like to publish an early alpha of this package so we can install it on external apps to get feedback and for demos please.

kinyoklion: You can do this, but the whole process is much easier if you do what I did in feat/proto-client-telemetry if we need to do this in the future. 

I have published several packages because I can just publish them from that branch using the manual publish and I don't have to worry about release configs and things because it isn't the main branch.

packages/sdk/react-universal/README.md
```diff
@@ -54,17 +52,17 @@ const nextConfig = {
 export default nextConfig;
 ```
 
-2. Create a new file `instrumentation.ts` at the root of your project. This will initialize the Node Server SDK.
+2. In `instrumentation.ts`, initialize the Node Server SDK:
```
kinyoklion, line 55: Isn't this the file usually used for configuring telemetry?

packages/sdk/react-universal/example/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
kinyoklion, line 1: Is there a reason there is another license in this file?

packages/sdk/react-universal/README.md
yusinto: Yes it is. We are using it for a different purpose, but it's the recommended solution by someone from nextjs in an issue. There's no other way to insert server startup code afaik.

packages/sdk/react-universal/example/LICENSE
yusinto: Bug. Will remove.

packages/sdk/react-universal/README.md
kinyoklion: That is unfortunate.

packages/sdk/react-universal/package.json
```diff
@@ -51,9 +60,17 @@
...
     "jest": "^29.5.0",
     "prettier": "^3.0.0",
+    "react": "^18",
     "rimraf": "^5.0.1",
     "ts-jest": "^29.1.0",
     "typedoc": "0.25.0",
     "typescript": "5.1.6"
+  },
+  "dependencies": {
+    "@launchdarkly/node-server-sdk": "workspace:^",
```
kinyoklion, line 70: This should have a version number instead of workspace.

packages/sdk/react-universal/package.json
```diff
@@ -51,9 +60,17 @@
...
+    "launchdarkly-js-client-sdk": "^3.4.0"
```
yusinto, line 71: Callout: Use the classic js sdk for now.

packages/sdk/react-universal/tsconfig.json
```diff
@@ -17,7 +17,8 @@
     "strict": true,
     "stripInternal": true,
     "target": "ES2017",
-    "types": ["jest", "node"]
+    "types": ["jest", "node", "react/canary"],
```
yusinto, line 20: Callout: Need `react/canary` to be able to use react cache. Nextjs does this so I'm not feeling too bad about it.

packages/sdk/react-universal/example/app/components/helloClientComponent.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const ldc = useLDClient();
+  const [_, setCookie] = useCookies(['ld']);
+
+  const flagValue = ldc.variation('my-boolean-flag-1');
+
+  function onClickLogin() {
+    const context = { kind: 'user', key: 'test-user-2' };
+    (ldc as JSSdk).identify(context).then(() => {
+      console.log('identify successful, persisting to cookies');
+      setCookie('ld', context);
```
yusinto, line 18: [setCookie](https://github.com/bendotcodes/cookies/tree/main/packages/react-cookie/#setcookiename-value-options) automatically stringifies the cookie value.

packages/sdk/react-universal/README.md
```diff
@@ -9,62 +9,49 @@
...
 
 - `LDProvider` - The react context provider.
 
-- `useLDClient` - Use this to get an ldClient for Client Components.
+- `useLDClient` - Gets a suitable ld client for Client Components.
 
 ## Usage
 
-1. Enable [instrumentationHook](https://nextjs.org/docs/app/building-your-application/optimizing/instrumentation) in `next.config.mjs`:
+1. On server start, initialize the Node Server SDK. If you are using NextJS App Router, do this in `instrumentation.ts`. You'll need to enable the [instrumentationHook](https://nextjs.org/docs/app/building-your-application/optimizing/instrumentation):
```
yusinto, line 44: This universal sdk should work with all framework which supports rsc (gatsby, remix, etc).

packages/sdk/react-universal/example/app/components/helloClientComponent.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { useCookies } from 'react-cookie';
+
+import type { JSSdk } from '@launchdarkly/react-universal-sdk';
+import { useLDClient } from '@launchdarkly/react-universal-sdk/client';
+
+export default function HelloClientComponent() {
+  const ldc = useLDClient();
+  const [_, setCookie] = useCookies(['ld']);
+
+  const flagValue = ldc.variation('my-boolean-flag-1');
```
kinyoklion, line 12: It seems like this would be really noisy with regards to flag evaluations? (or not noisy enough depending on the situation.)

packages/sdk/react-universal/src/ldClientRsc.ts
```diff
@@ -0,0 +1,25 @@
...
+  allFlags(): LDFlagSet {
+    return this.bootstrap;
+  }
+
+  getContext(): LDContext {
+    return this.ldContext;
+  }
+
+  variation(key: string, defaultValue?: LDFlagValue): LDFlagValue {
+    return this.bootstrap[key] ?? defaultValue;
```
kinyoklion, line 23: Will the variations always happen again client time to ensure we capture analytics?

yusinto: On the client side the js sdk will always be used so yes analytics on the client side should be ok. However on the server side, this won't send analytics. We can add a nodeClient.variation here to ensure we send analytics both server and client side.

kinyoklion: (A typical bootstrap just populates the client with data so the variation call happens.)

kinyoklion: Typically we attempt to get it once on either side. But if this it baking values, then it should call variation as you describe.

packages/sdk/react-universal/example/app/components/helloClientComponent.tsx
yusinto: Yes it will be noisy because of multiple re-renders. I think the customer should memoize at their own discretion. I'll add a comment for this. What do you think?

packages/sdk/react-universal/src/ldClientRsc.ts
yusinto: I'll add the server variation call.

packages/sdk/react-universal/example/app/components/helloClientComponent.tsx
kinyoklion: I am very skeptical of a non-hook access of a flag because of flag because it results in really unpredictable event volume. Sometimes thousands of times expected volume.

yusinto: Fair enough. This is definitely not a long term solution. The long term plan is to support `useVariation` hooks like react-native. I may not have time to fulfil this plan right now for moonshots. Perhaps a TODO and a warning suffice for now?

kinyoklion: Sounds reasonable.

yusinto:
```tsx
  // WARNING: Using the ldClient to evaluate flags directly like this in prod
  // can result in high event volumes. This example is contrived and is meant for
  // demo purposes only. The recommended way is to utilise the `useVariation` hooks
  // which will be supported soon.
  const flagValue = ldc.variation('my-boolean-flag-1');
```

packages/sdk/react-universal/src/client/LDProvider.tsx
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @param context The LDContext for evaluation.
+ * @param options Configuration options for the js sdk. See {@link LDOptions}.
+ * @param children Your react application to be rendered.
+ */
+
+export const LDProvider = ({ context, options, children }: PropsWithChildren<LDProps>) => {
+  let jsSdk: JSSdk = undefined as any;
+  if (!isServer) {
+    jsSdk = initialize(process.env.NEXT_PUBLIC_LD_CLIENT_SIDE_ID ?? '', context, options);
```
kinyoklion, line 31: Is it reasonable to require it come specifically from the ENV only?

packages/sdk/react-universal/src/ldClientRsc.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  getContext(): LDContext {
+    return this.ldContext;
+  }
+
+  variation(key: string, defaultValue?: LDFlagValue): LDFlagValue {
+    if (isServer) {
+      // On the server during ssr, call variation for analytics purposes.
+      global.nodeSdk.variation(key, this.ldContext, defaultValue).then(/* ignore */);
```
kinyoklion, line 26: Ok. We call the real SDK here, but are we configuring the server SDK such that its "bootstrap" will be the same as a variation? It doesn't look like it at a glance.

packages/sdk/react-universal/src/client/LDProvider.tsx
yusinto: I'll make it an argument.

packages/sdk/react-universal/src/ldClientRsc.ts
yusinto: For each request, the bootstrap values used for ssr should be the same throughout the server render cycle. The two hooks which construct LDClientRsc `useLDClient` and `useLDClientRsc` always use `getBootstrap` which uses the same LDContext per request so this variation call should be consistent with that.

PR #500, opened 2024-07-02
feat: Add stack trace processing.

kinyoklion:
This is prototype code targeting a prototype feature branch.

This PR adds code for processing stack traces into a uniform shape.

For productionizing the TraceKit code would need vendored and some updated made.


packages/telemetry/browser-telemetry/src/stack/StackParser.ts
```diff
@@ -0,0 +1,97 @@
+// import ErrorStackParser from 'error-stack-parser';
```
keelerm84, line 1:
```suggestion
```

keelerm84 (approved):

PR #504, opened 2024-07-09
feat: Implement error data spec and session id.

kinyoklion:
Prototype code for prototype feature branch.

### DIFF
--- a/packages/telemetry/browser-telemetry/src/BrowserTelemetryImpl.ts
+++ b/packages/telemetry/browser-telemetry/src/BrowserTelemetryImpl.ts
@@ -5,7 +5,7 @@ import type {
   LDInspection,
 } from 'launchdarkly-js-client-sdk';
 
-import { Breadcrumb } from './api/Breadcrumb';
+import { Breadcrumb, FeatureManagementBreadcrumb } from './api/Breadcrumb';
 import { BrowserTelemetry } from './api/BrowserTelemetry';
 import { Collector } from './api/Collector';
 import { Event } from './api/Event';
@@ -16,6 +16,17 @@ import { ParsedOptions } from './options';
 import parse from './stack/StackParser';
 
 // TODO: Add ring buffer instead of shifting.
+
+function safeValue(u: unknown): string | boolean | number | undefined {
+  switch (typeof u) {
+    case 'string':
+    case 'boolean':
+    case 'number':
+      return u;
+    default:
+      return undefined;
+  }
+}
 
 export default class BrowserTelemetryImpl implements BrowserTelemetry {
   private maxPendingEvents: number;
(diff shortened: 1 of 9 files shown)


packages/telemetry/browser-telemetry/src/randomUuidV4.ts
```diff
@@ -0,0 +1,98 @@
...
+}
+
+export function fallbackUuidV4(): string {
+  const bytes = getRandom128bit();
+  return formatDataAsUuidV4(bytes);
+}
+
+export default function randomUuidV4(): string {
+  if (typeof crypto !== undefined && typeof crypto.randomUUID === 'function') {
+    return crypto.randomUUID();
```
kinyoklion, line 94: Almost everyone should hit this code path.

packages/telemetry/browser-telemetry/src/randomUuidV4.ts
```diff
@@ -0,0 +1,98 @@
+// The implementation in this file generates UUIDs in v4 format and is suitable
```
kinyoklion, line 1: This implementation is potentially overkill, but it makes it clear what is happening and it can use the crypto RNG as an intermediate fallback instead of always depending on the low quality RNG when randomUUID isn't available.

kinyoklion: We will want to make this code common with the browser SDK implementation.

packages/telemetry/browser-telemetry/src/BrowserTelemetryImpl.ts
```diff
@@ -8,15 +8,21 @@ import type {
...
 import ClickCollector from './collectors/click';
 import ErrorCollector from './collectors/error';
 import makeInspectors from './inspectors';
 import { ParsedOptions } from './options';
+import randomUuidV4 from './randomUuidV4';
 import parse from './stack/StackParser';
 
 // TODO: Add ring buffer instead of shifting.
 
+const CUSTOM_KEY_PREFIX = '$ld:telemetry';
```
tanderson-ld, line 22: Is the dollar sign here intentional?

packages/telemetry/browser-telemetry/src/BrowserTelemetryImpl.ts
```diff
@@ -61,15 +68,20 @@ export default class BrowserTelemetryImpl implements BrowserTelemetry {
 
   register(client: LDClient): void {
     this.client = client;
+    if (this.pendingEvents.length) {
```
tanderson-ld, line 71: This if statement may be unnecessary.

packages/telemetry/browser-telemetry/src/randomUuidV4.ts
```diff
@@ -0,0 +1,98 @@
...
+
+function getRandom128bit(): number[] {
+  if (crypto && crypto.getRandomValues) {
+    const typedArray = new Uint8Array(16);
+    crypto.getRandomValues(typedArray);
+    return [...typedArray.values()];
+  }
+  const values = [];
+  for (let index = 0; index < 16; index += 1) {
+    values.push(Math.floor(Math.random() * 255));
```
tanderson-ld: 255 should be 256

tanderson-ld (approved):

packages/telemetry/browser-telemetry/src/randomUuidV4.ts
tanderson-ld: Resolved async

packages/telemetry/browser-telemetry/src/BrowserTelemetryImpl.ts
kinyoklion: Yes. Using a semi-unique string to prefix the custom events for the prototype.

PR #507, opened 2024-07-10
feat: Generate selector for click events.

kinyoklion:
Prototype implementation.

### DIFF
--- a/packages/telemetry/browser-telemetry/src/collectors/dom/toSelector.ts
+++ b/packages/telemetry/browser-telemetry/src/collectors/dom/toSelector.ts
@@ -1,0 +1,82 @@
+const CHILD_COMBINATOR = '>';
+const CHILD_SEPARATOR = ` ${CHILD_COMBINATOR} `;
+
+interface NodeWithParent {
+  parentNode?: NodeWithParent;
+}
+
+interface BasicElement {
+  tagName?: string;
+  id?: string;
+  className?: string;
+  // getAttribute(key: string): string;
+}
+
+function isNode(element: unknown): element is NodeWithParent {
+  const anyElement = element as any;
+  return typeof anyElement === 'object' && anyElement.parentNode !== undefined;
+}
+
+function isElement(element: unknown): element is BasicElement {
+  const anyElement = element as any;
+  return typeof anyElement === 'object' && typeof anyElement.getAttribute === 'function';
+}
+
+function getClassName(element: BasicElement): string | undefined {
+  if (typeof element.className !== `string`) {
+    return undefined;
+  }
+  let value = element.className;
+  // Elements should be space separated in a class attribute. If there are other kinds of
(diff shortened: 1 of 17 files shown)


packages/telemetry/browser-telemetry/src/collectors/dom/toSelector.ts
```diff
@@ -0,0 +1,138 @@
...
+  let ptr = element;
+  while (isNode(ptr) && ptr.parentNode) {
+    const asString = elementToString(ptr as BasicElement);
+    // We do not need to include the 'html' component in the selector.
+    if (asString === 'html') {
+      break;
+    }
+    if (components.length > options.maxDepth) {
+      break;
+    }
```
tanderson-ld, line 132: I think this condition could be moved to be in the while condition and would avoid an extra `elementToString` on the last iteration.

packages/telemetry/browser-telemetry/src/collectors/dom/toSelector.ts
```diff
@@ -0,0 +1,138 @@
...
+  // Limiting the components should generate reasonable selectors in most cases.
+  const components: string[] = [];
+  let ptr = element;
+  while (isNode(ptr) && ptr.parentNode) {
+    const asString = elementToString(ptr as BasicElement);
+    // We do not need to include the 'html' component in the selector.
+    if (asString === 'html') {
+      break;
+    }
+    if (components.length > options.maxDepth) {
```
tanderson-ld, line 130: If length is max depth, another push will happen.  Is that expected?

I guess it depends on if maxDepth includes the top element.  If that is the case, it is a little unexpected.  I would expect whatever option I provide to be the number of components in the returned string.

packages/telemetry/browser-telemetry/src/collectors/dom/toSelector.ts
```diff
@@ -0,0 +1,138 @@
...
+): string {
+  // For production we may want to consider if we additionally limit the maximum selector length.
+  // Limiting the components should generate reasonable selectors in most cases.
+  const components: string[] = [];
+  let ptr = element;
+  while (isNode(ptr) && ptr.parentNode) {
+    const asString = elementToString(ptr as BasicElement);
+    // We do not need to include the 'html' component in the selector.
+    if (asString === 'html') {
+      break;
```
tanderson-ld, line 128: Should this be a continue?

tanderson-ld (changes requested):

packages/telemetry/browser-telemetry/src/collectors/dom/toSelector.ts
kinyoklion: No. We want to stop if we get to HTML and not include it. Basically HTML is redundant information at the top.

tanderson-ld (approved):

PR #508, opened 2024-07-10
feat: Add HTTP breadcrumbs.

kinyoklion:
Prototype implementation.


packages/telemetry/browser-telemetry/src/collectors/error.ts
```diff
@@ -12,6 +12,15 @@ export default class ErrorCollector implements Collector {
       },
       true,
     );
+    window.addEventListener(
```
kinyoklion, line 15: I was not handling unhandled promise rejections. Treating it the same as an error when possible.

packages/telemetry/browser-telemetry/__tests__/collectors/dom/toSelector.test.ts
```diff
@@ -0,0 +1,84 @@
+import toSelector, { elementToString, getClassName } from '../../../src/collectors/dom/toSelector';
```
kinyoklion, line 1: Moved this file, but it should not have changes.

packages/telemetry/browser-telemetry/src/collectors/http/fetchDecorator.ts
```diff
@@ -0,0 +1,95 @@
...
+  return { url, method };
+}
+
+/**
+ * Decorate fetch and execute the callback whenever a fetch is completed providing a breadcrumb.
+ *
+ * @param callback Function which handles a breadcrumb.
+ */
+export default function decorateFetch(callback: (breadcrumb: HttpBreadcrumb) => void) {
+  // TODO: Check if already wrapped?
```
kinyoklion, line 48: There are a number of TODO items in this PR that will remain.

packages/telemetry/browser-telemetry/src/api/Options.ts
```diff
@@ -1,5 +1,49 @@
...
+   *
+   * Defaults to true.
+   */
+  instrumentXhr?: boolean;
+
+  /**
+   * Customize URL filtering. This will be applied in addition to some baseline filtering included
+   * which redacts components of LaunchDarkly URLs.
+   */
+  customUrlFilter?: UrlFilter;
```
kinyoklion, line 44: We include a base filter, and then a customer an supply this additional filter. We may need more advanced filtering. For instance excluding entire requests, for a production version.

packages/telemetry/browser-telemetry/src/collectors/http/xhrDecorator.ts
```diff
@@ -0,0 +1,123 @@
...
+    }
+  }
+
+  window.XMLHttpRequest.prototype.open = wrappedOpen;
+  window.XMLHttpRequest.prototype.send = wrappedSend;
+
+  try {
+    // Use defineProperties to prevent these values from being enumerable.
+    // The properties default to non-enumerable.
+    Object.defineProperties(window.XMLHttpRequest, {
```
kinyoklion, line 107: Not doing anything with these yet, but will probably check for their presence in a future version. For cases where people instantiate multiple SDKs and telemetry.

packages/telemetry/browser-telemetry/src/filters/defaultUrlFilter.ts
```diff
@@ -0,0 +1,29 @@
+const pollingRegex = /sdk\/evalx\/\w+\/contexts\/(?<context>[a-zA-Z0-9=]*)\??.*?/;
+const streamingREgex = /\/eval\/[^/]*\/(?<context>[a-zA-Z0-9=]*)\??.*?/;
+
+/**
+ * Filter which removes context information for browser JavaScrip endpoints.
```
kinyoklion, line 5:
```suggestion
 * Filter which removes context information for browser JavaScript endpoints.
```

packages/telemetry/browser-telemetry/src/filters/defaultUrlFilter.ts
```diff
@@ -0,0 +1,29 @@
+const pollingRegex = /sdk\/evalx\/\w+\/contexts\/(?<context>[a-zA-Z0-9=]*)\??.*?/;
+const streamingREgex = /\/eval\/[^/]*\/(?<context>[a-zA-Z0-9=]*)\??.*?/;
+
+/**
+ * Filter which removes context information for browser JavaScript endpoints.
```
kinyoklion, line 5: This is a sufficient implementation for a prototype. It isn't format agnostic, which means it could break easily. 

We may want to filter to something more generic for production. For instance if the header includes one of the LD specified headers, then replace the URL with something like `<LaunchDarkly-Polling>` or other generic identifiers.

packages/telemetry/browser-telemetry/src/BrowserTelemetryImpl.ts
```diff
@@ -85,7 +110,7 @@ export default class BrowserTelemetryImpl implements BrowserTelemetry {
         this.pendingEvents.shift();
       }
     }
-    this.client?.track(`$ld:telemetry:${type}`, event);
```
kinyoklion, line 88: Bug from earlier refactoring.

packages/telemetry/browser-telemetry/src/filters/defaultUrlFilter.ts
tanderson-ld: Could just lopping off everything after the domain be sufficient for our customers?  Then the filter could look for our domains, find end, truncate.

tanderson-ld: Discussed more in another channel.

tanderson-ld (approved):

packages/telemetry/browser-telemetry/src/collectors/http/fetch.ts
```diff
@@ -0,0 +1,27 @@
...
+      this.destination?.addBreadcrumb(breadcrumb);
+    });
+  }
+
+  register(telemetry: BrowserTelemetry): void {
+    this.destination = telemetry;
+  }
+
+  unregister(): void {
+    this.destination = undefined;
```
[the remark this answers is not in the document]
kinyoklion: A field before registration is undefined and after unregistration is undefined. 
`private destination?: BrowserTelemetry;`
is the same as 
`private destination: BrowserTelemetry | undefined`

As the field is not defined in the constructor it will be undefined, so it is conditionally accessed.

It isn't assigned in the constructor because I want to be able to make collectors immediately and then attach them.

The second thing is just because it was defined that way in the lib.dom.ts and I copied the definition.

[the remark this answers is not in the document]
kinyoklion: Generally speaking optional elements cannot follow non-optional elements, so there is another case where it is a parameter is non-optional, but can be undefined. Which just means you have to pass something.

PR #518, opened 2024-07-17
fix: Make it more clear what is happening when an event source is connecting.

kinyoklion:
The event source was treating the initial connection as a re-connection. This results in confusing messaging.

This also changes the retrying event, which we do not expose.

When we convert our polyfill to typescript we should use it instead and allow passing a connection method. The reason this is being used is theoretically an issue with fetch not streaming in RN, so it uses an XHR instead.

### DIFF
--- a/packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.test.ts
+++ b/packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.test.ts
@@ -6,6 +6,7 @@ import EventSource, { backoff, jitter } from './EventSource';
 describe('EventSource', () => {
   const uri = 'https://mock.events.uri';
   let eventSource: EventSource<EventName>;
+  let mockXhr: any;
 
   beforeAll(() => {
     jest.useFakeTimers();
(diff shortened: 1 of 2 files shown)


packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.test.ts
```diff
@@ -17,9 +18,17 @@ describe('EventSource', () => {
       .mockImplementationOnce(() => 0.888)
       .mockImplementationOnce(() => 0.999);
 
+      mockXhr = {
```
kinyoklion, line 21: I made a mock XHR so it makes the request to this instead of a general request.
I did this because I made the event source open method private, because it should be private.

We should make these tests respond to the XHR responses instead of directly manipulating the event source, but that was out of scope for the problem I wanted to resolve.

packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.ts
```diff
@@ -57,7 +57,7 @@ export default class EventSource<E extends string = never> {
   private body: any;
   private url: string;
   private xhr: XMLHttpRequest = new XMLHttpRequest();
-  private pollTimer: any;
+  private connectTimer: any;
```
kinyoklion, line 60: Changed the name to reflect what it is doing.

packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.ts
```diff
@@ -88,17 +88,25 @@ export default class EventSource<E extends string = never> {
...
+    let delay = initialConnection ? 0 : this.getNextRetryDelay();
+    if(initialConnection) {
+      this.logger?.debug(`[EventSource] opening new connection.`)
+    } else {
+      this.logger?.debug(`[EventSource] Will open new connection in ${delay} ms.`);
+      this.dispatch('retry', { type: 'retry', delayMillis: delay });
+    }
+
+    this.connectTimer = setTimeout(() => {
+      if(!initialConnection) {
```
kinyoklion, line 101: Nothing to close for the first connection.

louis-launchdarkly (approved):

PR #522, opened 2024-07-29
feat: Add connection mananger.

kinyoklion:
This adds a connection manager which is used to handle application state and network state and determine the desired connection mode.

### DIFF
--- a/packages/sdk/react-native/src/platform/ConnectionManager.ts
+++ b/packages/sdk/react-native/src/platform/ConnectionManager.ts
@@ -97,11 +97,11 @@ export class ConnectionManager {
     this.logger.debug(`Handling state: ${this.applicationState}:${this.networkState}`);
 
     switch (this.networkState) {
-      case NetworkState.Available:
-        this.destination.setNetworkAvailability(false);
-        break;
       case NetworkState.Unavailable:
         this.destination.setNetworkAvailability(false);
+        break;
+      case NetworkState.Available:
+        this.destination.setNetworkAvailability(true);
         switch (this.applicationState) {
           case ApplicationState.Foreground:
             this.setForegroundAvailable();
--- a/packages/sdk/react-native/src/platform/ConnectionManager.ts
+++ b/packages/sdk/react-native/src/platform/ConnectionManager.ts
@@ -2,22 +2,22 @@ import { ConnectionMode, LDLogger } from '@launchdarkly/js-client-sdk-common';
 
 export enum ApplicationState {
   /// The application is in the foreground.
-  Foreground,
+  Foreground = 'foreground',
 
   /// The application is in the background.
   ///
   /// Note, the application will not be active while in the background, but
   /// it will track when it is entering or exiting a background state.
-  Background,
+  Background = 'background',
 }
 
 export enum NetworkState {
   /// There is no network available for the SDK to use.
-  Unavailable,
+  Unavailable = 'unavailable',
 
   /// The network is available. Note that network requests may still fail
   /// for other reasons.
-  Available,
+  Available = 'available',
 }
 
 export interface ConnectionDestination {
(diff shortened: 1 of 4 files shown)


packages/shared/sdk-client/src/api/ConnectionMode.ts
```diff
@@ -9,6 +9,6 @@
  *
  * streaming - The SDK will use a streaming connection to receive updates from LaunchDarkly.
  */
-type ConnectionMode = 'offline' | 'streaming';
```
kinyoklion, line 12: Adding polling as a connection mode. An implementation of polling will be in a later PR.

packages/sdk/react-native/src/platform/ConnectionManager.ts
```diff
@@ -0,0 +1,149 @@
+import { ConnectionMode, LDLogger } from '@launchdarkly/js-client-sdk-common';
+
+export enum ApplicationState {
```
kinyoklion, line 3: These "state" enums are internal to the SDK, so I used enums for them. (Versus a union type.)

packages/sdk/react-native/src/platform/ConnectionManager.ts
```diff
@@ -0,0 +1,149 @@
...
+export enum NetworkState {
+  /// There is no network available for the SDK to use.
+  Unavailable,
+
+  /// The network is available. Note that network requests may still fail
+  /// for other reasons.
+  Available,
+}
+
+export interface ConnectionDestination {
```
kinyoklion, line 23: Eventually the common client will likely support this interface, but having the abstraction here is nice for testing.

The common interface for the client will likely not expose these as they are for internal control.

packages/sdk/react-native/src/platform/ConnectionManager.ts
```diff
@@ -0,0 +1,149 @@
...
+  setNetworkAvailability(available: boolean): void;
+  setEventSendingEnabled(enabled: boolean, flush: boolean): void;
+  setConnectionMode(mode: ConnectionMode): Promise<void>;
+  flush(): Promise<void>;
+}
+
+export interface StateDetector {
+  setApplicationStateListener(fn: (state: ApplicationState) => void): void;
```
kinyoklion, line 31: I used basic callback functions for "events" from the state detector.

packages/sdk/react-native/src/platform/ConnectionManager.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  /// The network is available. Note that network requests may still fail
+  /// for other reasons.
+  Available,
+}
+
+export interface ConnectionDestination {
+  setNetworkAvailability(available: boolean): void;
+  setEventSendingEnabled(enabled: boolean, flush: boolean): void;
+  setConnectionMode(mode: ConnectionMode): Promise<void>;
```
kinyoklion, line 26: The current version of setting the connection mode is asynchronous. I am not sure if that is ideal. The process itself is asynchronous, but having a promise that indicates it is "finished" seems risky. Especially lots of independent inputs and without a queue managing transitions.

In the LDClient interface is returns void, but the implementation actually returns a promise and we don't provide any specification of behavior.

packages/sdk/react-native/src/platform/ConnectionManager.ts
```diff
@@ -0,0 +1,149 @@
...
+  /// connection state will automatically change when network
+  /// availability changes.
+  readonly automaticNetworkHandling: boolean;
+
+  /// Enable handling associated with transitioning between the foreground
+  /// and background.
+  readonly automaticBackgroundHandling: boolean;
+}
+
+export class ConnectionManager {
```
kinyoklion, line 55: This is largely a port of the flutter implementation.

tanderson-ld (approved):

PR #523, opened 2024-07-29
feat: Refactor application state handling.

kinyoklion:


packages/sdk/react-native/src/provider/useAppState.ts
```diff
@@ -1,58 +0,0 @@
-import { useEffect, useRef } from 'react';
```
kinyoklion, line 1: This file contained this big problem that needs resolved. The event source, which isn't the only update processor type possible, was being accessed via the platform.

packages/sdk/react-native/src/provider/LDProvider.tsx
```diff
@@ -26,8 +25,6 @@ const LDProvider = ({ client, children }: PropsWithChildren<LDProps>) => {
     setupListeners(client, setState);
   }, []);
 
-  useAppState(client);
```
kinyoklion, line 29: Another problem with this is that the overall desired connection mode would fight with the connection mode determined by the "app state".

If you tell the client to be offline, then you don't expect a foreground transition to set it back online, for example.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -538,4 +543,42 @@ export default class LDClientImpl implements LDClient {
...
+      this.eventProcessor?.start();
+    } else if (flush) {
+      this.flush().then(() => {
+        // While waiting for the flush event sending could be re-enabled, in which case
+        // we do not want to close the event processor.
+        if (!this.eventSendingEnabled) {
+          this.logger?.debug('Stopping event processor.');
+          this.eventProcessor?.close();
+        }
+      });
```
tanderson-ld, line 581: Do you need code in the !enabled and !flush case?  Such as

```
        if (!this.eventSendingEnabled) {
          this.logger?.debug('Stopping event processor.');
          this.eventProcessor?.close();
        }
```

tanderson-ld (approved):

PR #524, opened 2024-07-30
feat: Implement polling support.

kinyoklion:


packages/shared/sdk-client/src/LDClientImpl.storage.test.ts
```diff
@@ -37,7 +37,7 @@ const onChangePromise = () =>
   });
 
 // Common setup code for all tests
-// 1. Sets up streamer
```
kinyoklion, line 40: Replaced all streamer with either `updateProcessor` or `streaming`.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -95,31 +99,29 @@ export default class LDClientImpl implements LDClient {
...
         break;
       case 'polling':
-        this.logger.warn('Polling not supported. Using streaming.');
-      // Intentionally falling through to streaming.
-      // eslint-disable-next-line no-fallthrough
       case 'streaming':
         if (this.context) {
-          // identify will start streamer
+          // identify will start the update processor
           return this.identify(this.context, { timeout: this.identifyTimeout });
```
kinyoklion, line 118: I think this isn't the correct way to do this long-term. We don't need the identify events, for instance, and are probably doing more work than required.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -339,28 +350,76 @@ export default class LDClientImpl implements LDClient {
...
+    context: any,
+    checkedContext: Context,
+    identifyReject: any,
+  ) {
+    let pollingPath = this.createPollUriPath(context);
+    if (this.config.withReasons) {
+      pollingPath = `${pollingPath}?withReasons=true`;
+    }
+    this.updateProcessor = new PollingProcessor(
+      this.sdkKey,
```
kinyoklion, line 382: Streaming takes a "ClientContext", but providing the individual items allows better encapsulation and also mocking without needing to mock as many components.

packages/shared/sdk-client/src/configuration/Configuration.ts
```diff
@@ -41,7 +43,6 @@ export default class Configuration {
   public readonly privateAttributes: string[] = [];
 
   public readonly initialConnectionMode: ConnectionMode = 'streaming';
-  public connectionMode: ConnectionMode;
```
kinyoklion, line 44: Runtime state is part of the client now, not part of the configuration.

packages/shared/sdk-client/src/polling/Requestor.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  constructor(message: string, status?: number) {
+    super(message);
+    this.status = status;
+    this.name = 'LaunchDarklyRequestError';
+  }
+}
+
+/**
+ * Note: The requestor is implemented independently from polling such that it can be used to
```
kinyoklion, line 25: In the browser version of JS we should maintain something close to the existing behavior, and we can use this to do that. Current js-client-sdk will make a single request first. If someone listens to changes, or specifically requests streaming, then it will stream.

We may make the requestor in the client, but for now I am putting it in the polling implementation.

tanderson-ld (approved):

PR #530, opened 2024-07-31
fix: Fix field visibility.

kinyoklion:
This fixes the field visibility on a number of what should have been private or readonly fields.

This is technically a breaking change because the implementation types were exposed, but they are not part of the interface.

One attribute that is in the interface is the logger, and it has been marked as readonly now, which is also a breaking change.

Setting or manipulating any of these fields would have resulted in potentially inconsistent internal state.

### DIFF
--- a/packages/shared/sdk-client/src/LDClientImpl.test.ts
+++ b/packages/shared/sdk-client/src/LDClientImpl.test.ts
@@ -65,40 +65,6 @@ describe('sdk-client object', () => {
     jest.resetAllMocks();
   });
 
-  test('instantiate with blank options', () => {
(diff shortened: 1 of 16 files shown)


tanderson-ld (approved):
Approving even though this is in draft in case you want to merge today.

packages/shared/sdk-client/src/LDClientImpl.test.ts
```diff
@@ -65,40 +65,6 @@ describe('sdk-client object', () => {
     jest.resetAllMocks();
   });
 
-  test('instantiate with blank options', () => {
```
kinyoklion, line 68: This test didn't seem to add value over config tests.

tanderson-ld (approved):

PR #531, opened 2024-08-01
feat: refactors the implementation of context caching.  You can now s…

tanderson-ld:
…pecify the maxCachedContexts (default 5) in the LDOptions for client-side SDKs.

**Requirements**

TODOs:
- [x] FlagManager and FlagPersistence tests
- [x] need to do some validation on platforms and manual examination of storage
- [x] implement data migration instructions in FlagPersistence.ts
- [x] need to validate migration of existing data
- [x] [update various namespace concat usages to use static helper funcs](https://github.com/launchdarkly/js-core/pull/531/files#diff-2ea53243d0cda26506ba1e932279540f87359ef3647357de41dd6961553b3908R6)
- [x] [discuss immutability of return values](https://github.com/launchdarkly/js-core/pull/531/files#diff-d292bcc848729d6a937835ad9b3aefb2217a4ab9057a1b9caa22eb32b6412cbdR22)
- [x] [add tests for namespace hashing and utf8 encoding](https://github.com/launchdarkly/js-core/pull/531/files#diff-2ea53243d0cda26506ba1e932279540f87359ef3647357de41dd6961553b3908R23)
- [x] [verify deleteFlag handling](https://github.com/launchdarkly/js-core/pull/531/files#diff-5ab8d43c9c238bf0ca76fae92d46229147658df4640f421264a1f57eebfdde79R215)

**Related issues**

SC-247373

Introduces FlagManager, FlagUpdater, FlagPersistence, and flag CacheIndex for creating a LRU for cached flag evaluations.  This follows the implementations of other client SDKs.


packages/shared/common/src/Context.ts
```diff
@@ -338,6 +338,38 @@ export default class Context {
...
   }
 
+  /**
+   * Creates a {@link LDContext} from a {@link Context}.
+   * @param context to be converted
+   * @returns an {@link LDContext}
+   */
+  public static toLDContext(context: Context): LDContext {
+    if (!context.valid) {
+      return <LDContext>{kind: context.kind, message: context.message}
```
kinyoklion, line 348: This doesn't comply with the typing of a context. I know the original context likely didn't either, but we would need to do something different here.

Like the typing could be a union. 
`LDContext | LDContextError`

packages/shared/common/__tests__/Context.test.ts
```diff
@@ -45,7 +45,7 @@ describe.each([
     name: 'context name',
     cat: 'calico',
     anonymous: true,
-    _meta: { privateAttributes: ['/~1dog~0~0~1~1'] },
```
kinyoklion, line 48: So, this specific redaction verifies that all the escape codes are working in context redaction. This change breaks two tests:
image

They aren't running in CI because of the typescript error causing it to fail our earlier.

packages/shared/sdk-client/src/api/LDOptions.ts
```diff
@@ -139,6 +139,14 @@ export interface LDOptions {
    */
   logger?: LDLogger;
 
+  /**
+   * The maximum number of locally cached contexts.  Local caches hold data for use in situations
+   * where the SDK cannot reach the LaunchDarkly services.
```
kinyoklion, line 144: Maybe something like:
```suggestion
   * The maximum number of locally cached contexts.  The cache is used to decrease initialization
   * latency and to provide fallback when the SDK cannot reach LaunchDarkly services.
```
But I need to review the rest of the PR before I know if that is true.

packages/shared/sdk-client/src/flag-manager/FlaggManager.ts
kinyoklion, line 1: `FlaggManager` too many `g`.

packages/shared/common/src/Context.ts
```diff
@@ -338,6 +338,38 @@ export default class Context {
...
+        }}
+      }
+    } 
+      const result : LDMultiKindContext = {
+        kind: 'multi'
+      }
+      result.kind = 'multi'
+      context.getContexts().forEach(kindAndContext => {
+        const kind = kindAndContext[0]
+        const innerContect = kindAndContext[1]
```
kinyoklion, line 366:
```suggestion
        const innerContext = kindAndContext[1]
```

packages/shared/sdk-client/src/flag-manager/FlagPersistence.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+  }
+
+  private async loadIndex(): Promise<ContextIndex> {
+    if (this.contextIndex !== undefined) {
+      return this.contextIndex;
+    }
+
+    const json = await this.platform.storage?.get(this.indexKey);
+    if (json === null || json === undefined) {
```
kinyoklion, line 89: I think this could be a falsy check? If an empty string is also not a valid context index.

packages/shared/sdk-client/src/flag-manager/FlagPersistence.ts
kinyoklion, line 1: Do we need any exception handling around the platform storage, or do we require the handling be inside the platform storage? (I see that the RN storage implementation is doing that. We probably want to document that in the storage interface. "Implementation may not throw exceptions.")

packages/shared/sdk-client/src/utils/namespaceUtils.ts
kinyoklion, line 1: I think these should probably have a home other than utils. 

For instance, I would put all the context key stuff in a `context` folder, http stuff in a `network` folder and probably namespace stuff in a `storage` folder. Though here I am only worried about the new things which would be namespace.

(For me `utils` is basically either a banned word or one that requires a special conditions. Like we need to turn two keys simultaneously to put things into utils.)

packages/shared/sdk-client/src/utils/getOrGenerateKey.ts
```diff
@@ -1,38 +1,12 @@
 import { Platform } from '@launchdarkly/js-sdk-common';
 
-export type Namespace = 'anonymous' | 'context';
+export const getOrGenerateKey = async (storageKey: string, { crypto, storage }: Platform) => {
```
kinyoklion, line 3: I think we need docs on this. Specifically we should cover key requirements. If someone wants to use this for something new, they should know to not just use some random string, but formulate their string using other utilities.

packages/shared/sdk-client/src/utils/ensureKey.ts
```diff
@@ -62,7 +64,7 @@ const ensureKeyLegacy = async (c: LDUser, platform: Platform) => {
  * @param platform
  */
 const ensureKey = async (context: LDContext, platform: Platform) => {
-  const cloned = clone<LDContext>(context);
+  const cloned: LDContext = clone<LDContext>(context);
```
kinyoklion, line 67: Looks like the typing on the clone function is wrong. Maybe we could fix that instead. (Generally speaking every function should have a return type annotation. Maybe we are missing a lint rule.)

packages/shared/sdk-client/src/utils/calculateChangedKeys.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  // flag deleted or updated
+  Object.entries(existingObject).forEach(([k, f]) => {
+    const subObject = newObject[k];
+    if (!subObject || !fastDeepEqual(f, subObject)) {
+      changedKeys.push(k);
+    }
+  });
+
+  // flag added
```
kinyoklion, line 18: Is this isn't flag specific now, then flag goes away. (Though I am not sure this should be generalized. The implementation happens to be non-specific, but the requirements aren't that it must be non-specific.)

packages/shared/sdk-client/src/utils/calculateChangedKeys.ts
```diff
@@ -0,0 +1,26 @@
+import { fastDeepEqual } from '@launchdarkly/js-sdk-common';
+
+// eslint-disable-next-line import/prefer-default-export
```
kinyoklion, line 3: This is a default export, so what is actually happening here? Is this vestigial?

packages/shared/sdk-client/src/flag-manager/ItemDescriptor.ts
```diff
@@ -0,0 +1,6 @@
+import { Flag } from '../types';
+
+export interface ItemDescriptor {
```
kinyoklion, line 3: It would be nice to have docs on this.

packages/shared/sdk-client/src/flag-manager/FlagUpdater.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    const currentValue = this.flagStore.get(key);
+    if (currentValue !== undefined && currentValue.version >= item.version) {
+      // this is an out of order update that can be ignored
+      return false;
+    }
+
+    this.flagStore.insertOrUpdate(key, item);
+    this.changeCallbacks.forEach((callback) => {
+      callback(context, [key]);
```
kinyoklion, line 52: Do we need to safely invoke there here, or are they wrapped someplace further up the chain?

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -26,16 +26,19 @@ import Configuration from './configuration';
...
-import { addAutoEnv, calculateFlagChanges, ensureKey } from './utils';
+import { addAutoEnv, ensureKey } from './utils';
 
 const { createErrorEvaluationDetail, createSuccessEvaluationDetail, ClientMessages, ErrorKinds } =
   internal;
 
 export default class LDClientImpl implements LDClient {
   private readonly config: Configuration;
-  private context?: LDContext;
+  private inputContext?: LDContext;
```
kinyoklion, line 40: I don't know that we should call this 'input' context. It makes it seem like it would be close to what was 'input' but actually has been modified by the client.

packages/shared/sdk-client/src/flag-manager/ItemDescriptor.ts
tanderson-ld: Planning to do a docs pass once first pass of comments / changes are addressed.

packages/shared/sdk-client/src/flag-manager/FlagUpdater.ts
tanderson-ld: Will add protection, good call.

tanderson-ld: Added protection at the call site to make the tracking of the callback instances less complex.

packages/shared/sdk-client/src/LDClientImpl.ts
tanderson-ld: Maybe that begs the question, why is auto env attributes operating on the input type?

packages/shared/common/src/Context.ts
tanderson-ld: Bugger, I meant to come back and fix this before updating to under review.

packages/shared/sdk-client/src/LDClientImpl.ts
kinyoklion: I do think that getContext should give you the context that has everything added to it.

If the context was verbatim the context the user provided, then we maybe shouldn't even have getContext().

packages/shared/common/src/Context.ts
tanderson-ld: Done.  Now returns undefined.

packages/shared/sdk-client/src/flag-manager/FlagPersistence.ts
tanderson-ld: Added documentation for not throwing exceptions.

packages/shared/sdk-client/src/utils/calculateChangedKeys.ts
tanderson-ld: Yeah this wasn't relevant.  Removed.

packages/shared/common/src/api/platform/Storage.ts
```diff
@@ -1,3 +1,25 @@
...
+ * is then encoded.
+ *
+ * Implementations may not throw exceptions.
+ *
+ * The SDK assumes that the persistence is only being used by a single instance
+ * of the SDK per SDK key (two different SDK instances, with 2 different SDK
+ * keys could use the same persistence instance).
+ *
+ * The SDK, with correct usage, will not have overlapping writes to the same
+ * key. The Read/Write methods may not always be called from the same thread.
```
kinyoklion, line 17:
```suggestion
 * key.
```
No threads here.

kinyoklion (approved):

packages/shared/sdk-client/src/utils/ensureKey.ts
tanderson-ld: Reverted this change as it ultimately ended up being out of scope of this PR.

packages/shared/sdk-client/src/LDClientImpl.ts
tanderson-ld: Renamed to uncheckedContext and added explanation in getContext() for the rationale.

PR #532, opened 2024-08-01
feat: Add configuration validation for ReactNative specific configuration.

kinyoklion:


packages/sdk/react-native/src/ReactNativeLDClient.ts
```diff
@@ -59,7 +61,7 @@ export default class ReactNativeLDClient extends LDClientImpl {
       sdkKey,
       autoEnvAttributes,
       createPlatform(logger),
-      { ...options, logger },
+      { ...filterToBaseOptions(options), logger },
```
kinyoklion, line 64: If we were to pass all options to the base, then we would be warned about unrecognized ones.

An alternate approach here would be to pass the validation information to the base, but it would be a pretty substantial change and not provide as solid typing.

tanderson-ld (approved):

PR #533, opened 2024-08-01
fix: Handle non-status code errors for streaming connection.

kinyoklion:
Also the buttons had gotten too tall, so I re-arranged them.
image


packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.ts
```diff
@@ -165,8 +165,9 @@ export default class EventSource<E extends string = never> {
             this.logger?.debug('[EventSource][onreadystatechange][DONE] Operation done.');
             this.tryConnect();
           }
-        } else if (this.xhr.status !== 0) {
+        } else {
```
kinyoklion, line 167: So far I have tested this on android. Need to validate iOS and others.

tanderson-ld (approved):

PR #537, opened 2024-08-12
chore: adding max contexts validator

tanderson-ld:
**Additional context**

Forgot to add the max cached context to the validator and realized when bench testing.

### DIFF
--- a/packages/shared/sdk-client/src/configuration/Configuration.test.ts
+++ b/packages/shared/sdk-client/src/configuration/Configuration.test.ts
@@ -89,4 +89,12 @@ describe('Configuration', () => {
       expect.stringMatching(/should be of type LDFlagSet, got string/i),
     );
   });
+
+  test('recognize max cached contexts', () => {
+    // @ts-ignore
+    const config = new Configuration({ maxCachedContexts: 3 });
+
+    expect(config.maxCachedContexts).toBeDefined();
+    expect(console.error).not.toHaveBeenCalled();
+  });
 });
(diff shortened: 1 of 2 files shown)


packages/shared/sdk-client/src/configuration/Configuration.test.ts
```diff
@@ -89,4 +89,12 @@ describe('Configuration', () => {
       expect.stringMatching(/should be of type LDFlagSet, got string/i),
     );
   });
+
+  test('recognize max cached contexts', () => {
+    // @ts-ignore
```
kinyoklion, line 94: What is being ignored here?

packages/shared/sdk-client/src/configuration/Configuration.test.ts
kinyoklion, line 1: Should update the default config test.

packages/shared/sdk-client/src/configuration/validators.ts
```diff
@@ -27,6 +27,7 @@ class ConnectionModeValidator implements TypeValidator {
 const validators: Record<keyof LDOptions, TypeValidator> = {
   initialConnectionMode: new ConnectionModeValidator(),
   logger: TypeValidators.Object,
+  maxCachedContexts: TypeValidators.Number,
```
kinyoklion, line 30: Should this be ranged?

kinyoklion (changes requested):

packages/shared/sdk-client/src/configuration/validators.ts
tanderson-ld: I thought about that for a moment when I was making the changes.  The code internally clamps it to [0, infinity), so it won't hurt to have a negative.  But now thinking on it more, the validator error would probably be more helpful to the customer.  Will update.

kinyoklion: Yeah, we do have the `numberWithMin` validator. Like used on polling and a few other things.

kinyoklion (approved):

Issue #436, opened 2024-04-12
Provide an option to set custom storage


artyorsh:
**Is your feature request related to a problem? Please describe.**

In react-native, the SDK uses `@react-native-async-storage/async-storage` package without an option to specify a custom destination for SDK keys. This mixes app-related keys with SDK keys, which is especially painful when migrating to a different library (say, [MMKV](https://github.com/mrousavy/react-native-mmkv/blob/master/docs/MIGRATE_FROM_ASYNC_STORAGE.md#migrate-from-asyncstorage)). This also blocks apps from removing async-storage dependency.

**Describe the solution you'd like**
```ts
import { AutoEnvAttributes, ReactNativeLDClient, LDStorage } from '@launchdarkly/react-native-client-sdk';

const customStorage: LDStorage = {
  getItem: (k) => { mmkv.get(k) },
  setItem: (k, v) => { mmkv.set(k, v) },
};

const client = new ReactNativeLDClient(
  'mobile-key',
  AutoEnvAttributes.Enabled,
  { storage: customStorage },
);
```

**Describe alternatives you've considered**

Providing a documentation on what keys are stored by SDK would also work, giving an option to filter async-storage.

yusinto:
We will investigate this. Internally logged 240413.

angelo-hub:
+1 on this, we try to use mmkv for everything

PR #539, opened 2024-08-13
feat: custom storage option for React Native SDK

oblador:
This PR fixes https://github.com/launchdarkly/js-core/issues/436 by implementing a custom storage option to the React Native package. 

The main motivation is to get rid of the obsolete async storage package, but we also found that having multiple clients as [recommended by the official docs](https://docs.launchdarkly.com/sdk/features/multiple-environments#react-native) when migrating away from `secondaryMobileKeys` in v9 caused them to overwrite each other's storage and therefore effectively disabling the storage/cache altogether.


kinyoklion:
Hello @oblador,

Thank you for the contribution. I agree that we should allow custom storage and we will look into your PR.

Part of the issue you were encountering was a bug and we have released a new version which addresses it. Each environment should be stored in its own namespace and not overlap.

Thank you,
Ryan

kinyoklion:
Also is there a storage package which you consider idiomatic for the react native ecosystem?

My current understanding is that built-in async storage has been removed, but the @react-native-async-storage/async-storage package is a compatible and maintained replacement.

Thank you,
Ryan

packages/sdk/react-native/src/RNOptions.ts
```diff
@@ -25,6 +25,13 @@ export interface RNSpecificOptions {
    * Defaults to true.
    */
   readonly automaticBackgroundHandling?: boolean;
+
+  /**
+   * Custom persistence layer for offline mode.
```
kinyoklion, line 30: Storage is used in all modes it is loaded before the networked data source to decrease latency and prevent the usage of default values.

packages/sdk/react-native/src/RNOptions.ts
kinyoklion, line 1: Additionally the options.ts file validates options.

The validation is only cursory and many examples are available in the common client `Configuration.ts`.

In this instance we should validate that storage is an object.

The other thing we need to do is remove it from the options passed to the base configuration. Otherwise it will log that it has received an unknown configuration.

oblador:
AsyncStorage has wide adoption mostly due to legacy reasons (previously being part of core). Nowadays I'd argue that most new projects will pick MMKV, but I would personally prefer to decide this myself over being forced by the library.

kinyoklion:
Hello @oblador,

I've made some updates to the PR. Please let me know if these work for you.

I am still keeping the async storage base dependency for now.

Thanks,
Ryan

oblador:
Looks good, thanks!

kinyoklion (approved):

PR #544, opened 2024-08-14
build(deps): Update react-native example to newer expo and react-native.

kinyoklion:
There have been some changes to expo that cause the current versions to fail. (Even though the versions in the package.json did not change.)

Related expo issue: https://github.com/expo/expo/issues/31005

### DIFF
--- a/packages/sdk/react-native/example/package.json
+++ b/packages/sdk/react-native/example/package.json
@@ -26,7 +26,7 @@
     "expo": "51.0.28",
     "expo-status-bar": "~1.11.1",
     "react": "18.2.0",
-    "react-native": "0.75.0",
+    "react-native": "0.74.5",
     "react-native-dotenv": "^3.4.9"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


packages/sdk/react-native/example/package.json
```diff
@@ -23,10 +23,10 @@
   "dependencies": {
     "@launchdarkly/react-native-client-sdk": "workspace:^",
     "@react-native-async-storage/async-storage": "^1.21.0",
-    "expo": "51.0.21",
+    "expo": "51.0.28",
     "expo-status-bar": "~1.11.1",
     "react": "18.2.0",
-    "react-native": "0.74.3",
+    "react-native": "0.75.0",
```
kinyoklion, line 29: When running it says it is using 0.74.5, but if I put that here it fails in the same way.

brentvatne: hello! updating this to 0.74.5 should be sufficient

kinyoklion: Thank you. It does seem to be working with a clean repo with 0.74.5.

I wasn't running `npx expo prebuild --clean -p android`, but my main concern is someone doing a clone, or having done `git clean -xdf`  can build.

I forgot to clean when I downgraded back to 0.74.5 after trying 0.75.0.

PR #547, opened 2024-08-15
chore(test): Remove globals from mocks.

kinyoklion:
This removes the global platform, logger, clientContext, and hash.

I've not investigated fixing the event processor mock or streaming processor mock yet.

It also removes the global setupMocks requirement for the mocks package.

This PR, aside from the mocks package, should only change tests.

The mocks package is not published and can only be used within the repo, so we don't need to do a release.


packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.test.ts
```diff
@@ -1,8 +1,14 @@
 import { type EventName } from '@launchdarkly/js-client-sdk-common';
-import { logger } from '@launchdarkly/private-js-mocks';
+import { createLogger } from '@launchdarkly/private-js-mocks';
 
 import EventSource, { backoff, jitter } from './EventSource';
 
+let logger: ReturnType<typeof createLogger>;
```
kinyoklion, line 6: This lets the createX functions have inferred typed from typescript, so you can tell they are mocks and call the mock functions, while also allowing them to still be typed and provide intellisense.

Here logger is the return type of the type of the createLogger function.

packages/sdk/server-node/__tests__/LDClientNode.bucketing.test.ts
```diff
@@ -3,11 +3,12 @@
  * These tests exercise the hashing from the node platform.
  */
 import { integrations } from '@launchdarkly/js-server-sdk-common';
-import { TestData } from '@launchdarkly/js-server-sdk-common/dist/integrations';
```
kinyoklion, line 6: Fixed a few of the imports from dist. Generally speaking we do not want to do this.

packages/shared/sdk-client/src/LDClientImpl.events.test.ts
```diff
@@ -18,6 +17,17 @@ import * as mockResponseJson from './evaluation/mockResponse.json';
 import LDClientImpl from './LDClientImpl';
 import { Flags } from './types';
 
+type InputCustomEvent = internal.InputCustomEvent;
+type InputIdentifyEvent = internal.InputIdentifyEvent;
+
+let mockPlatform: ReturnType<typeof createBasicPlatform>;
```
kinyoklion, line 23: I did this rote conversion for  most of the tests. Most test suites could probably be further improved by reducing the scope of the mocks to specific fixtures (describe) or tests.

packages/shared/sdk-client/src/context/addAutoEnv.test.ts
```diff
@@ -1,16 +1,30 @@
...
+  type LDContext,
+  LDMultiKindContext,
+  LDUser,
+} from '@launchdarkly/js-sdk-common';
+import { createBasicPlatform, createLogger } from '@launchdarkly/private-js-mocks';
 
 import Configuration from '../configuration';
 import { addApplicationInfo, addAutoEnv, addDeviceInfo, toMulti } from './addAutoEnv';
 
+let mockPlatform: ReturnType<typeof createBasicPlatform>;
```
kinyoklion, line 13: The rename to mock was just to aid in progress being able to tell which files were not updated. Slowly removing all 'basicPlatform' references.

packages/shared/common/__tests__/internal/events/EventProcessor.test.ts
```diff
@@ -1,13 +1,42 @@
-import { clientContext, ContextDeduplicator } from '@launchdarkly/private-js-mocks';
```
kinyoklion, line 1: The global clientContext mock was only used in a few tests, so I removed the functionality entirely. The test can just specify its clientContext.

tanderson-ld (approved):

Issue #503, opened 2024-07-08
Caching on React native


Bardiamist:
I use l @launchdarkly/react-native-client-sdk@10.2.0 (also tested 10.1.5 and 10.1.3, same problem)
I need only get all flags once on app start and close connection.

```
const launchDarkly = new ReactNativeLDClient(config.launchDarklyKey, AutoEnvAttributes.Disabled);

await launchDarkly.identify({
  kind: 'user',
  key: userId,
});

const allFlags = launchDarkly.allFlags();

launchDarkly.close();
```

I noticed that: 
If close app and change some flag (tested on JSON flag) then wait even 30 minutes and launch app it will receive old cached data. Need to relauch app one more or even two times for receive the fresh data. Tested on iOS and Android, same problem.

Expected behaviour:
App should receive fresh data on start.

PR #548, opened 2024-08-16
feat: Allow waiting for the network response on identify.

kinyoklion:
Fixes: #503 
Fixes: #505 

When non-cached values are desired after identify the `waitForNetworkResults` identify option can be used. This will caused the identify to wait for configuration from the configured data source or timeout.

The recommended approach, for resilience and performance, will be to use cached flag values and respond to fresh values reactively and that will remain the default.


packages/shared/mocks/src/crypto.ts
```diff
@@ -1,12 +1,9 @@
 import type { Hasher } from '@common';
 
-// eslint-disable-next-line import/no-mutable-exports
```
kinyoklion, line 3: Forgot to remove this piece of global data in the test updates.

packages/shared/mocks/src/hasher.ts
kinyoklion, line 1: This file was redundant.

packages/shared/mocks/src/streamingProcessor.ts
```diff
@@ -16,6 +16,10 @@ export const setupMockStreamingProcessor = (
   errorTimeoutSeconds: number = 0,
   initTimeoutMs: number = 0,
 ) => {
+  let initTimeoutHandle: any;
```
kinyoklion, line 19: The simulated stream needs to be able to cancel its timers or it will run after tests are expected to end.

packages/shared/sdk-client/src/LDClientImpl.test.ts
```diff
@@ -49,19 +41,23 @@ const autoEnv = {
     os: { name: 'An OS', version: '1.0.1', family: 'orange' },
   },
 };
-let ldc: LDClientImpl;
-let defaultPutResponse: Flags;
-
 describe('sdk-client object', () => {
+  let ldc: LDClientImpl;
```
kinyoklion, line 45: Decreasing the scope of these to the fixture.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -354,15 +356,23 @@ export default class LDClientImpl implements LDClient {
...
+    }
 
     if (this.isOffline()) {
       if (loadedFromCache) {
-        this.logger.debug('Offline identify using storage flags.');
+        this.logger.debug('Offline identify - using cached flags.');
       } else {
-        this.logger.debug('Offline identify no storage. Defaults will be used.');
+        this.logger.debug(
+          'Offline identify - no cached flags, using defaults or already loaded flags.',
```
tanderson-ld, line 374:
```suggestion
          'Identify - Currently offline with no cached flags for provided context.  Maintaining current flags. This may result in default flag evaluations if no flags have been loaded before.'
```

Is the trailing comma intentional?

packages/shared/sdk-client/src/api/LDIdentifyOptions.ts
```diff
@@ -8,4 +8,16 @@ export interface LDIdentifyOptions {
...
    */
   timeout: number;
+
+  /**
+   * When true indicates that the SDK will attempt to wait for values from
+   * LaunchDarkly instead of depending on cached values. The cached values will
+   * still be loaded, but the promise returned by the identify function will not
+   * resolve as a result of those cached values being loaded. Generally this
+   * option should NOT be used and instead flag changes should be listened to.
+   * It the client is set to offline mode, then this option is ignored.
```
tanderson-ld, line 18:
```suggestion
   * If the client is set to offline mode, then this option is ignored.
```

tanderson-ld (changes requested):

packages/shared/sdk-client/src/LDClientImpl.ts
kinyoklion: Our linting requires training coma for parameters that are not on one line. The linter also forces this line to be wrapped as such.

tanderson-ld (approved):

packages/shared/sdk-client/src/api/LDIdentifyOptions.ts
```diff
@@ -8,4 +8,16 @@ export interface LDIdentifyOptions {
    * Defaults to 5 seconds.
    */
   timeout: number;
```
Bardiamist, line 10: @kinyoklion `timeout` should be optional then

kinyoklion: Unfortunately timeout was already implemented as non-optional. That said it defaults to 5, so setting it to 5 is equal to not providing it.

kinyoklion: We will consider it. 

There are advantages to having it non-optional and some SDKs require it with every call identify. Which makes it more obvious that the operation can timeout. We document it, but people are still often surprised that it times out.

PR #551, opened 2024-08-26
feat: Add support for Payload Filtering

keelerm84:


packages/shared/common/src/internal/events/EventSender.test.ts
```diff
@@ -25,6 +25,9 @@ const basicConfig = {
     analyticsEventPath: '/bulk',
     diagnosticEventPath: '/diagnostic',
     includeAuthorizationHeader: true,
+    getPollingUri: (_: string): string => '/sdk/latest-all',
+    getStreamingUri: (_: string): string => '/all',
+    getFilteredUri: (_: string): string => '',
```
keelerm84, line 30: Originally, the `getFilteredUri` was declared private in the `ServiceEndpoints` class. 

When that was private, I would get errors that an implementation was missing from here. I would add this line, but then it would complain that it was private in the class but not here. 

What is the syntax to resolve that? My editor didn't seem to like anything.

packages/shared/common/src/options/ServiceEndpoints.ts
```diff
@@ -36,12 +37,49 @@ export default class ServiceEndpoints {
...
+   * Constructs and returns the URI to be used for a streaming connection.
+   */
+  public getStreamingUri(path: string): string {
+    return this.getFilteredUri(`${this.streaming}${path}`);
+  }
+
+  /**
+   * Constructs and returns the URI to be used for a polling connection.
+   */
+  public getPollingUri(path: string): string {
```
keelerm84, line 61: I wasn't sure if you would want these types of methods here. Do you want me to add one for events to keep it consistent?

tanderson-ld:
I reviewed and don't have any comments.  Deferring to Ryan for approval since he is the js goto.

packages/shared/common/src/internal/events/EventSender.test.ts
kinyoklion: Yeah, things are a little different after adding implementations to the ServiceEndpoints. It was just a data class and all the fields were public. So you can make anything that looks like it and typescript structural typing will be happy. 

Once you have functions and private data the situation is different. 

You would have to make a class that was basically like service endpoints with the same public/private fields and methods. When you define getFilteredUri here you are making it public, but it needs to be private. The type isn't the same if it doesn't exist.

kinyoklion: Probably these should be changed to instances of the ServiceEndpoint type now that it isn't just data. (Probably already should have been.)

packages/shared/common/src/options/ServiceEndpoints.ts
kinyoklion: I think that we need to be careful adding them here. As is these methods are not correct for all client types, but this is common code.

kinyoklion: https://github.com/launchdarkly/flutter-client-sdk/blob/4534fb89e893cce8c1c790d1b8784d3b828ab8f6/packages/common_client/lib/src/config/defaults/io_config.dart#L14

https://github.com/launchdarkly/flutter-client-sdk/blob/4534fb89e893cce8c1c790d1b8784d3b828ab8f6/packages/common_client/lib/src/config/defaults/js_config.dart#L14

For instance compare these io/js formats for client-side.

packages/shared/common/src/internal/events/EventSender.test.ts
kinyoklion: (After going forward I am leaning toward we keep this as just a data class.)

packages/shared/common/src/internal/stream/StreamingProcessor.ts
```diff
@@ -37,6 +38,7 @@ class StreamingProcessor implements LDStreamProcessor {
     sdkKey: string,
     clientContext: ClientContext,
     streamUriPath: string,
+    parameters: { key: string; value: string }[],
```
keelerm84, line 41: I put this after the URI because it felt like a better spot for it. Happy to move it if you don't like all the shifting parameters.

packages/shared/mocks/src/platform.ts
```diff
@@ -1,11 +1,7 @@
-import type { Encoding, Platform, PlatformData, Requests, SdkData, Storage } from '@common';
```
keelerm84, line 1: Linter complained about all this being unused.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -401,15 +401,17 @@ export default class LDClientImpl implements LDClient {
     identifyResolve: any,
     identifyReject: any,
   ) {
-    let pollingPath = this.createPollUriPath(context);
+    let parameters: { key: string; value: string }[] = [];
     if (this.config.withReasons) {
-      pollingPath = `${pollingPath}?withReasons=true`;
+      parameters = [{ key: 'withReasons', value: 'true' }];
```
kinyoklion, line 406: nit: I would probably push to the existing array in both these cases.

kinyoklion (approved):

PR #552, opened 2024-08-27
feat: Make timeout optional in LDIdentifyOptions.

kinyoklion:
In the strictest sense this would be a breaking change, but in the practical I think it is not likely to be.

The way this would be breaking is if someone was using this for something other than just passing to the identify function. This would be somewhat unusual.

Example:
```
interface MyOptions: {
  timeout: number;
}

let myOptions: MyOptions = someLdIdentifyOptions;
```

For standard usage this will not be breaking. 

I am tagging this as a feat instead of a fix to prevent any breakage for someone floating patches and also doing unusual things with the type.

### DIFF
--- a/packages/shared/sdk-client/src/api/LDIdentifyOptions.ts
+++ b/packages/shared/sdk-client/src/api/LDIdentifyOptions.ts
@@ -7,7 +7,7 @@ export interface LDIdentifyOptions {
    *
    * Defaults to 5 seconds.
    */
-  timeout: number;
+  timeout?: number;
 
   /**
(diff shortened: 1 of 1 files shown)


kinyoklion:
[Bardiamist](https://github.com/Bardiamist) requested this in another PR.

kinyoklion:
There is another issue, which is:
```
if (identifyOptions?.timeout) {
```

The timeout is being truthy checked, which means a timeout of 0 is not accepted. Changing this would be a more severe breaking change, so that will have to be reserved for a breaking SDK release.

tanderson-ld (approved):

PR #554, opened 2024-08-27
feat: Correct client evaluation typings.

kinyoklion:
Some code was refactored to make server and client share evaluation results. This should not have been done because the results are not the same in client and server SDKs. Server SDKs can always produce a reason and client SDKs cannot.

This meant that the typing said that reason was required in the client SDK, but it could be null.

This is a 'feat' to ensure a minor version bump in case of minor incompatibilities.


packages/shared/common/src/api/data/LDEvaluationDetail.ts
```diff
@@ -1,6 +1,8 @@
 import { LDEvaluationReason } from './LDEvaluationReason';
 import { LDFlagValue } from './LDFlagValue';
 
+// TODO: On major version change "variationIndex" to only be optional and not nullable.
```
kinyoklion, line 4: The variationIndex can be null because we maintained the 7.0 API when we released 8. We should change it to just be optional.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -483,7 +485,7 @@ export default class LDClientImpl implements LDClient {
     defaultValue: any,
     eventFactory: EventFactory,
     typeChecker?: (value: any) => [boolean, string],
-  ): LDFlagValue {
+  ): LDEvaluationDetail {
```
kinyoklion, line 488: This type was wrong, basically it was `any` which means that the type checker could not reveal the problem.

packages/shared/sdk-client/src/LDEmitter.test.ts
kinyoklion, line 1: This file is not used outside of the common client code, and therefore is not part of the API.

packages/shared/sdk-client/src/api/LDEvaluationDetail.ts
```diff
@@ -0,0 +1,37 @@
...
+  LDEvaluationDetail as CommonDetail,
+  LDEvaluationDetailTyped as CommonDetailTyped,
+  LDEvaluationReason,
+} from '@launchdarkly/js-sdk-common';
+
+// Implementation note: In client-side SDKs the reason is optional. The common type, which is also
+// used by server SDKs, has a required reason. This file contains a client specific
+// LDEvaluationDetail which has an optional reason.
+
+// TODO: On major version change "reason" to be optional instead of nullable.
```
kinyoklion, line 11: Reason should have been optional from day 1, but it was not. 

Second it should not be null, but that is what the code was actually returning, so it will continue to do so.

The null fields were only for compatibility with the pre-typescript node API.

packages/shared/sdk-client/src/api/LDInspection.ts
kinyoklion, line 1: Inspectors were not implemented, and this is not in the interface, so remove it until we implement them. (Unless we only implement hooks.)

packages/shared/sdk-client/src/configuration/validators.ts
kinyoklion, line 1: Removed unimplemented fields.

packages/shared/sdk-server/src/evaluation/EvalResult.ts
```diff
@@ -31,11 +30,22 @@ export default class EvalResult {
   }
 
   static forError(errorKind: internal.ErrorKinds, message?: string, def?: any): EvalResult {
-    return new EvalResult(true, createErrorEvaluationDetail(errorKind, def), message);
+    return new EvalResult(
```
kinyoklion, line 33: Changed this code back to the way it was before the refactoring.

packages/shared/sdk-client/src/api/LDEvaluationDetail.ts
```diff
@@ -0,0 +1,37 @@
...
+  reason: LDEvaluationReason | null;
+};
+
+/**
+ * An object that combines the result of a feature flag evaluation with information about
+ * how it was calculated.
+ *
+ * This is the result of calling detailed variation methods.
+ */
+export type LDEvaluationDetailTyped<TFlag> = Omit<CommonDetailTyped<TFlag>, 'reason'> & {
```
tanderson-ld, line 32: Is the `LDEvaluationDetailTyped as CommonDetailTyped` required to make the type checking / compilation happy?

packages/shared/sdk-client/src/api/LDInspection.ts
tanderson-ld, line 1: Not consumable publicly?

tanderson-ld (approved):

packages/shared/sdk-client/src/api/LDEvaluationDetail.ts
kinyoklion: I am exporting a new type with the same name as the original type. So I need to import the original type under a different name, otherwise it will conflict.

packages/shared/sdk-client/src/api/LDInspection.ts
kinyoklion: It was exported, but it wasn't in the options and it wasn't used. 

So, another technically breaking change, but unrelated to any usable code.

We may not actually implement inspectors at all, as hooks have now replaced them.

PR #555, opened 2024-08-28
chore: Scaffold browser SDK project.

kinyoklion:
For now I am using Vite to build. If we encounter problems with this we can move to a rollup based build like we are using in other projects. Vite is easier to use and is reasonably popular, so I think it is a good starting point.

Currently all tests in the monorepo use jest, so this continues to use jest. We may want to consider, in the future, if we want to move to vitest if it supports everything we need. This would make the testing configuration simpler.


packages/sdk/browser/__tests__/index.test.ts
kinyoklion, line 1: I am putting tests in their own folder. I am considering doing a consistency sweep to put the tests in a `__tests__` folder in all the packages. It is an older style, but it does make it simpler to keep test functionality isolated from non-test functionality.

tanderson-ld (approved):

PR #561, opened 2024-09-03
fix: Conditionally access constructor name.

kinyoklion:


packages/telemetry/browser-telemetry/src/BrowserTelemetryImpl.ts
```diff
@@ -134,13 +134,23 @@ export default class BrowserTelemetryImpl implements BrowserTelemetry {
...
-      message: exception.message,
-      stack: parse(exception, this.options.stack),
-      breadcrumbs: [...this.breadcrumbs],
-      sessionId: this.sessionId,
-    };
+    const validException = exception !== undefined && exception !== null;
+
+    const data: ErrorData = validException
+      ? {
+          type: exception.name || exception.constructor?.name || 'generic',
```
kinyoklion, line 141: Theoretically the `constructor` should not be able to be null or undefined, but potentially it could be with some prototype pollution.

erangeles (approved):

PR #562, opened 2024-09-03
fix: Use validated exception data in dispatchError.

kinyoklion:


packages/telemetry/browser-telemetry/src/BrowserTelemetryImpl.ts
```diff
@@ -138,21 +141,22 @@ export default class BrowserTelemetryImpl implements BrowserTelemetry {
 
     const data: ErrorData = validException
       ? {
-          type: exception.name || exception.constructor?.name || 'generic',
-          message: exception.message,
+          type: exception.name || exception.constructor?.name || GENERIC_EXCEPTION,
+          // Only coalesce null/undefined, not empty.
+          message: exception.message ?? MISSING_MESSAGE,
```
kinyoklion, line 146: In practice this shouldn't happen, but may was well be ultra safe.

packages/telemetry/browser-telemetry/src/BrowserTelemetryImpl.ts
```diff
@@ -138,21 +141,22 @@ export default class BrowserTelemetryImpl implements BrowserTelemetry {
...
-          message: 'null or undefined exception',
+          type: GENERIC_EXCEPTION,
+          message: NULL_EXCEPTION_MESSAGE,
           stack: { frames: [] },
           breadcrumbs: [...this.breadcrumbs],
           sessionId: this.sessionId,
         };
     this.capture(ERROR_KEY, data);
-    this.dispatchError(exception);
+    this.dispatchError(data);
```
kinyoklion, line 159: This will have the already checked type and message.

erangeles (approved):

PR #563, opened 2024-09-03
chore: Build releases in dependency order.

kinyoklion:
This contains several conditions that appear to be illogical, but are required by github actions. For some background this issue can be read:
https://github.com/actions/runner/issues/491

`always()`:

Imagine two packages. Package "a" and package "b".

"b" depends on "a" and we want to automatically release "a" and "b" using release please. We make "b" `needs` "a", so that "b" will build after "a". But if we do that, then "b" will not build unless "a" also builds. We often need to release "b" even though there are no changes to "a".

By using `if: always()` we can cause "b" to run after "a" even when "a" doesn't run.

`failure()` and `canceled()`.

With `always()` package "b" will be built and released even if package "a" failed to build. You would expect that to only require `failure()`, but it seems that it only works if you have `canceled()` as well. (It may be we don't actually need failure.)

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -52,11 +52,11 @@ jobs:
 
   release-sdk-client:
     runs-on: ubuntu-latest
-    needs: ['release-please']
+    needs: ['release-please', 'release-common']
     permissions:
       id-token: write
       contents: write
-    if: ${{ needs.release-please.outputs.package-sdk-client-released == 'true'}}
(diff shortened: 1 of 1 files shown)


.github/workflows/release-please.yml
```diff
@@ -339,7 +339,7 @@ jobs:
 
   release-tooling-react-universal:
```
kinyoklion, line 340: This build is disabled. I didn't update the condition, but will add it to the condition comment.

keelerm84 (approved):
I believe you.

PR #566, opened 2024-09-04
feat: Add support for localStorage for the browser platform.

kinyoklion:


packages/sdk/browser/jest.config.js
```diff
@@ -0,0 +1,12 @@
+
+export default {
```
kinyoklion, line 2: Added some missing config and dependencies to allow imports in jest.

packages/sdk/browser/src/platform/BrowserPlatform.ts
```diff
@@ -0,0 +1,22 @@
+import {
+  LDOptions,
+  Storage,
+  /* platform */
+} from '@launchdarkly/js-client-sdk-common';
+
+import LocalStorage, { isLocalStorageSupported } from './LocalStorage';
+
+export default class BrowserPlatform /* implements platform.Platform */ {
```
kinyoklion, line 9: Once we implement enough to support the platform interface, then we can uncomment some things.

tanderson-ld (approved):

PR #567, opened 2024-09-04
fix: Flag store should not access values from prototype. 

kinyoklion:
This should be merged after the test organization PR.


packages/shared/sdk-client/__tests__/flag-manager/FlagStore.test.ts
```diff
@@ -0,0 +1,60 @@
+import { DefaultFlagStore } from '../../src/flag-manager/FlagStore';
+
+describe('given an empty flag store', () => {
+  let store: DefaultFlagStore;
+
+  beforeEach(() => {
+    store = new DefaultFlagStore();
+  });
+
+  it.each(['unknown', 'toString', 'length'])(
```
kinyoklion, line 10: The 'toString' test would have failed here before.

'toString' is a member of the prototype of objects. `{}['toString']` exists and is a function.

Once a value of 'toString' is set, then it can be used as a flag, but the object has the property at that point. Versus the inheriting it from Object.

tanderson-ld (approved):

PR #573, opened 2024-09-06
feat: Add platform support for async hashing.

kinyoklion:
This adds platform support for async hashing for use in client-side SDKs.

It does not implement async hashing for any existing platform, but provides it as an option to allow for use of standard browser APIs. Allowing the usage of standard browser crypto APIs means that browser SDKs will not need to include an additional dependency to replicate built-in functionality.

### DIFF
--- a/packages/shared/common/src/api/platform/Crypto.ts
+++ b/packages/shared/common/src/api/platform/Crypto.ts
@@ -7,7 +7,19 @@
  */
 export interface Hasher {
   update(data: string): Hasher;
-  digest(encoding: string): string;
+  /**
+   * Note: All server SDKs MUST implement synchronous digest.
+   *
+   * Server SDKs have high performance requirements for bucketing users.
+   */
+  digest?(encoding: string): string;
+
+  /**
+   * Note: Client-side SDKs MUST implement either synchronous or asynchronous digest.
+   *
+   * Client SDKs do not have high throughput hashing operations.
+   */
+  asyncDigest?(encoding: string): Promise<string>;
 }
 
 /**
@@ -17,7 +29,7 @@ export interface Hasher {
  *
  * The has implementation must support digesting to 'hex'.
  */
-export interface Hmac extends Hasher {
+export interface Hmac {
   update(data: string): Hasher;
   digest(encoding: string): string;
 }
@@ -27,6 +39,9 @@ export interface Hmac extends Hasher {
  */
 export interface Crypto {
   createHash(algorithm: string): Hasher;
-  createHmac(algorithm: string, key: string): Hmac;
+  /**
+   * Note: Server SDKs MUST implement createHmac.
+   */
+  createHmac?(algorithm: string, key: string): Hmac;
   randomUUID(): string;
 }
--- a/packages/shared/sdk-client/src/crypto/digest.ts
+++ b/packages/shared/sdk-client/src/crypto/digest.ts
@@ -1,6 +1,6 @@
 import { Hasher } from '@launchdarkly/js-sdk-common';
 
-export default function digest(hasher: Hasher, encoding: string) {
+export default async function digest(hasher: Hasher, encoding: string): Promise<string> {
   if (hasher.digest) {
     return hasher.digest(encoding);
   }
--- a/packages/shared/sdk-client/src/crypto/digest.ts
+++ b/packages/shared/sdk-client/src/crypto/digest.ts
@@ -1,0 +1,12 @@
+import { Hasher } from '@launchdarkly/js-sdk-common';
+
+export default function digest(hasher: Hasher, encoding: string) {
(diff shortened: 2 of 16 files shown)


packages/shared/common/src/api/platform/Crypto.ts
```diff
@@ -27,6 +39,9 @@ export interface Hmac extends Hasher {
  */
 export interface Crypto {
   createHash(algorithm: string): Hasher;
-  createHmac(algorithm: string, key: string): Hmac;
+  /**
+   * Note: Server SDKs MUST implement createHmac.
+   */
+  createHmac?(algorithm: string, key: string): Hmac;
```
kinyoklion, line 45: I think we *could* create Hmac on all platforms, but we don't actually need to.

packages/shared/common/src/api/platform/Crypto.ts
```diff
@@ -7,7 +7,19 @@
  */
 export interface Hasher {
```
kinyoklion, line 8: This is a little clunky. It would be nice if as much was statically checked as possible, but the platforms are not likely to be wrong outside initial implementation. We do have a couple other dynamic elements.

packages/shared/common/src/api/platform/Crypto.ts
kinyoklion: Basically I would prefer implementation of this not block or slow down platform implementations, and I would like for SDKs to not have unused bloat to implement it.

packages/shared/sdk-client/src/flag-manager/FlagManager.ts
```diff
@@ -31,10 +31,26 @@ export default class FlagManager {
     logger: LDLogger,
     private readonly timeStamper: () => number = () => Date.now(),
   ) {
-    const environmentNamespace = namespaceForEnvironment(platform.crypto, sdkKey);
-
     this.flagUpdater = new FlagUpdater(this.flagStore, logger);
-    this.flagPersistence = new FlagPersistence(
+    this.flagPersistencePromise = this.initPersistence(
```
kinyoklion, line 35: Assign the promise synchronously and then await it when persistence needs to be used.

Getting the promise immediately removes any race conditions you would have with code that awaits and assigns the value.

kinyoklion: This is async now as generating the hash can be async.

packages/shared/sdk-client/src/crypto/digest.ts
```diff
@@ -0,0 +1,12 @@
+import { Hasher } from '@launchdarkly/js-sdk-common';
+
+export default function digest(hasher: Hasher, encoding: string) {
```
tanderson-ld, line 3: Can someone accidentally not await a call to digest?

packages/shared/sdk-server/src/BigSegmentsManager.ts
```diff
@@ -144,6 +144,10 @@ export default class BigSegmentsManager {
   private hashForUserKey(userKey: string): string {
     const hasher = this.crypto.createHash('sha256');
     hasher.update(userKey);
+    if (!hasher.digest) {
+      // This represents an error in platform implementation.
+      throw new Error('Platform must implement digest or asyncDigest');
+    }
```
tanderson-ld, line 150: Is there a straightforward way to make a default digest implementation that throws the error to centralize this case?  I suppose that would advocate for hasher.digest and hasher.asyncDigest becoming non-optional with default erroring implementations.

kinyoklion: If you do that, then you cannot choose one to call without throwing an exception.

packages/shared/sdk-client/src/crypto/digest.ts
kinyoklion: Digest isn't async, so you don't have to await it.

But, generally speaking, there isn't anything that forces you to await async functions, and often times you may not want to.

If you think you awaited, and you didn't, then your result type would be wrong though.

kinyoklion: I forgot the type annotation on this.

kinyoklion: As picky as our linter is it doesn't complain about this, which is arguably more important than 90% of what it does complain about.

packages/shared/sdk-server/src/BigSegmentsManager.ts
kinyoklion: It could be centralized similar to the client, but it didn't seem worth it for the limited call sites, and that it is just a pass-through call, versus the tiny bit of logic required in the client.

tanderson-ld (approved):

PR #574, opened 2024-09-06
feat: Implement browser crypto and encoding.

kinyoklion:


packages/sdk/browser/package.json
```diff
@@ -27,15 +27,15 @@
   ],
   "scripts": {
     "clean": "rimraf dist",
-    "build": "vite build",
+    "build": "tsc --noEmit && vite build",
```
kinyoklion, line 30: Need to have `tsc --noEmit` to validate the typescript `vite` just builds it and doesn't type check.

packages/sdk/browser/package.json
```diff
@@ -27,15 +27,15 @@
...
     "lint": "eslint . --ext .ts,.tsx",
     "prettier": "prettier --write '**/*.@(js|ts|tsx|json|css)' --ignore-path ../../../.prettierignore",
     "test": "jest",
     "coverage": "yarn test --coverage",
     "check": "yarn prettier && yarn lint && yarn build && yarn test"
   },
   "dependencies": {
-    "@launchdarkly/js-client-sdk-common": "1.5.0"
+    "@launchdarkly/js-client-sdk-common": "1.7.0"
```
kinyoklion, line 38: Until we automate releases we need to manually update this or it will used the published version instead of the repo version. We could enable release-please and just not publish.

packages/sdk/browser/src/platform/randomUuidV4.ts
```diff
@@ -0,0 +1,99 @@
+// The implementation in this file generates UUIDs in v4 format and is suitable
```
kinyoklion, line 1: This code was implemented for error monitoring, but that is still on a feature branch.

I think, after the client is working, we need to make a `browser-common` containing things that can be used in browser SDKs.

packages/sdk/browser/tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "declaration": true,
     "declarationMap": true,
     "jsx": "react-jsx",
-    "lib": ["es6", "dom"],
+    "lib": ["ES2017", "dom"],
```
kinyoklion, line 7: I think that es2017 should be an acceptable target, but we can lower it if we need to.

If we do lower it, then we would need to remove the TextEncoder usage.

tanderson-ld (approved):

PR #577, opened 2024-09-10
feat: Add support for conditional event source capabilities.

kinyoklion:
This adds the ability to specify what capabilities an even source supports.

The idea is, for something like report, the common code can check if the event source supports it. If it is configured, and the event source doesn't support it, then it can take appropriate action.

The browser SDK can calculate this at runtime by checking for a polyfill and filling out the appropriate capabilities.

### DIFF
--- a/packages/sdk/react-native/src/platform/PlatformRequests.ts
+++ b/packages/sdk/react-native/src/platform/PlatformRequests.ts
@@ -1,6 +1,7 @@
 import type {
   EventName,
   EventSource,
+  EventSourceCapabilities,
   EventSourceInitDict,
   LDLogger,
   Options,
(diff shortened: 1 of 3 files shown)


packages/sdk/react-native/src/platform/PlatformRequests.ts
```diff
@@ -21,6 +22,14 @@ export default class PlatformRequests implements Requests {
     });
   }
 
+  getEventSourceCapabilities(): EventSourceCapabilities {
+    return {
+      readTimeout: false,
```
kinyoklion, line 27: It has a timeout, but that is a timeout for the XmlHttpRequest.

Which I think is the duration of the entire request.

tanderson-ld (approved):

PR #578, opened 2024-09-10
feat: Implement support for browser requests.

kinyoklion:


packages/sdk/browser/src/platform/DefaultBrowserEventSource.ts
```diff
@@ -0,0 +1,106 @@
+import {
```
kinyoklion, line 1: For now I am not adding tests for this. Running the tests in node would require using an event source polyfill, effectively means it won't be testing much reality.

As a result it is likely some things will need to be tweaked when it is used for real.

kinyoklion: I have done some manual testing in a browser.

tanderson-ld (approved):

PR #582, opened 2024-09-13
feat: Add support for browser contract tests.

kinyoklion:
This adds the contract tests, but does not yet automate running them.


packages/sdk/browser/contract-tests/entity/src/ClientEntity.ts
```diff
@@ -0,0 +1,231 @@
+import {
```
kinyoklion, line 1: This is mostly a typescript port of the contract test service from node-client-sdk + typed evaluations.

tanderson-ld (approved):

PR #585, opened 2024-09-18
feat: Implement goals for client-side SDKs.

kinyoklion:
Implements goals. Additionally adds browser specific configuration as it is required by goals.

Also addresses SDK-563


packages/sdk/browser/__tests__/platform/LocalStorage.test.ts
kinyoklion, line 1: I made some changes to the jest configuration to handle ESM dependencies and it required importing jest types to make things happy.

packages/sdk/browser/src/goals/GoalTracker.ts
```diff
@@ -0,0 +1,100 @@
+import escapeStringRegexp from 'escape-string-regexp';
```
kinyoklion, line 1: The previous SDK was using this dependency, and I checked the source of it and it seems very reasonable. It has no runtime dependencies and is a single function. The function just does a couple replacements. I don't see a reason to vendor it.

Including it did increase the complexity of the jest test running.

packages/shared/sdk-client/src/LDClientImpl.ts
kinyoklion, line 1: I've added these functions to the client or now, but I think we maybe actually need to move to a PIMPL style implementation like we use in fluter and C++. Which would allow us to expose more internals more safely.

packages/sdk/browser/src/goals/GoalManager.ts
```diff
@@ -0,0 +1,57 @@
...
+  constructor(
+    credential: string,
+    private readonly requests: Requests,
+    baseUrl: string,
+    private readonly reportError: (err: Error) => void,
+    private readonly reportGoal: (url: string, goal: Goal) => void,
+    locationWatcherFactory: (cb: () => void) => LocationWatcher = (cb) =>
+      new DefaultLocationWatcher(cb),
+  ) {
+    // TODO: Generate URL in a better way.
```
kinyoklion, line 22: Maybe we can address this in the HTTP properties change as well. 

We never have used ServiceEndpoints in a browser SDK and it is clear that goals wasn't considered. Polling/baseUri happen to be interchangeable.

packages/sdk/browser/src/BrowserClient.ts
```diff
@@ -17,18 +20,77 @@ import BrowserPlatform from './platform/BrowserPlatform';
...
 export class BrowserClient extends LDClientImpl {
+  private readonly goalManager?: GoalManager;
   constructor(
     private readonly clientSideId: string,
     autoEnvAttributes: AutoEnvAttributes,
-    options: LDOptions = {},
+    options: BrowserOptions = {},
   ) {
-    super(clientSideId, autoEnvAttributes, new BrowserPlatform(options), options, {
+    const { logger: customLogger, debug } = options;
```
kinyoklion, line 29: Logging is a chicken/egg problem. I think a switch to PIMPL style for the client could help. 

I am not 100% sure that we will actually use the common client-side client by the time things are sorted (initialization and bootstrap work quite different), but the same thing applied to react native.

packages/sdk/browser/src/goals/GoalTracker.ts
```diff
@@ -0,0 +1,100 @@
...
+ * Tracks the goals on an individual "page" (combination of route, query params, and hash).
+ */
+export default class GoalTracker {
+  private clickHandler?: (event: Event) => void;
+  constructor(goals: Goal[], onEvent: EventHandler) {
+    const goalsMatchingUrl = goals.filter((goal) =>
+      goal.urls?.some((matcher) =>
+        matchesUrl(matcher, window.location.href, window.location.search, window.location.hash),
+      ),
+    );
```
tanderson-ld, line 73: Consider filtering in GoalManager before passing goals in.

packages/sdk/browser/src/goals/LocationWatcher.ts
```diff
@@ -0,0 +1,60 @@
+export const LOCATION_WATCHER_INTERVAL = 300;
```
tanderson-ld, line 1: Consider specifying time unit.

tanderson-ld (approved):

PR #589, opened 2024-09-23
fix: Ensure browser contract tests run during top-level build.

kinyoklion:


packages/sdk/browser/contract-tests/entity/package.json
```diff
@@ -6,6 +6,7 @@
   "description": "Contract test service implementation for @launchdarkly/js-client-sdk",
   "scripts": {
     "start": "vite --open=true",
+    "build": "tsc --noEmit && vite build",
```
kinyoklion, line 9: There wasn't a build command, so it didn't run with the root build.

packages/sdk/browser/contract-tests/entity/src/ClientEntity.ts
kinyoklion, line 1: I forgot to update this when I removed AutoEnvAttributes.

packages/sdk/browser/contract-tests/entity/package.json
```diff
@@ -5,7 +5,8 @@
   "type": "module",
   "description": "Contract test service implementation for @launchdarkly/js-client-sdk",
   "scripts": {
-    "start": "vite --open=true",
+    "start": "tsc --noEmit && vite --open=true",
```
kinyoklion, line 8: Make sure the typescript it valid when running the dev server.

packages/sdk/browser/contract-tests/entity/src/ClientEntity.ts
```diff
@@ -72,6 +66,8 @@ function makeSdkConfig(options: SDKConfigParams, tag: string) {
     };
   }
 
+  cf.fetchGoals = false;
```
kinyoklion, line 69: Don't fetch goals. It confuses the test harness.

packages/sdk/browser/src/BrowserClient.ts
```diff
@@ -135,8 +136,8 @@ export class BrowserClient extends LDClientImpl {
     };
   }
 
-  override async identify(context: LDContext): Promise<void> {
-    await super.identify(context);
+  override async identify(context: LDContext, identifyOptions?: LDIdentifyOptions): Promise<void> {
```
kinyoklion, line 139: I missed this, which doesn't make the SDK not work, but does cause problems for the test harness.

tanderson-ld (approved):

PR #590, opened 2024-09-23
feat: adds datasource status to sdk-client

tanderson-ld:
**Requirements**

**Related issues**

SDK-170

Adds DataSourceStatusManager.
Refactors data source errors into common.
Adds DataSourceErrorKind to classify errors so manager can track state.

### DIFF
--- a/packages/shared/common/src/internal/datasource/DataSourceErrorKinds.ts
+++ b/packages/shared/common/src/internal/datasource/DataSourceErrorKinds.ts
@@ -1,0 +1,20 @@
+export enum DataSourceErrorKind {
+  /// An unexpected error, such as an uncaught exception, further
+  /// described by the error message.
+  Unknown,
+
+  /// An I/O error such as a dropped connection.
+  NetworkError,
+
+  /// The LaunchDarkly service returned an HTTP response with an error
+  /// status, available in the status code.
+  ErrorResponse,
+
+  /// The SDK received malformed data from the LaunchDarkly service.
+  InvalidData,
+
+  /// The data source itself is working, but when it tried to put an
+  /// update into the data store, the data store failed (so the SDK may
+  /// not have cached the latest data).
+  StoreError,
+}
--- a/packages/shared/common/src/internal/datasource/errors.ts
+++ b/packages/shared/common/src/internal/datasource/errors.ts
@@ -1,0 +1,39 @@
+/* eslint-disable max-classes-per-file */
+import { DataSourceErrorKind } from './DataSourceErrorKinds';
+
+export class LDFileDataSourceError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = 'LaunchDarklyFileDataSourceError';
+  }
+}
+
+export class LDPollingError extends Error {
+  public readonly kind: DataSourceErrorKind;
+  public readonly status?: number;
+  public readonly recoverable: boolean;
+
+  constructor(kind: DataSourceErrorKind, message: string, status?: number, recoverable = true) {
+    super(message);
+    this.kind = kind;
+    this.status = status;
+    this.name = 'LaunchDarklyPollingError';
+    this.recoverable = recoverable;
+  }
+}
+
+export class LDStreamingError extends Error {
+  public readonly kind: DataSourceErrorKind;
+  public readonly code?: number;
+  public readonly recoverable: boolean;
+
+  constructor(kind: DataSourceErrorKind, message: string, code?: number, recoverable = true) {
+    super(message);
+    this.kind = kind;
+    this.code = code;
+    this.name = 'LaunchDarklyStreamingError';
+    this.recoverable = recoverable;
+  }
+}
+
+export type StreamingErrorHandler = (err: LDStreamingError) => void;
--- a/packages/shared/sdk-client/__tests__/LDClientImpl.storage.test.ts
+++ b/packages/shared/sdk-client/__tests__/LDClientImpl.storage.test.ts
@@ -89,8 +89,7 @@ describe('sdk-client storage', () => {
 
     expect(mockPlatform.storage.get).toHaveBeenCalledWith(flagStorageKey);
 
-    // 'change' should not have been emitted
-    expect(emitter.emit).toHaveBeenCalledTimes(2);
+    expect(emitter.emit).toHaveBeenCalledTimes(3);
     expect(emitter.emit).toHaveBeenNthCalledWith(1, 'change', context, defaultFlagKeys);
     expect(emitter.emit).toHaveBeenNthCalledWith(
       2,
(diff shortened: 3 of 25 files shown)


packages/shared/sdk-client/src/LDEmitter.ts
```diff
@@ -1,6 +1,7 @@
 import { LDLogger } from '@launchdarkly/js-sdk-common';
 
-export type EventName = 'error' | 'change';
+// TODO: is one emitter with many event names what we want long term?
+export type EventName = 'error' | 'change' | 'dataSourceStatus';
```
tanderson-ld, line 4: For reviewers: Please provide thoughts.

packages/shared/sdk-client/src/datasource/DataSourceEventHandler.ts
tanderson-ld, line 1: For reviewers: Content for this file is from LDClientImpl

packages/shared/sdk-client/src/datasource/DataSourceEventHandler.ts
```diff
@@ -0,0 +1,66 @@
...
+    });
+  }
+
+  handleStreamingError(error: LDStreamingError) {
+    this.statusManager.setError(error.kind, error.message, error.code, error.recoverable);
+  }
+
+  handlePollingError(error: LDPollingError) {
+    this.statusManager.setError(error.kind, error.message, error.status, error.recoverable);
+  }
```
tanderson-ld, line 65: For reviewers: If there is a better way to handle the mapping of StreamingError.code and PollingError.status to status I can unify these functions.

packages/shared/sdk-client/src/polling/PollingProcessor.ts
```diff
@@ -14,7 +14,9 @@ import { PollingDataSourceConfig } from '../streaming/DataSourceConfig';
 import { Flags } from '../types';
 import Requestor, { LDRequestError } from './Requestor';
 
-export type PollingErrorHandler = (err: LDPollingError) => void;
+// TODO: revisit usage of internal and figure out best practice
+const { DataSourceErrorKind, LDPollingError } = internal;
+export type PollingErrorHandler = (err: internal.LDPollingError) => void;
```
tanderson-ld, line 19: For reviewers: Please provide thoughts on internal import usage here.

packages/shared/common/src/internal/datasource/DataSourceErrorKinds.ts
```diff
@@ -0,0 +1,20 @@
...
+  /// status, available in the status code.
+  ErrorResponse,
+
+  /// The SDK received malformed data from the LaunchDarkly service.
+  InvalidData,
+
+  /// The data source itself is working, but when it tried to put an
+  /// update into the data store, the data store failed (so the SDK may
+  /// not have cached the latest data).
+  StoreError,
```
tanderson-ld, line 19: TODO: verify `StoreError` is necessary and if so, where is the error case missing.

tanderson-ld:
Looking into test failures.

packages/shared/common/src/internal/datasource/DataSourceErrorKinds.ts
tanderson-ld: Removed.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -172,64 +170,6 @@ export default class LDClientImpl implements LDClient {
     return this.checkedContext;
   }
 
-  private createStreamListeners(
```
tanderson-ld, line 175: For reviewers: This was unused.

packages/shared/common/src/internal/datasource/errors.ts
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable max-classes-per-file */
+import { DataSourceErrorKind } from './DataSourceErrorKinds';
+
+export class LDFileDataSourceError extends Error {
```
kinyoklion, line 4: Errors are not intended to be internal. In that the internal package should only be used by SDK implementations, but `instance of LDPollingError` theoretically is something you could do in an application. Maybe we don't want that, but this one is maybe breaking.

kinyoklion: Maybe not expected to be constructed outside though. That part is likely fine.

packages/shared/sdk-client/__tests__/LDClientImpl.storage.test.ts
```diff
@@ -256,8 +251,8 @@ describe('sdk-client storage', () => {
       JSON.stringify(putResponse),
     );
 
-    expect(emitter.emit).toHaveBeenNthCalledWith(1, 'change', context, defaultFlagKeys);
```
kinyoklion, line 259: I may need some explanation on these changes also, as the ordering of events mattered and that isn't tested anymore.

packages/shared/sdk-client/src/LDEmitter.ts
kinyoklion: Yeah, this is intended to provide an EventTarget like interface to our objects to make them familiar. https://developer.mozilla.org/en-US/docs/Web/API/EventTarget

packages/shared/sdk-client/src/index.ts
```diff
@@ -30,4 +31,4 @@ export { DataSourcePaths } from './streaming';
 export { BaseDataManager } from './DataManager';
 export { Requestor };
 
-export { LDClientImpl, LDClientInternalOptions };
+export { DataSourceStatus, LDClientImpl, LDClientInternalOptions };
```
tanderson-ld, line 34: Need to add DataSourceState, DataSourceStatusErrorInfo, and DataSourceErrorKind

packages/shared/sdk-client/__tests__/LDClientImpl.storage.test.ts
tanderson-ld: I thought about that too.  I can go through and get them all to be passing, but I don’t think there is any specification around when a change event should come out vs when a dataSourceStatus change should come out and that is leading to them all breaking as I added dataSourceStatus events.

Perhaps an improved way could be to filter to just “change” and then do positional checking on the filtered output.

packages/shared/sdk-client/src/LDEmitter.ts
tanderson-ld: Removed TODO

packages/shared/common/src/internal/datasource/errors.ts
tanderson-ld: Started fixing this, but running into some import/export configuration issue.  Pushed most recent commit before having to leave for flight.

kinyoklion (approved):

PR #591, opened 2024-09-24
feat: Refactor data source connection handling. 

kinyoklion:
Most of this PR is refactoring to move data source handling our of the common code. This also moves functionality, such as `setConnectionMode` out of the common code.

Currently the MobileDataManager is in the RN SDK, but it should be moved to some kind of non-browser common when we implement SDKs like node or electron.

A large amount of the code in this PR is tests for data source handling and moving existing code.

This PR does not add automatic starting/stopping of the stream.

BEGIN_COMMIT_OVERRIDE
feat: Refactor data source connection handling. 
feat: Add support for js-client-sdk style initialization.
END_COMMIT_OVERRIDE


packages/shared/sdk-client/src/polling/Requestor.ts
```diff
@@ -18,8 +18,6 @@ export class LDRequestError extends Error implements HttpErrorResponse {
 /**
  * Note: The requestor is implemented independently from polling such that it can be used to
  * make a one-off request.
- *
- * @internal
```
kinyoklion, line 22: To allow usage outside the common package.

packages/sdk/browser/jest.config.js
```diff
@@ -6,5 +6,6 @@ export default {
   transform: {
     '^.+\\.tsx?$': ['ts-jest', { useESM: true, tsconfig: 'tsconfig.json' }],
   },
-  testPathIgnorePatterns: ['./dist'],
+  testPathIgnorePatterns: ['./dist', './src'],
+  testMatch: ["**.test.ts"]
```
kinyoklion, line 10: The shared test code didn't contain tests, which made it angry.

packages/sdk/browser/package.json
```diff
@@ -27,7 +27,7 @@
   ],
   "scripts": {
     "clean": "rimraf dist",
-    "build": "rollup -c rollup.config.js",
+    "build": "tsc --noEmit && rollup -c rollup.config.js",
```
kinyoklion, line 30: Make sure TS errors actually fail the build.

packages/sdk/browser/src/BrowserDataManager.ts
```diff
@@ -0,0 +1,136 @@
...
+    if (await this.flagManager.loadCached(context)) {
+      this.debugLog('Identify - Flags loaded from cache. Continuing to initialize via a poll.');
+    }
+    const plainContextString = JSON.stringify(Context.toLDContext(context));
+    const requestor = this.getRequestor(plainContextString);
+
+    // TODO: Handle wait for network results in a meaningful way. SDK-707
+
+    try {
+      const payload = await requestor.requestPayload();
```
kinyoklion, line 66: The current SDK only does this once. We may want to consider if we want to at least do it twice.

packages/sdk/browser/src/BrowserDataManager.ts
```diff
@@ -0,0 +1,136 @@
...
+    _identifyOptions?: LDIdentifyOptions,
+  ): Promise<void> {
+    this.context = context;
+    if (await this.flagManager.loadCached(context)) {
+      this.debugLog('Identify - Flags loaded from cache. Continuing to initialize via a poll.');
+    }
+    const plainContextString = JSON.stringify(Context.toLDContext(context));
+    const requestor = this.getRequestor(plainContextString);
+
+    // TODO: Handle wait for network results in a meaningful way. SDK-707
```
kinyoklion, line 63: Maybe it just isn't an option for this SDK?

packages/sdk/browser/src/BrowserDataManager.ts
```diff
@@ -0,0 +1,136 @@
...
+    if (this.config.useReport) {
+      method = 'REPORT';
+      headers['content-type'] = 'application/json';
+      body = plainContextString; // context is in body for REPORT
+    }
+
+    const uri = getPollingUri(this.config.serviceEndpoints, path, parameters);
+    return new Requestor(this.platform.requests, uri, headers, method, body);
+  }
+  // TODO: Automatically start streaming if event handlers are registered.
```
kinyoklion, line 135: Made a ticket for this. Was originally in scope for this ticket, but it already ended up being too large.

packages/shared/sdk-client/src/DataManager.ts
```diff
@@ -0,0 +1,194 @@
...
+import { DataSourcePaths, StreamingProcessor } from './streaming';
+import { DeleteFlag, Flags, PatchFlag } from './types';
+
+export interface DataManager {
+  identify(
+    identifyResolve: () => void,
+    identifyReject: (err: Error) => void,
+    context: Context,
+    identifyOptions?: LDIdentifyOptions,
+  ): Promise<void>;
```
tanderson-ld, line 28: What is the returned promise for?  Consider moving comments that relate to interfaced functionality from LDClientImpl to this file.

packages/shared/sdk-client/src/IdentityProcessor.ts
```diff
@@ -0,0 +1,30 @@
...
+    identifyResolve: any,
+    identifyReject: any,
+  ) => void;
+  createPollingProcessor: (
+    context: LDContext,
+    checkedContext: Context,
+    identifyResolve: any,
+    identifyReject: any,
+  ) => void;
+  context: LDContext;
```
tanderson-ld, line 25: Is this LDContext still required?  I think this was for serialization into URLs/requests so that we didn't re-serialize the checkedContext.

packages/shared/sdk-client/src/IdentityProcessor.ts
```diff
@@ -0,0 +1,30 @@
+import { Context, LDContext, LDLogger } from '@launchdarkly/js-sdk-common';
+import { LDStreamProcessor } from '@launchdarkly/js-sdk-common/dist/api/subsystem';
+
+import ConnectionMode from './api/ConnectionMode';
+import DefaultFlagManager from './flag-manager/FlagManager';
+
+export type IdentifyProcessor = (params: {
+  waitForNetworkResults: boolean;
```
tanderson-ld, line 8: Could LDIdentifyOptions be used here?

packages/shared/sdk-client/src/context/addAutoEnv.ts
```diff
@@ -103,7 +103,11 @@ export const addDeviceInfo = async (platform: Platform) => {
   return undefined;
 };
 
-export const addAutoEnv = async (context: LDContext, platform: Platform, config: Configuration) => {
+export const addAutoEnv = async (
+  context: LDContext,
+  platform: Platform,
+  config: ConfigurationImpl,
```
tanderson-ld, line 109: Should this be Configuration type?  Also some other spots where Impl is being passed around.  Maybe the interface was created for testing injection?

packages/shared/sdk-client/src/flag-manager/FlagManager.ts
```diff
@@ -6,13 +6,23 @@ import { DefaultFlagStore } from './FlagStore';
...
 
+export interface FlagManager {
+  get(key: string): ItemDescriptor | undefined;
+  getAll(): { [key: string]: ItemDescriptor };
+  init(context: Context, newFlags: { [key: string]: ItemDescriptor }): Promise<void>;
+  upsert(context: Context, key: string, item: ItemDescriptor): Promise<boolean>;
+  loadCached(context: Context): Promise<boolean>;
+  on(callback: FlagsChangeCallback): void;
+  off(callback: FlagsChangeCallback): void;
+}
```
tanderson-ld, line 17: Consider moving comments from class to interface.

packages/shared/sdk-client/src/DataManager.ts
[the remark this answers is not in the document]
kinyoklion: I don't follow.

packages/shared/sdk-client/src/DataManager.ts
```diff
@@ -0,0 +1,194 @@
...
+}
+
+export interface DataManagerFactory {
+  (
+    flagManager: FlagManager,
+    configuration: Configuration,
+    baseHeaders: LDHeaders,
+    emitter: LDEmitter,
```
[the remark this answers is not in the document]
kinyoklion: The emitter is not error only, but all emissions. Though this does only use it in that domain. It doesn't need to. I could make an emit error function if that makes more sense.

packages/shared/sdk-client/src/IdentityProcessor.ts
kinyoklion: I think it could be removed. Maybe I can remove this whole thing now.

packages/shared/sdk-client/src/context/addAutoEnv.ts
kinyoklion: This is a stupid effect of vscode renaming, but it doesn't really matter. The purpose of the interface is to not use a concrete implementation outside of the package that defined the concrete implementation. So leaf-nodes can use configuraton, but not need the implementation expose.

I will still update them to be configuration though, as that is what I actually want.

packages/shared/sdk-client/src/DataManager.ts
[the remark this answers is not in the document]
kinyoklion: There isn't a directly related interface, but I can document these, but I am not sure how to make it non-redundant. Like "identifyResolve resolves the identify promise."

So, which part is confusing, and I can document that.

packages/shared/sdk-client/src/IdentityProcessor.ts
kinyoklion: It can be removed.

packages/shared/sdk-client/src/events/createEventProcessor.ts
```diff
@@ -17,7 +16,7 @@ const createEventProcessor = (
       baseHeaders,
       undefined,
       diagnosticsManager,
-      start,
+      false,
```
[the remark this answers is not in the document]
kinyoklion: Yes. It makes it easier to control from the specific SDK implementations. RN conditionally starts it and the Browser SDK always starts it.

packages/shared/sdk-client/src/IdentityProcessor.ts
kinyoklion: Removing this file.

packages/shared/sdk-client/src/DataManager.ts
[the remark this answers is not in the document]
kinyoklion: Maybe the name of the function itself needs to be different.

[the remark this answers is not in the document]
tanderson-ld: After discussion on Slack, seems like comments should be sufficient.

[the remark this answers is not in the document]
kinyoklion: Added an additional developer note.

tanderson-ld (approved):

PR #592, opened 2024-09-24
feat: Browser-SDK Automatically start streaming based on event handlers.

kinyoklion:


packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -105,15 +105,13 @@ export default class LDClientImpl implements LDClient {
       this.diagnosticsManager,
     );
     this.emitter = new LDEmitter();
-    this.emitter.on('change', (c: LDContext, changedKeys: string[]) => {
```
kinyoklion, line 108: I've left the default err catcher for now, but we may need to remove it in the future.

The old emitter had this functionality:
```
  emitter.maybeReportError = function(error) {
    if (!error) {
      return;
    }
    if (listeningTo('error')) {
      this.emit('error', error);
    } else {
      (logger || console).error(error.message);
    }
  };
```
Basically if you were listening for errors, then you get to handle them. If you were not listening to errors, then it would log them.

Which is potentially useful functionality to have, but out of the scope of this PR.

tanderson-ld (approved):

PR #594, opened 2024-09-26
fix: Multi-kind context containing only 1 kind conveted incorrectly.

kinyoklion:


packages/shared/common/__tests__/ContextFilter.test.ts
```diff
@@ -430,4 +448,20 @@ describe('when handling mult-kind contexts', () => {
     const uf = new ContextFilter(false, [new AttributeReference('name', true)]);
     expect(uf.filter(orgAndUserContext)).toEqual(orgAndUserGlobalNamePrivate);
   });
+
+  it('should produce event with valid single context', () => {
+    const uf = new ContextFilter(false, []);
+    expect(uf.filter(multiWithSingleContext)).toEqual({
+      kind: 'user',
```
kinyoklion, line 455: This attribute would have been missing.

tanderson-ld (approved):

PR #599, opened 2024-09-27
fix: Ensure client logger is always wrapped in a safe logger.

kinyoklion:


packages/shared/sdk-client/src/configuration/Configuration.ts
```diff
@@ -21,9 +22,6 @@ export interface LDClientInternalOptions extends internal.LDInternalOptions {
 
 export interface Configuration {
   readonly logger: LDLogger;
-  readonly baseUri: string;
```
kinyoklion, line 24: This change isn't related to the logger, but code should ALWAYS use the service endpoints, so we shouldn't be exposing these.

packages/shared/sdk-client/src/configuration/Configuration.ts
```diff
@@ -161,6 +168,8 @@ export default class ConfigurationImpl implements Configuration {
           } else {
             errors.push(OptionMessages.wrongOptionType(k, validator.getType(), typeof v));
           }
+        } else if (k === 'logger') {
```
kinyoklion, line 171: We need a logger before doing any of these other config things, so it is assigned earlier. Then we still let the user know if their logger is itself broken.

keelerm84 (approved):

PR #601, opened 2024-09-30
chore: Remove shared mocks package.

kinyoklion:
The dependencies between the mocks, and the unpublished nature of the package, made maintenance complex. Additionally the mocks themselves cause the tests to be harder to maintain and understand.

This PR changes all test dependencies to be in their own project. If we need shared test code, then we could make shared code and not a package.

The main motivator for removal is the circular dependency relationship between the mocks and common. Updating common to support ESM is made more difficult, but needs to be done.

Overall this makes things much simpler at the cost of a couple hundred lines of total duplication. (As indicated by the PR additions versus subtractions.) There was code that was shared that was actually only used in singular test locations.

Fixes SDK-729.


packages/sdk/cloudflare/jsr.json
```diff
@@ -3,15 +3,7 @@
...
-      "LICENSE",
-      "README.md",
-      "package.json",
-      "jsr.json",
-      "src/**/*.ts"
-    ],
-    "exclude": [
-      "src/**/*.test.ts"
-    ]
+    "include": ["LICENSE", "README.md", "package.json", "jsr.json", "src/**/*.ts"],
```
kinyoklion, line 6: Unrelated lint change.

keelerm84 (approved):

PR #603, opened 2024-10-02
feat: Add ESM support for common and sdk-client.

kinyoklion:
This PR uses TSC to build both an ESM and CJS version of common and sdk-client.

I was going to use rollup for this, but removing rollup provides more debuggable code for non-browser cases. The browser build can do the minification.


packages/sdk/browser/src/BrowserClient.ts
```diff
@@ -10,11 +10,11 @@ import {
   LDClientImpl,
   LDContext,
   LDEmitter,
+  LDEmitterEventName,
   LDHeaders,
+  LDIdentifyOptions,
   Platform,
 } from '@launchdarkly/js-client-sdk-common';
-import { LDIdentifyOptions } from '@launchdarkly/js-client-sdk-common/dist/api/LDIdentifyOptions';
```
kinyoklion, line 16: Building the new way helps find these bad imports.

PR #605, opened 2024-10-02
feat: Add support for hooks.

kinyoklion:


packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -239,14 +243,25 @@ export default class LDClientImpl implements LDClient {
     );
     this.logger.debug(`Identifying ${JSON.stringify(this.checkedContext)}`);
 
+    const afterIdentify = this.hookRunner.identify(context, identifyOptions?.timeout);
```
tanderson-ld, line 246: Mustn't this be invoked before all possible returns such as on [line 235](https://github.com/launchdarkly/js-core/pull/605/files#diff-5ab8d43c9c238bf0ca76fae92d46229147658df4640f421264a1f57eebfdde79R235)?

packages/shared/sdk-client/src/api/integrations/Hooks.ts
```diff
@@ -0,0 +1,155 @@
...
+export interface IdentifySeriesData {
+  readonly [index: string]: unknown;
+}
+
+/**
+ * The result applies to a single identify operation. An operation may complete
+ * with an error and then later complete successfully. Only the first completion
+ * will be executed in the evaluation series.
+ */
+export type IdentifyResult = 'completed' | 'error';
```
tanderson-ld, line 64: What do you think of this being an interface with a status field for future expandability?

packages/shared/sdk-client/src/LDClientImpl.ts
kinyoklion: The early return is for an invalid context and no identification happens or changes. I am not sure about it.

Though, this would be a good place for an error hook.

packages/shared/sdk-client/src/api/integrations/Hooks.ts
kinyoklion: Probably not a bad idea.

tanderson-ld (approved):

PR #607, opened 2024-10-04
feat: Add visibility handling to allow proactive event flushing.

kinyoklion:
Additionally use the `keep-alive` setting of fetch to ensure delivery of events even when the page may be closing.
Also better encapsulating browser APIs to ensure that the SDK can operate smoothly in service workers and browser extensions.
Also set browser appropriate flush interval.

Currently no testing for the state detector because of: https://github.com/jestjs/jest/issues/10025

I think we need to switch to a babel based Jest setup to get things working better. So far Jest has not been a pleasant experience for ESM.


packages/sdk/browser/src/options.ts
```diff
@@ -6,6 +6,8 @@ import {
   TypeValidators,
 } from '@launchdarkly/js-client-sdk-common';
 
+const DEFAULT_FLUSH_INTERVAL_SECONDS = 2;
```
kinyoklion, line 9: Realized this wasn't in as I was testing things out.

packages/sdk/browser/__tests__/goals/GoalTracker.test.ts
```diff
@@ -78,7 +78,7 @@ it('should add click event listener for click goals', () => {
 
   new GoalTracker(goals, mockOnEvent);
 
-  expect(document.addEventListener).toHaveBeenCalledWith('click', expect.any(Function));
```
kinyoklion, line 81: Consolidating adding event handlers resulted in extra parameters for options.

tanderson-ld (approved):

PR #612, opened 2024-10-09
chore: Prepare for browser client releases.

kinyoklion:


packages/sdk/browser/contract-tests/entity/src/ClientEntity.ts
```diff
@@ -1,4 +1,4 @@
-import { init, LDClient, LDLogger, LDOptions } from '@launchdarkly/js-client-sdk';
+import { initialize, LDClient, LDLogger, LDOptions } from '@launchdarkly/js-client-sdk';
```
kinyoklion, line 1: Changed the name to match existing package.

It means client/server are not consistent, so we may want to add init to the client, or add initialize to the server, then mark as deprecated.

packages/sdk/browser/src/BrowserClient.ts
```diff
@@ -31,16 +31,19 @@ import BrowserPlatform from './platform/BrowserPlatform';
  * Applications should configure the client at page load time and reuse the same instance.
  *
  * For more information, see the [SDK Reference Guide](https://docs.launchdarkly.com/sdk/client-side/javascript).
- *
- * @ignore Implementation Note: We are not supporting dynamically setting the connection mode on the LDClient.
```
kinyoklion, line 35: This caused the entire type not to be in the docs. Moved these implementation notes.

packages/sdk/browser/contract-tests/entity/src/main.ts
```diff
@@ -2,8 +2,6 @@
 import './style.css';
 import TestHarnessWebSocket from './TestHarnessWebSocket';
 
-// const client = init('618959580d89aa15579acf1d', AutoEnvAttributes.Enabled);
```
kinyoklion, line 5: Forgot this at some point. Ok though client-side IDs are not secret.

tanderson-ld (approved):

PR #613, opened 2024-10-09
fix: Fix base64 encoding of unicode characters.

kinyoklion:


packages/sdk/react-native/src/polyfills/btoa.ts
```diff
@@ -1,15 +1,10 @@
+/* eslint-disable no-bitwise */
 import { fromByteArray } from 'base64-js';
 
-function convertToByteArray(s: string) {
```
kinyoklion, line 3: This code would only work for ASCII basically.

packages/sdk/react-native/__tests__/platform/PlatformEncoding.test.ts
```diff
@@ -0,0 +1,13 @@
+import PlatformEncoding from '../../src/platform/PlatformEncoding';
```
kinyoklion, line 1: Copied the tests from the browser implementation.

packages/sdk/react-native/src/polyfills/toUtf8Array.ts
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable no-plusplus */
```
kinyoklion, line 1: There are a few packages we could use instead, but all have more things in them than we need. The closure library seems like a reasonable source for an implementation.

tanderson-ld (approved):

PR #618, opened 2024-10-10
chore: Float browser contract test dependency.

kinyoklion:


packages/sdk/browser/contract-tests/entity/package.json
```diff
@@ -11,7 +11,7 @@
     "prettier": "prettier --write '**/*.@(js|ts|tsx|json|css)' --ignore-path ../../../../.prettierignore"
   },
   "dependencies": {
-    "@launchdarkly/js-client-sdk": "0.0.0"
+    "@launchdarkly/js-client-sdk": "*"
```
kinyoklion, line 14: Floating the dep here causes it to get it correctly from the workspace. Versus a version that needs updated each time to avoid it fetching from npm.

tanderson-ld (approved):

PR #619, opened 2024-10-10
feat: Add AI SDK for Server-Side JavaScript.

InTheCloudDan:


packages/sdk/ai/README.md
```diff
@@ -0,0 +1,5 @@
+# LaunchDarkly Server-Side SDK for Node.js - AI integration
```
mmrj, line 1: I'm curious about the naming here -- 

@InTheCloudDan from our call, my understanding was you were starting to think about this as "LaunchDarkly AI SDK for Node.js" or "LaunchDarkly Node.js (AI) SDK" rather than "LaunchDarkly Node.js (server-side) SDK - AI integration."

Please keep me in the loop on how you'd like to position this, because it's going to make a pretty big difference in the docs information architecture. thanks!

InTheCloudDan: @mmrj great catch thanks, updated!

package.json
```diff
@@ -23,7 +23,8 @@
     "packages/store/node-server-sdk-dynamodb",
     "packages/telemetry/node-server-sdk-otel",
     "packages/tooling/jest",
-    "packages/sdk/browser"
+    "packages/sdk/browser",
+    "packages/sdk/ai"
```
kinyoklion, line 27: We should be really sure we want just "ai". Versus including node in the name for example.

Though there is the possibility we could make this work with more than just the node server. Something to think about eventually. It just requires a few methods, so we could probably make it just depend on those methods. Not important for now.

packages/sdk/ai/src/index.ts
```diff
@@ -0,0 +1,93 @@
...
+   * Retrieves and processes a prompt template based on the provided key, LaunchDarkly context, and variables.
+   *
+   * @param key - A unique identifier for the prompt template. This key is used to fetch the correct prompt from storage or configuration.
+   * @param context - The LaunchDarkly context object that contains relevant information about the current environment, user, or session. This context may influence how the prompt is processed or personalized.
+   * @param variables - A map of key-value pairs representing dynamic variables to be injected into the prompt template. The keys correspond to placeholders within the template, and the values are the corresponding replacements.
+   * @param defaultValue - A fallback value to be used if the prompt template associated with the key is not found or if any errors occur during processing.
+   *
+   * @returns The processed prompt after all variables have been substituted in the stored prompt template. If the prompt cannot be retrieved or processed, the `defaultValue` is returned.
+   *
+   * @example
```
kinyoklion, line 36: Examples should just be in triple backtick blocks. They can be tagged with the language, but not required. 

https://github.com/launchdarkly/js-core/blob/a986478ed8e39d0f529ca6adec0a09b484421390/packages/shared/common/src/api/context/LDContextMeta.ts#L34

Rendered example.
https://launchdarkly.github.io/js-core/packages/shared/common/docs/interfaces/LDContextMeta.html

packages/sdk/ai/src/types.ts
```diff
@@ -0,0 +1,50 @@
+export interface TokenMetrics {
```
kinyoklion, line 1: I would put the types you want to export in `/api` and then export * from API. For instance it looks like LDAIConfig isn't exported currently, which would be inconvenient for users who want to make a config and assign it to a variable.
I would put your internal types anywhere else. Generally not a fan of a file called 'types'.

packages/sdk/ai/src/index.ts
```diff
@@ -0,0 +1,95 @@
+import Mustache from 'mustache';
+
+import { LDClient, LDContext } from '@launchdarkly/node-server-sdk';
+
+import { LDAIConfigTracker } from './LDAIConfigTracker';
+import { LDAIConfig } from './api/config';
+
+export class AIClient {
```
mmrj, line 8: Can you say more about why this isn't prefaced with `LD`? The [Python equivalent](https://github.com/launchdarkly/python-server-sdk-ai/blob/main/ldai/client.py#L10) is `LDAIClient` but I'm not sure what goes into all the naming decisions across languages.

kinyoklion: It is now.

kinyoklion:
I would like to get the bulk of this in now. It is enough to do a pre-release. Additional PRs will contain fixes and tests.

kinyoklion (approved):

PR #620, opened 2024-10-10
feat: Apply private property naming standard. Mangle browser private properties.

kinyoklion:
1. Update private properties and methods to start with a leading underscore.
2. Update browser terser config to mangle properties that start with an underscore that are not `_meta`. (It would be nice to have a more elegant way to do this, but this is what we have.)
3. Add linting rules for typescript naming conventions.
5. Fix lint issues from rules.

Conventions

Private properties and methods start with _ and CamelCase case names.
Public methods and properties use CamelCase and do not start with an underscore.
Public static properties use PascalCase and do not start with an underscore. (This is negotiable, but required fewer changes.)

sdk-764

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -47,6 +47,38 @@ module.exports = {
     'jest/no-focused-tests': 'error',
     'jest/no-identical-title': 'error',
     'jest/valid-expect': 'error',
+    'no-underscore-dangle': ['error', { allowAfterThis: true }],
+    '@typescript-eslint/naming-convention': [
+      'error',
+      {
+        selector: ['method'],
+        format: ['camelCase'],
+        leadingUnderscore: 'forbid',
+      },
+      {
+        selector: ['method'],
+        format: ['camelCase'],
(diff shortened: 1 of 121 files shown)


.eslintrc.js
```diff
@@ -47,6 +47,38 @@ module.exports = {
     'jest/no-focused-tests': 'error',
     'jest/no-identical-title': 'error',
     'jest/valid-expect': 'error',
+    'no-underscore-dangle': ['error', { allowAfterThis: true }],
+    '@typescript-eslint/naming-convention': [
```
kinyoklion, line 51: Basically these go broad and then specific. Forbid underscore!... unless private and then require. And again for the properties.

packages/sdk/browser/README.md
```diff
@@ -9,6 +9,7 @@
 -->
 
 # ⛔️⛔️⛔️⛔️
+
```
kinyoklion, line 12: Lint.

tanderson-ld (approved):

PR #622, opened 2024-10-10
fix: Do not mangle _meta.

kinyoklion:


packages/sdk/browser/rollup.config.js
```diff
@@ -21,6 +21,15 @@ const getSharedConfig = (format, file) => ({
   },
 });
 
+const terserOpts = {
+  mangle: {
+    properties: {
+      regex: /^_/,
+      reserved: ['_meta']
```
kinyoklion, line 28: Better than a regex even.

tanderson-ld: Consider adding some comments to help prevent a developer from overlooking the importance of this line.  I could see someone seeing _meta here and not realizing it is LD system related.

tanderson-ld (approved):

PR #624, opened 2024-10-11
feat: adds ping stream support

tanderson-ld:
**Requirements**

-I have tested this against browser and mobile.

**Related issues**

SDK-198

Added ping pathing
Passing Requestor into StreamingProcessor that now uses it when a ping event is received.
Commonized requestor creation logic and using that helper function in various data managers.


packages/sdk/browser/__tests__/BrowserDataManager.test.ts
tanderson-ld, line 1: For reviewers: Updating paths to avoid confusing future developers about usage of /msdk

packages/sdk/browser/src/BrowserClient.ts
```diff
@@ -140,6 +140,9 @@ export class BrowserClient extends LDClientImpl implements LDClient {
             pathReport(_encoding: Encoding, _plainContextString: string): string {
               return `/sdk/evalx/${clientSideId}/context`;
             },
+            pathPing(_encoding: Encoding, _plainContextString: string): string {
+              throw new Error('Ping for polling unsupported.');
+            },
```
tanderson-ld, line 145: For reviewers: Thoughts? There are other spots like this. I considered just using the /ping or /mping endpoint. Seems like if this case does get encountered, a request failure may be better than throwing the error. Seems like the odds of that error handling is better.

packages/sdk/browser/src/BrowserDataManager.ts
```diff
@@ -196,35 +207,29 @@ export default class BrowserDataManager extends BaseDataManager {
     const rawContext = Context.toLDContext(context)!;
 
     this.updateProcessor?.close();
-    this.createStreamingProcessor(rawContext, context, identifyResolve, identifyReject);
 
-    this.updateProcessor!.start();
-  }
-
-  private _getRequestor(plainContextString: string): Requestor {
```
tanderson-ld, line 204: For reviewers: _getRequestor got commonized to be a helper method in the Requestor file, the generated diff is confusing.

packages/shared/sdk-client/src/datasource/Requestor.ts
```diff
@@ -47,3 +56,41 @@ export default class Requestor {
     throw new LDRequestError(`Unexpected status code: ${status}`, status);
   }
 }
+
```
tanderson-ld, line 59: For reviewers: This is the combination of the handful of requestor making implementations.

packages/shared/sdk-client/src/polling/PollingProcessor.ts
```diff
@@ -115,7 +80,7 @@ export default class PollingProcessor implements subsystem.LDStreamProcessor {
     }
 
     const elapsed = Date.now() - startTime;
-    const sleepFor = Math.max(this._pollInterval * 1000 - elapsed, 0);
+    const sleepFor = Math.max(this._pollIntervalSeconds * 1000 - elapsed, 0);
```
tanderson-ld, line 83: For reviewers: Just a rename for clarity.

packages/shared/sdk-client/src/DataManager.ts
```diff
@@ -133,6 +124,7 @@ export abstract class BaseDataManager implements DataManager {
         this._dataSourceEventHandler.handlePollingError(err);
         identifyReject?.(err);
       },
+      this.logger,
```
tanderson-ld, line 127: For reviewers: Please verify if this logger was not intentionally being omitted up to this point.

packages/sdk/browser/src/BrowserClient.ts
kinyoklion: Encountering this situation would be a coding error on our part, so I think this is the correct action. Effectively an assert which will get us close.

I do think we need to add a comment here. That explains as such.
```
// Implementation note: If you are seeing this error it means ...
```

packages/sdk/browser/src/BrowserDataManager.ts
```diff
@@ -196,35 +207,29 @@ export default class BrowserDataManager extends BaseDataManager {
...
-      parameters.push({ key: 'withReasons', value: 'true' });
-    }
-    if (this._secureModeHash) {
-      parameters.push({ key: 'h', value: this._secureModeHash });
-    }
+    const plainContextString = JSON.stringify(Context.toLDContext(context));
+    const pollingRequestor = makeRequestor(
+      plainContextString,
+      this.config.serviceEndpoints,
+      this.getPollingPaths(), // note: this is the polling path because the requestor is only used to make polling requests.
```
kinyoklion, line 215: This comment is a bit confusing because it feels like it implies the requestor could be used to make a streaming request and we are choosing not to.

packages/sdk/react-native/__tests__/MobileDataManager.test.ts
```diff
@@ -132,6 +132,9 @@ describe('given a MobileDataManager with mocked dependencies', () => {
         pathReport(_encoding: Encoding, _plainContextString: string): string {
           return `/msdk/evalx/context`;
         },
+        pathPing(_encoding: Encoding, _plainContextString: string): string {
+          throw new Error('Ping for polling unsupported.');
```
kinyoklion, line 136: Should make sure these are uniformly commented.

packages/shared/sdk-client/src/streaming/StreamingProcessor.ts
```diff
@@ -178,6 +178,41 @@ class StreamingProcessor implements subsystem.LDStreamProcessor {
         }
       });
     });
+
+    // here we set up a listener that will poll when ping is received
+    eventSource.addEventListener('ping', async () => {
```
kinyoklion, line 183: It seems a little strange that this handler is here instead of with the other listeners.

packages/sdk/browser/src/BrowserDataManager.ts
tanderson-ld: Ahh, I see what you're saying.  I'll remove the comment.

kinyoklion (approved):

PR #628, opened 2024-10-17
fix: Prerequisites should not trigger hooks.

kinyoklion:


packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -357,7 +357,7 @@ export default class LDClientImpl implements LDClient {
     }
 
     prerequisites?.forEach((prereqKey) => {
-      this.variation(prereqKey, undefined);
+      this._variationInternal(prereqKey, undefined, this._eventFactoryDefault);
```
kinyoklion, line 360: Variation internal will not execute hooks. We do not use the detailed evaluation factory. (Equivalent to calling `variation`.)

tanderson-ld (approved):

PR #632, opened 2024-10-23
feat: add initial SDK library boilerplate and basic svelte LD SDK

nosnibor89:
**Requirements**

- [x] I have followed the repository's [pull request submission guidelines](../blob/main/CONTRIBUTING.md#submitting-pull-requests) (TBH `yarn run contract-tests` failed for me even in `main`)

**Related issues**

No issue

Introducing the new `@launchdarkly/svelte-client-sdk` package. Some of the details included in this PR are

1.  Svelte Library Boilerplate 
2. Basic Svelte SDK functionality:
2.1 `LDProvider` component
2.2 `LDFlag` component
2.3 Svelte-compatible LD instance (exposes API to work with feature flags)

I don't know what to write here.

**Additional context**

This is the first of a series of PRs. Some of the following PR should be about
1. Adding Documentation for @launchdarkly/svelte-client-sdk
2. Adding Example project that uses `@launchdarkly/svelte-client-sdk`


kinyoklion:
Hello @nosnibor89,

Thank you for the contribution.

Those contract tests aren't relevant to client-side SDKs, but they should all work correctly (and do run in CI and passed for your PR). When developing locally running `yarn build` in the root of the repo should get everything into a functioning state. (Builds all packages, and everything in the workspace will use local packages.) The contract tests do run a go binary downloaded from github, so that could cause problems for some configurations.

I won't be able to review this deeply this week, but hopefully I can next week.

Thank you,
Ryan

packages/sdk/svelte/__tests__/lib/client/SvelteLDClient.test.ts
```diff
@@ -0,0 +1,244 @@
+import * as LDClient from 'launchdarkly-js-client-sdk';
```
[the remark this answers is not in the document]
kinyoklion: `@launchdarkly/js-client-sdk` If you use the same version number as is in the repo it will use the version from the workspace.

`@launchdarkly/js-client-sdk/compat` provide a compatibility interface to be mostly compatible with the old package.

packages/sdk/svelte/package.json
```diff
@@ -0,0 +1,89 @@
...
+    "svelte"
+  ],
+  "type": "module",
+  "svelte": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "svelte": "./dist/index.js",
+      "default": "./dist/index.js"
```
kinyoklion, line 23: Sounds like svelte doesn't support CJS? That makes things very pleasant.

packages/sdk/svelte/package.json
```diff
@@ -0,0 +1,89 @@
...
+    "prepublishOnly": "npm run package",
+    "lint": "eslint . --ext .ts,.tsx",
+    "prettier": "prettier --write '**/*.@(js|ts|tsx|json|css)' --ignore-path ../../../.prettierignore",
+    "check": "yarn prettier && yarn lint && yarn build && yarn test",
+    "test": "playwright test",
+    "test:unit": "vitest",
+    "test:unit-ui": "vitest --ui"
+  },
+  "peerDependencies": {
+    "@launchdarkly/js-client-sdk-common": "^1.1.4",
```
kinyoklion, line 46: This is an interesting mix. 

`launchdarkly-js-client-sdk` doesn't use `@launchdarkly/js-client-sdk-common`

But the new one does, but theoretically we would only want to use one or the other.

packages/sdk/svelte/src/lib/client/SvelteLDClient.ts
```diff
@@ -0,0 +1,105 @@
+import { initialize } from 'launchdarkly-js-client-sdk';
+import type {
+  LDClient,
+  LDFlagSet,
+  LDFlagValue,
+  LDContext as NodeLDContext,
```
kinyoklion, line 6: Maybe some non-node name. BrowserLDContext?

packages/sdk/svelte/src/lib/client/SvelteLDClient.ts
```diff
@@ -0,0 +1,105 @@
...
+    isClientInitialized(jsSdk);
+    return jsSdk.identify(context);
+  }
+
+  /**
+   * Watches a flag for changes.
+   * @param {string} flagKey - The key of the flag to watch.
+   * @returns {Readable<LDFlagsValue>} A readable store of the flag value.
+   */
+  const watch = (flagKey: string): Readable<LDFlagsValue> =>
```
kinyoklion, line 80: I am not sure how it needs to be handled with Svelte, but when a flag is accessed we have to make sure an event is sent to launchdarkly. This is accomplished with a variation/variation detail call.

Part of the SDKs job is providing access to flags, but the other part is driving product features.

Flag insights and experimentation, for instance, require those events to function.

For react we use a `Proxy`: https://github.com/launchdarkly/react-client-sdk/blob/21c2402f6d591caf61764ba35263ff1e4cb18ae7/src/getFlagsProxy.ts#L61

Ideally though we would listen to specific flags.

packages/sdk/svelte/src/lib/client/SvelteLDClient.ts
```diff
@@ -0,0 +1,105 @@
...
+    derived<Writable<LDFlags>, LDFlagsValue>(flagsWritable, ($flags) => $flags[flagKey]);
+
+  /**
+   * Checks if a flag is on.
+   * @param {string} flagKey - The key of the flag to check.
+   * @returns {boolean} True if the flag is on, false otherwise.
+   */
+  const isOn = (flagKey: string): boolean => {
```
kinyoklion, line 88: So, we could have something like this that was useValue, but we wouldn't want a pure boolean result.

Ideally each flag access takes a default, and that default goes to variation, and then that can be included in events to drive product features. So ideally we would focus the interface on individual flag access.

On means something different though. Lets imagine a flag called backgroundColor, when it is on it serves 5% of users blue, and the rest yellow, but when it is off it serves white. 

On/Off is part of the reason for an evaluation, but not derived from its value.

nosnibor89: @kinyoklion Jum! 🤔 

I think I get the part where we don't want to restrict ourselves to booleans only. I could refactor this to be something like: 
```
 const isValue = (flagKey: string, value: LDFlagsValue = true): boolean => {
    isClientInitialized(jsSdk);
    const currentFlags = get(flagsWritable);
    return currentFlags[flagKey] === value;
  };
```
so later I could do `isValue('backgroundColor', true)`, assuming "backgroundColor" is a boolean flag (if I understand [this](https://docs.launchdarkly.com/sdk/concepts/flag-types) correctly)


However, I'm not sure I understand how to handle default values. Can you clarify that a little bit better for me?

packages/sdk/svelte/package.json
nosnibor89: Not by default. I think there are ways to set it up if we need it though. 

Sounds like we don't need it here.

packages/sdk/svelte/__tests__/lib/client/SvelteLDClient.test.ts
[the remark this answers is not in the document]
nosnibor89: Ok. I think I refactored this to use `@launchdarkly/js-client-sdk` now. Please take a peek.

packages/sdk/svelte/package.json
nosnibor89: Yeah. I think I added this in an attempt to keep other SDK standards. I just removed it.

packages/sdk/svelte/src/lib/client/SvelteLDClient.ts
nosnibor89: Now that we use `@launchdarkly/js-client-sdk` , I no longer need this.

packages/sdk/svelte/src/lib/client/SvelteLDClient.ts
nosnibor89: That makes sense. I just added a similar function `toFlagsProxy` to track flag reads and use the `client.variation(prop, currentValue);` behind the scenes.

packages/sdk/svelte/src/lib/client/SvelteLDClient.ts
kinyoklion: Anywhere we "use" a flag, we need to use the `variation` method and provide it with a default value.

So, for example:

```
function useFlag<TFlag extends LDFlagsValue>(flagKey: string, defaultValue: TFlag): TFlag => {
 isClientInitialized(jsSdk);
  // Ideally we would add some real type safety here.
 return jsSdk.variation(flagKey, defaultValue) as TFlag;
}
```

```
if(useFlag('my-flag', false)) { // the generic should cause this to be a boolean

}
```
Though you could be explicit as well:
```
if(useFlag<boolean>('my-flag', false)) { // the generic should cause this to be a boolean

}
```

Or use other types:
```
if(useFlag('my-string-flag', 'cheese') === 'potato')) {
 // I am a potato.
}
```

kinyoklion: We do have individually type safe methods as well, so those could be directly proxies as an alternative to a generic. Or in addition to the generic.

`useBooleanFlag`

For example. The thing is that in the implementation we can check if the value was actually a boolean. If it is not a boolean, then we use the default value.

We also report this back to launchdarkly so that there are analytics that the wrong type is being used.

nosnibor89: Thanks @kinyoklion . I used your suggestion, please take a look whenever you have a chance.

packages/sdk/svelte/package.json
```diff
@@ -0,0 +1,89 @@
+{
+  "name": "@launchdarkly/svelte-client-sdk",
+  "version": "1.0.0",
```
kinyoklion, line 3:
```suggestion
  "version": "0.1.0",
```
We will want to do some prereleases to validate things. So we should drop this to 0.1.0.

kinyoklion (approved):

nosnibor89:
Awesome @kinyoklion . Thanks for improving and merging this one. 

Next, we'll create PR for "example" project and improve documentation.

PR #637, opened 2024-10-25
feat: Add a module for increased backward compatibility.

kinyoklion:


packages/sdk/browser/package.json
```diff
@@ -17,9 +17,16 @@
     "sdk"
   ],
   "exports": {
-    "types": "./dist/src/index.d.ts",
-    "require": "./dist/index.cjs.js",
-    "import": "./dist/index.es.js"
+    ".": {
+      "types": "./dist/src/index.d.ts",
+      "require": "./dist/index.cjs.js",
```
kinyoklion, line 22: I may need to make this fancier when we verify CJS/ESM package inclusion. Given the problems we had in other packages.

tanderson-ld (approved):

PR #640, opened 2024-10-28
feat: Refine CJS/ESM build configuration for browser SDK.

kinyoklion:
The goals here are to:
- Make the build simpler.
- Make the build "modern".

Currently we have worked around some build issues in other packages. Mainly with node module resolution. Despite the settings in the package.json node wants certain file extensions for certain types including .cts for the CJS type definitions.

So far tsup seems to be the simplest way to get all things with expected names without a multi-step process. Or without having to make a hack package.json that just has the modules type.

Please review https://github.com/launchdarkly/js-core/pull/637 first.


packages/sdk/browser/tsconfig.json
```diff
@@ -8,28 +8,27 @@
...
     "noImplicitOverride": true,
     "resolveJsonModule": true,
-    // Uses "." so it can load package.json.
-    "rootDir": ".",
+    "rootDir": "src",
     "outDir": "dist",
     "skipLibCheck": true,
-    // enables importers to jump to source
-    "sourceMap": true,
+    "sourceMap": false,
```
kinyoklion, line 14: For the bundle-size task I will add some conditional configuration to a dev build and source maps.

packages/sdk/browser/tsup.config.ts
```diff
@@ -0,0 +1,29 @@
...
+        // data model.
+        reserved: ['_meta'],
+      },
+    },
+  },
+  format: ['esm', 'cjs'],
+  splitting: false,
+  sourcemap: false,
+  clean: true,
+  noExternal: ['@launchdarkly/js-sdk-common', '@launchdarkly/js-client-sdk-common'],
```
kinyoklion, line 26: So, the one thing that wasn't great about tsup is that it doesn't bundle dependencies unless they are dev dependencies. Which sounds fine, but then you need to figure out how to handle the typescript for those packages. 

It has a method to bundle the typescript into .d.ts files as well, but that very much didn't work.

So this instructs it to not treat these as external. It does mean that you get these installed for a production build even though they are not then used. But that has been the precedent.

If the bunding of the typing worked, then that would be a reasonable alternative.

packages/sdk/browser/tsconfig.json
kinyoklion: The esbuild support actually means it can generate a meta data file, which can be used for the analysis instead. So no debug builds for now.

tanderson-ld (approved):

packages/sdk/browser/tsup.config.ts
```diff
@@ -0,0 +1,27 @@
...
+  dts: true,
+  metafile: true,
+  esbuildOptions(opts) {
+    // This would normally be `^_(?!meta|_)`, but go doesn't support negative look-ahead assertions,
+    // so we need to craft something that works without it.
+    // So start of line followed by a character that isn't followed by m or underscore, but we
+    // want other things that do start with m, so we need to progressively handle more characters
+    // of meta with exclusions.
+    // eslint-disable-next-line no-param-reassign
```
kinyoklion, line 24: The interface works by modifying the input options, which means re-assigning.

kinyoklion:
I need to appease tsdoc. Which may be some minor changes.

PR #649, opened 2024-10-31
chore:  Revert - Fix incompatibility with @types/node release.

kinyoklion:
Reverts launchdarkly/js-core#646


kinyoklion:
I'll commit override the other one.

kinyoklion:
Tagged as a chore as it was not released, so we can avoid including changes for either PR.

tanderson-ld (approved):

PR #654, opened 2024-11-01
chore: No package-size comments on forks.

kinyoklion:


actions/package-size/action.yml
```diff
@@ -39,7 +41,7 @@ runs:
         body-includes: '${{ inputs.package_name }} size report'
 
     - name: Create comment
-      if: steps.fc.outputs.comment-id == ''
+      if: steps.fc.outputs.comment-id == '' && github.event.pull_request.head.repo.full_name == github.repository
```
kinyoklion, line 44: For forks the repo name will have a different owner and will not match.

kinyoklion: I am not 100% sure if a composite can access this, so we will hopefully find out from the CI run.

tanderson-ld (approved):

PR #658, opened 2024-11-04
chore: checking documentation

tanderson-ld:


packages/shared/common/src/api/context/LDContextCommon.ts
```diff
@@ -1,5 +1,8 @@
 import { LDContextMeta } from './LDContextMeta';
 
+/**
+ * Common attributes shared by the various derived context types.
```
kinyoklion, line 4: I think we could drop derived here. It may imply something and I think the meaning is basically the same without it.

kinyoklion (approved):

PR #668, opened 2024-11-06
fix: Do not include _ldMeta in returned config.

kinyoklion:
The primary purpose of this was to add unit tests, but it ended up with a bug fix, so that is the PR title.


packages/sdk/server-ai/__tests__/LDAIConfigTrackerImpl.test.ts
```diff
@@ -0,0 +1,270 @@
...
+});
+
+it('tracks Bedrock conversation with error response', () => {
+  const tracker = new LDAIConfigTrackerImpl(mockLdClient, configKey, versionKey, testContext);
+
+  const response = {
+    $metadata: { httpStatusCode: 400 },
+  };
+
+  // TODO: We may want a track failure.
```
kinyoklion, line 199: This can be added in another PR once we have an answer.

tanderson-ld (approved):

PR #676, opened 2024-11-08
feat: Add stack trace parsing.

kinyoklion:
Review after: https://github.com/launchdarkly/js-core/pull/675

### DIFF
--- a/packages/telemetry/browser-telemetry/src/stack/StackParser.ts
+++ b/packages/telemetry/browser-telemetry/src/stack/StackParser.ts
@@ -4,6 +4,16 @@ import { StackFrame } from '../api/stack/StackFrame';
 import { StackTrace } from '../api/stack/StackTrace';
 import { ParsedStackOptions } from '../options';
 
+/**
+ * In the browser we will not always be able to determine the source file that code originates
+ * from. When you access a route it may just return HTML with embedded source, or just source,
+ * in which case there may not be a file name.
+ *
+ * There will also be cases where there is no source file, such as when running with various
+ * dev servers.
+ *
+ * In these situations we use this constant in place of the file name.
+ */
 const INDEX_SPECIFIER = '(index)';
 
 /**
@@ -18,12 +28,19 @@ export function processUrlToFileName(input: string, origin: string): string {
   let cleaned = input;
   if (input.startsWith(origin)) {
     cleaned = input.slice(origin.length);
+    // If the input is a single `/` then it would get removed and we would
+    // be left with an empty string. That empty string would get replaced with
+    // the INDEX_SPECIFIER. In cases where a `/` remains, either singular
+    // or at the end of a path, then we will append the index specifier.
+    // For instance the route `/test/` would ultimately be `test/(index)`.
     if (cleaned.startsWith('/')) {
       cleaned = cleaned.slice(1);
     }
+
     if (cleaned === '') {
-      cleaned = INDEX_SPECIFIER;
+      return INDEX_SPECIFIER;
     }
+
     if (cleaned.endsWith('/')) {
       cleaned += INDEX_SPECIFIER;
     }
@@ -62,7 +79,27 @@ export function trimSourceLine(options: TrimOptions, line: string, column: numbe
 }
 
 /**
+ * Given a context get trimmed source lines within the specified range.
+ *
+ * The context is a list of source code lines, this function returns a subset of
+ * lines which have been trimmed.
+ *
+ * If an error is on a specific line of source code we want to be able to get
+ * lines before and after that line. This is done relative to the originating
+ * line of source.
+ *
+ * If you wanted to get 3 lines before the origin line, then this function would
+ * need to be called with `start: originLine - 3, end: originLine`.
+ *
+ * If the `start` would underflow the context, then the start is set to 0.
+ * If the `end` would overflow the context, then the end is set to the context
+ * length.
+ *
  * Exported for testing.
+ *
+ * @param start The inclusive start index.
+ * @param end The exclusive end index.
+ * @param trimmer Method which will trim individual lines.
  */
 export function getLines(
   start: number,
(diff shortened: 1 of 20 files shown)


keelerm84 (approved):

packages/telemetry/browser-telemetry/src/stack/StackParser.ts
```diff
@@ -0,0 +1,154 @@
...
+ * @returns The output file name.
+ */
+export function processUrlToFileName(input: string, origin: string): string {
+  let cleaned = input;
+  if (input.startsWith(origin)) {
+    cleaned = input.slice(origin.length);
+    if (cleaned.startsWith('/')) {
+      cleaned = cleaned.slice(1);
+    }
+    if (cleaned === '') {
```
cwaldren-ld, line 24: nit: this and the next case are mutually exclusive right? I think an if/else if could be clearer when reading it

packages/telemetry/browser-telemetry/src/stack/StackParser.ts
```diff
@@ -0,0 +1,154 @@
...
+}
+
+export interface TrimOptions {
+  /**
+   * The maximum length of the trimmed line.
+   */
+  maxLength: number;
+
+  /**
+   * If the line needs trimmed, then this is the number of character to retain before the
```
cwaldren-ld, line 41:
```suggestion
   * If the line needs to be trimmed, then this is the number of character to retain before the
```

packages/telemetry/browser-telemetry/src/stack/StackParser.ts
```diff
@@ -0,0 +1,154 @@
...
+  }
+  const captureStart = Math.max(0, column - options.beforeColumnCharacters);
+  const captureEnd = Math.min(line.length, captureStart + options.maxLength);
+  return line.slice(captureStart, captureEnd);
+}
+
+/**
+ * Exported for testing.
+ */
+export function getLines(
```
cwaldren-ld, line 67: nit: could still use a comment

packages/telemetry/browser-telemetry/src/stack/StackParser.ts
```diff
@@ -0,0 +1,154 @@
...
+  if (adjustedStart < adjustedEnd) {
+    return context.slice(adjustedStart, adjustedEnd).map(trimmer);
+  }
+  return [];
+}
+
+/**
+ * Exported for testing.
+ */
+export function getSrcLines(
```
cwaldren-ld, line 84: ditto

packages/telemetry/browser-telemetry/src/stack/StackParser.ts
```diff
@@ -0,0 +1,154 @@
...
+} {
+  const { context } = inFrame;
+  // It should be present, but we don't want to trust that it is.
+  if (!context) {
+    return {};
+  }
+  const { maxLineLength } = options.source;
+  const beforeColumnCharacters = Math.floor(maxLineLength / 2);
+
+  // The before and after lines will not be precise while we use TraceKit.
```
cwaldren-ld, line 105: I feel this comment might not have enough context to be useful, while still pointing out a deficiency that should potentially be fixed. Could be more useful as a link to a ticket (or just more context on why the lines won't be precise with that lib?)

packages/telemetry/browser-telemetry/src/stack/StackParser.ts
```diff
@@ -0,0 +1,154 @@
...
+  // By forking it we should be able to achieve a more optimal result.
+
+  // Trimmer for non-origin lines. Starts at column 0.
```
cwaldren-ld, line 108: what does 'non-origin' mean?

packages/telemetry/browser-telemetry/src/stack/StackParser.ts
```diff
@@ -0,0 +1,154 @@
+import { computeStackTrace } from 'tracekit';
+
+import { StackFrame } from '../api/stack/StackFrame';
+import { StackTrace } from '../api/stack/StackTrace';
+import { ParsedStackOptions } from '../options';
+
+const INDEX_SPECIFIER = '(index)';
```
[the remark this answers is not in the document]
kinyoklion: Yes, it is a placeholder for the root of the document.

packages/telemetry/browser-telemetry/src/stack/StackParser.ts
```diff
@@ -0,0 +1,154 @@
...
+    srcBefore: getLines(origin - options.source.beforeLines, origin, context, trimmer),
+    srcLine: trimSourceLine(
+      {
+        maxLength: maxLineLength,
+        beforeColumnCharacters,
+      },
+      context[origin],
+      inFrame.column || 0,
+    ),
+    srcAfter: getLines(origin + 1, origin + 1 + options.source.afterLines, context, trimmer),
```
cwaldren-ld, line 130: I think the details of this should be noted in comments (like why `origin + 1`, etc.?) I suppose what I'm really asking for is an explanation of what this whole function is trying to achieve

cwaldren-ld (approved):

packages/telemetry/browser-telemetry/src/stack/StackParser.ts
kinyoklion: So, from a code execution perspective there would be an else here. They are mutually exclusive. But they are also not required to be mutually exclusive. Basically these are independent checks that can be made, and using else doesn't change the logic, but it does change the runtime.

If it was when it is empty, then this other operation must not be done, then I would certainly want the else.

My general concern is that independent conditions we can add more and move them around.

So technically cleaned being empty could be first, then it could have an else, and inside the else the other two if statements. 

I'll consider.

packages/telemetry/browser-telemetry/src/stack/StackParser.ts
kinyoklion: So the origin line we need it to at least start from where the error is.
Theoretically the origin could be, and often will be, very long.

Non-origin lines we don't have a starting point that is critical, so we start from the beginning of the line.

PR #684, opened 2024-11-13
feat: Add support for getting the logger from server-side SDKs.

kinyoklion:


packages/shared/sdk-server/src/api/LDClient.ts
```diff
@@ -461,4 +462,11 @@ export interface LDClient {
    * @param Hook The hook to add.
    */
   addHook?(hook: Hook): void;
+
+  /**
+   * Get the logger used by this LDClient instance.
+   *
+   * For all platforms that support logging the logger should be present.
+   */
+  get logger(): LDLogger | undefined;
```
kinyoklion, line 471: Using an accessor here instead of a readonly property which should make this a little more flexible and also less likely to be assigned over in JS.

keelerm84 (approved):

PR #702, opened 2024-12-04
chore: Standardize typescript version.

kinyoklion:


packages/shared/sdk-server/src/store/sortDataSet.ts
```diff
@@ -51,7 +51,7 @@ function topologicalSort(
 
   while (unvisitedItems.size > 0) {
     // Visit the next item, the order we visit doesn't matter.
-    const key = unvisitedItems.values().next().value;
+    const key = unvisitedItems.values().next().value!;
```
kinyoklion, line 54: New typescript thinks this can be undefined, but it cannot.

PR #708, opened 2024-12-10
feat: Add underlying session replay support. (Not enabled)

kinyoklion:
This adds components for implementing session replay. These components are not going to be exported or enabled until a later time.


packages/telemetry/browser-telemetry/src/collectors/rrweb/ContinuousReplay.ts
```diff
@@ -0,0 +1,89 @@
+import type { eventWithTime } from '@rrweb/types';
+import * as rrweb from 'rrweb';
+
+import { Recorder } from '../../api';
+import { Collector } from '../../api/Collector';
+import { ContinuousCapture } from './SessionReplayOptions';
+
+export default class ContinuousReplay implements Collector {
+  private _telemetry?: Recorder;
+  // TODO: Use a better buffer. (SDK-972)
```
tanderson-ld, line 10: I saw in the one test a drop ratio of 1.5 and had a question about that, but perhaps that will be invalidated by this story's work.

packages/telemetry/browser-telemetry/src/collectors/rrweb/ContinuousReplay.ts
```diff
@@ -0,0 +1,89 @@
+import type { eventWithTime } from '@rrweb/types';
+import * as rrweb from 'rrweb';
+
+import { Recorder } from '../../api';
+import { Collector } from '../../api/Collector';
+import { ContinuousCapture } from './SessionReplayOptions';
+
+export default class ContinuousReplay implements Collector {
+  private _telemetry?: Recorder;
```
tanderson-ld, line 9: Is _telemetry too specific of a name?  I see _destination in other collectors.

packages/telemetry/browser-telemetry/src/collectors/rrweb/RollingBuffer.ts
```diff
@@ -0,0 +1,66 @@
...
+  toArray(): any[] {
+    const asArray: any[] = [];
+    const size = this._buffers.reduce((acc: number, item: EventBuffer) => {
+      if (item.isPopulated()) {
+        return acc + 1;
+      }
+      return acc;
+    }, 0);
+
+    for (let index = this._headPointer; index < this._headPointer + size; index += 1) {
```
tanderson-ld, line 53: Just double checking to make sure this loop will work.  Is it true that all populated items are contiguous and starting at headPointer?  If so, it could be possible to have the populated check in the loop condition and then no need to calculate size above.  One less iteration on the buffers.  `index < this_headPointer + this._buffers.length` would probably be needed as well.

packages/telemetry/browser-telemetry/src/collectors/rrweb/RollingReplay.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+    this._index += 1;
+    this._restartCapture();
+  }
+
+  private _restartCapture() {
+    this._buffer.reset();
+    this._stopper?.();
+    this._startCapture();
+  }
```
tanderson-ld, line 57: I see start and restart.  Why does stop not exist?  Is the usage "never ending"?

packages/telemetry/browser-telemetry/src/collectors/rrweb/SessionReplay.ts
```diff
@@ -0,0 +1,59 @@
...
+  constructor(options?: SessionReplayOptions) {
+    const captureOptions = options?.capture;
+    if (isContinuousCapture(captureOptions)) {
+      this.impl = new ContinuousReplay(captureOptions);
+    } else if (isRollingCapture(captureOptions)) {
+      this.impl = new RollingReplay(captureOptions);
+    } else {
+      this.impl = new ContinuousReplay({
+        type: 'continuous',
+        intervalMs: 5 * 1000,
```
tanderson-ld, line 35: How is this number picked?  Or how should it be picked once closer to shipping?

packages/telemetry/browser-telemetry/src/collectors/rrweb/ContinuousReplay.ts
kinyoklion: This wouldn't address any ratios, just the performance of the buffer. The tests are specifically designed to cause bad things.

packages/telemetry/browser-telemetry/src/collectors/rrweb/RollingReplay.ts
kinyoklion: Yeah, it is "rolling" all the time. Which allows you to at any point capture whatever is in the buffer, and then reset the buffer.

packages/telemetry/browser-telemetry/src/collectors/rrweb/SessionReplay.ts
kinyoklion: It is totally arbitrary right now. This mode hopefully is removed for production.

packages/telemetry/browser-telemetry/src/collectors/rrweb/RollingReplay.ts
kinyoklion: It does stop when unregistered though. Currently there isn't any signal that would be stopping.

tanderson-ld (approved):

PR #711, opened 2024-12-11
fix: Remove bundled dependency from package.json.

kinyoklion:


packages/sdk/cloudflare/package.json
```diff
@@ -76,8 +76,5 @@
     "ts-jest": "^29.1.0",
     "typedoc": "0.25.0",
     "typescript": "5.1.6"
-  },
```
kinyoklion, line 79: This didn't do what it was intended to do, but did cause a warning on install.
Basically this dependency is bundled with rollup, but this package.json indicates that the package would contain a .tgz file with this dependency in it.

tanderson-ld (approved):

PR #723, opened 2025-01-03
feat: Add Fastly Edge SDK

ldhenry:
**Requirements**

This PR adds a new Fastly Compute SDK. This SDK must be used in conjunction with our upcoming Fastly KV integration. The SDK is essentially the same as the other edge integrations (Akamai, Cloudflare, Vercel), with the following changes:
- `node:events` is not compatible with Fastly's runtime. As a result, we cannot use `@launchdarkly/sdk-server-edge`. Instead, I copied the contents of `@launchdarkly/sdk-server-edge` into the fastly package and replaced `createCallbacks.ts` with an empty implementation. 
- Enabled `sendEvents` by default.
- Added the `eventsUri` option to allow for sending events to a custom endpoint.
- Added a new optional `eventsBackendName` option. A [Fastly Backend](https://developer.fastly.com/reference/api/services/backend/) configured to `https://events.launchdarkly.com` is required for sending events. The default value is `launchdarkly`. This option is passed to [Fastly's customized fetch()](https://js-compute-reference-docs.edgecompute.app/docs/globals/fetch#explicit-backends).

I added an example app that demonstrates using the SDK to evaluate a feature flag edge to control the static image served.

I published an beta version to npm [here](https://www.npmjs.com/package/@launchdarkly/fastly-server-sdk?activeTab=readme).


packages/sdk/fastly/__tests__/api/LDClient.test.ts
```diff
@@ -0,0 +1,46 @@
+import { internal } from '@launchdarkly/js-server-sdk-common';
```
ldhenry, line 1: All tests in this directory were copied over from `@launchdarkly/sdk-server-edge`

packages/sdk/fastly/__tests__/createBasicPlatform.ts
```diff
@@ -0,0 +1,59 @@
+import { PlatformData, SdkData } from '@launchdarkly/js-server-sdk-common';
```
ldhenry, line 1: This was copied from `@launchdarkly/sdk-server-edge`

packages/sdk/fastly/__tests__/createPlatformInfo.test.ts
```diff
@@ -0,0 +1,17 @@
+import createPlatformInfo from '../src/createPlatformInfo';
+
+describe('Fastly Platform Info', () => {
+  it('platformData shows correct information', () => {
+    const platformData = createPlatformInfo();
+
+    expect(platformData.platformData()).toEqual({
+      name: 'Fastly Edge',
```
ldhenry, line 8: I'm thinking about changing this to Fastly Compute based on Fastly's branding [here](https://www.fastly.com/documentation/guides/compute/). What do you think, @kinyoklion?

mmrj:
(docs ticket is at DOCS-973 and currently scheduled for week of 1/6)

packages/sdk/fastly/package.json
```diff
@@ -0,0 +1,63 @@
...
+      }
+    }
+  },
+  "main": "../dist/index.cjs",
+  "types": "./dist/index.d.cts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "tsup && ../../../scripts/replace-version.sh .",
```
ldhenry, line 38: @kinyoklion, let me know if there is a better way to get the version in `src.createPlatformInfo.ts`

packages/sdk/fastly/src/api/EdgeFeatureStore.ts
```diff
@@ -0,0 +1,116 @@
+import type {
```
ldhenry, line 1: This was copied from `@launchdarkly/sdk-server-edge`

packages/sdk/fastly/src/api/LDClient.ts
```diff
@@ -0,0 +1,24 @@
+import { Info, internal, LDClientImpl, LDOptions } from '@launchdarkly/js-server-sdk-common';
```
ldhenry, line 1: This was copied from `@launchdarkly/sdk-server-edge`

packages/sdk/fastly/src/api/createCallbacks.ts
```diff
@@ -0,0 +1,9 @@
+const createCallbacks = () => ({
```
ldhenry, line 1: This is the new empty implementation to get around the lack of `node:events`

packages/sdk/fastly/src/api/createOptions.ts
```diff
@@ -0,0 +1,17 @@
+import { BasicLogger, LDOptions } from '@launchdarkly/js-server-sdk-common';
```
ldhenry, line 1: This was copied from @launchdarkly/sdk-server-edge

packages/sdk/fastly/src/index.ts
```diff
@@ -0,0 +1,48 @@
...
+    logger,
+    ...ldOptions,
+  };
+
+  validateOptions(sdkKey, finalOptions);
+  return new LDClient(
+    sdkKey,
+    createPlatformInfo(),
+    finalOptions,
+    eventsBackendName || DEFAULT_EVENTS_BACKEND_NAME,
```
ldhenry, line 46: I chose to pass `eventsBackendName` as a new parameter instead of changing the `options` type used by LDClient. Let me know if you think this is the wrong approach.

packages/sdk/fastly/src/platform/crypto/cryptoJSHasher.ts
```diff
@@ -0,0 +1,49 @@
+import CryptoJS from 'crypto-js';
```
ldhenry, line 1: Everything in this directory was copied from @launchdarkly/sdk-server-edge

packages/sdk/fastly/src/platform/requests.ts
```diff
@@ -0,0 +1,34 @@
...
+export default class EdgeRequests implements Requests {
+  eventsBackend: string;
+
+  constructor(eventsBackend: string) {
+    this.eventsBackend = eventsBackend;
+  }
+
+  fetch(url: string, options: Options = {}): Promise<Response> {
+    // @ts-ignore
+    return fetch(url, { ...options, backend: this.eventsBackend });
```
ldhenry, line 20: This is the main differentiator between the other edge SDKs. We are required to pass a [fastly-specific `backend` parameter](https://js-compute-reference-docs.edgecompute.app/docs/globals/fetch#explicit-backends).

.sdk_metadata.json
```diff
@@ -28,6 +28,15 @@
         "tag-prefix": "cloudflare-server-sdk-"
       }
     },
+    "Fastly": {
```
kinyoklion, line 31: I am not sure about this key being capitalized.

packages/sdk/fastly/__tests__/createPlatformInfo.test.ts
kinyoklion: That works for me.

packages/sdk/fastly/src/api/createCallbacks.ts
kinyoklion: So I think we can make a task, not blocking this, which removes the dependency from the common package. 

That would then allow for the eventually removal of most of the code for this package.

packages/sdk/fastly/src/index.ts
kinyoklion: The LDClient implementation is your implementation, so it could take the fastly options without any modification to the base options. You would just need to move how you destructure it.

That said I am indifferent to it being its own parameter.

packages/sdk/fastly/src/platform/crypto/cryptoJSHasher.ts
kinyoklion: Do we know if we need crypto-js, or does fastly compute have its own suitable alternatives?

packages/sdk/fastly/example/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "node": "^16 || >=18"
+  },
+  "devDependencies": {
+    "@fastly/cli": "^10.17.1",
+    "rimraf": "^6.0.1",
+    "typescript": "^5.7.2"
+  },
+  "dependencies": {
+    "@fastly/js-compute": "^3.28.0",
+    "@launchdarkly/fastly-server-sdk": "workspace:^"
```
kinyoklion, line 15: We may want to consider using the version number here and having release-please update it. That makes it more like a real example.

```
      "extra-files": [
        {
          "type": "json",
          "path": "example/package.json",
          "jsonpath": "$.dependencies['@launchdarkly/cloudflare-server-sdk']"
        },
...
      ]
```

kinyoklion: For example then the user can copy it out of the repo and it should work. The yarn workspace will link it locally during development.

kinyoklion (dismissed):
I've not looked through everything yet, but I did do an initial pass with a few items. Sorry for the delay.

ldhenry:
> I've not looked through everything yet, but I did do an initial pass with a few items. Sorry for the delay.

No problem and I appreciate your feedback. I'll take another pass next week.

packages/sdk/fastly/package.json
ldhenry: Thanks. I just updated this and also added a jsr.json file.

packages/sdk/fastly/src/platform/crypto/cryptoJSHasher.ts
ldhenry: It seems like they have explicitly tested crypto-js as noted [here](https://www.fastly.com/documentation/guides/compute/javascript/#using-dependencies). With that being said, I'll do a test to see if we can get away without it.

packages/sdk/fastly/example/package.json
ldhenry: Good idea. I just updated it.

packages/sdk/fastly/src/platform/crypto/cryptoJSHasher.ts
ldhenry: I did some digging and it looks like Fastly exposes [SimpleCrypto.](https://js-compute-reference-docs.edgecompute.app/docs/globals/crypto/subtle) I started to implement our Hasher with this but hit a roadblock because [SimpleCrypto's digest function returns a promise](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest#algorithm).

I'll stick with crypto-js for now unless you are aware of a workaround.

packages/sdk/fastly/src/index.ts
ldhenry: I decided to move the `eventsBackendName` into the client options.

packages/sdk/fastly/src/api/createOptions.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const defaultOptions: LDOptions = {
+  stream: false,
+  sendEvents: true,
```
ldhenry, line 5: Unlike our other Edge SDKs, this will send events by default.

.sdk_metadata.json
```diff
@@ -28,6 +28,15 @@
         "tag-prefix": "cloudflare-server-sdk-"
       }
     },
+    "fastly": {
+      "name": "Fastly SDK",
+      "type": "edge",
+      "path": "packages/sdk/fastly",
+      "languages": ["JavaScript", "TypeScript"],
+      "releases": {
+        "tag-prefix": "faslty-server-sdk-"
```
keelerm84, line 37:
```suggestion
        "tag-prefix": "fastly-server-sdk-"
```

packages/sdk/fastly/README.md
```diff
@@ -0,0 +1,62 @@
+# LaunchDarkly SDK for Fastly
+
+The LaunchDarkly SDK for Fastly is designed primarily for use in [Fastly Compute Platform](https://www.fastly.com/documentation/guides/compute/). It follows the server-side LaunchDarkly model for multi-user contexts. It is not intended for use in desktop and embedded systems applications.
```
keelerm84, line 3: Does it support any other use case?

```suggestion
The LaunchDarkly SDK for Fastly is designed for use in [Fastly Compute Platform](https://www.fastly.com/documentation/guides/compute/). It follows the server-side LaunchDarkly model for multi-user contexts. It is not intended for use in desktop and embedded systems applications.
```

packages/sdk/fastly/__tests__/api/EdgeFeatureStore.test.ts
```diff
@@ -0,0 +1,125 @@
+import { AsyncStoreFacade, LDFeatureStore } from '@launchdarkly/js-server-sdk-common';
+
+import { EdgeFeatureStore } from '../../src/api/EdgeFeatureStore';
+import mockEdgeProvider from '../../src/utils/mockEdgeProvider';
+import * as testData from './testData.json';
+
+describe('EdgeFeatureStore', () => {
+  const sdkKey = 'sdkKey';
```
keelerm84, line 8: This is actually the client ID, right? Can we rename that to better reflect the expected key?

packages/sdk/fastly/__tests__/api/EdgeFeatureStore.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    featureStore = new EdgeFeatureStore(mockEdgeProvider, sdkKey, 'MockEdgeProvider', mockLogger);
+    asyncFeatureStore = new AsyncStoreFacade(featureStore);
+  });
+
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  describe('get', () => {
+    test('get flag', async () => {
```
keelerm84, line 31: From my understanding, we would ideally like these tests to read like sentences. Combining the nested `describe`s with a final `it` block should yield something descriptive.

Perhaps we should convert the test to something like `it('can retrieve valid flag')`.

packages/sdk/fastly/__tests__/api/EdgeFeatureStore.test.ts
```diff
@@ -0,0 +1,125 @@
...
+      const flag = await asyncFeatureStore.get({ namespace: 'features' }, 'testFlag1');
+
+      expect(mockGet).toHaveBeenCalledWith(kvKey);
+      expect(flag).toMatchObject(testData.flags.testFlag1);
+    });
+
+    test('invalid flag key', async () => {
```
keelerm84, line 38: Then this becomes 
```suggestion
    it('returns undefined for an invalid flag key', async () => {
```

And so on for these tests.

packages/sdk/fastly/__tests__/api/EdgeFeatureStore.test.ts
```diff
@@ -0,0 +1,125 @@
...
+      const flag = await asyncFeatureStore.get({ namespace: 'features' }, 'invalid');
+
+      expect(flag).toBeUndefined();
```
keelerm84, line 41: We would still expect the mockGet call, like in other tests right? I think we should add that assertion back, but just confirm that it yields "undefined" as the final result.

packages/sdk/fastly/__tests__/api/LDClient.test.ts
```diff
@@ -0,0 +1,68 @@
...
+        polling: 'https://sdk.launchdarkly.com',
+        streaming: 'https://stream.launchdarkly.com',
+      },
+    });
+  });
+  it('uses custom eventsUri when specified', async () => {
+    const client = new LDClient('client-side-id', createBasicPlatform().info, {
+      sendEvents: true,
+      eventsBackendName: 'launchdarkly',
+      eventsUri: 'https://custom-base-uri.launchdarkly.com',
```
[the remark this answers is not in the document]
keelerm84: @kinyoklion is there a way to exclude that validation?

packages/sdk/fastly/__tests__/index.test.ts
```diff
@@ -0,0 +1,110 @@
...
+import { KVStore } from 'fastly:kv-store';
+
+import { LDClient } from '../src/api';
+import { init } from '../src/index';
+import * as testData from './utils/testData.json';
+
+// Tell Jest to use the manual mock
+jest.mock('fastly:kv-store');
+
+const sdkKey = 'test-sdk-key';
```
keelerm84, line 11: Same comment as before about this being the client ID and potentially renaming.

packages/sdk/fastly/__tests__/index.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    });
+    ldClient = init(sdkKey, mockKVStore);
+    await ldClient.waitForInitialization();
+  });
+
+  afterAll(() => {
+    ldClient.close();
+  });
+
+  describe('flags', () => {
```
keelerm84, line 41: Same note about test naming convention here as well.

packages/sdk/fastly/example/README.md
```diff
@@ -0,0 +1,74 @@
...
+yarn && yarn build
+```
+
+2. Replace `LAUNCHDARKLY_CLIENT_ID` in [src/index.ts](src/index.ts) with your LaunchDarkly SDK key.
+
+3. Create a new Fastly Compute service and KV Store:
+
+```shell
+# Login to Fastly CLI
+fastly login
```
keelerm84, line 43: This command doesn't exist for me. I had to make a token through the UI and then do `fastly profile create`.

packages/sdk/fastly/example/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+# Create a new Compute@Edge service
+fastly compute init
```
keelerm84, line 46: This step isn't strictly necessary since we have an existing `fastly.toml` file. We should also note that running this issues a warning about running it in a directory that isn't empty.

packages/sdk/fastly/example/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+# Create a new KV Store
+fastly kv-store create --name launchdarkly
```
keelerm84, line 49: Running this without having done the init before resulted in me first creating a service. I then had to link that up to the KV store through the Fastly UI.

packages/sdk/fastly/src/index.ts
```diff
@@ -0,0 +1,33 @@
...
+
+import { BasicLogger } from '@launchdarkly/js-server-sdk-common';
+
+import { EdgeFeatureStore, EdgeProvider, LDClient } from './api';
+import { DEFAULT_EVENTS_BACKEND_NAME } from './api/LDClient';
+import createPlatformInfo from './createPlatformInfo';
+import validateOptions, { FastlySDKOptions } from './utils/validateOptions';
+
+export const init = (
+  sdkKey: string,
```
keelerm84, line 12: Not sure if I've missed this elsewhere, but I think this probably should be clientId or something similar as well.

packages/sdk/fastly/src/createPlatformInfo.ts
```diff
@@ -0,0 +1,24 @@
...
+    return {
+      name: 'Fastly Compute',
+    };
+  }
+
+  sdkData(): SdkData {
+    return {
+      name,
+      version,
+      userAgentBase: 'FastlyEdgeSDK',
```
tanderson-ld, line 17: Do you want to distinguish server sdk from a future client sdk?

packages/sdk/fastly/src/utils/validateOptions.ts
```diff
@@ -0,0 +1,40 @@
...
+    throw new Error('You must configure the client with a client key');
+  }
+
+  if (!featureStore || typeof featureStore !== 'object' || !featureStore.get) {
+    throw new Error('You must configure the client with a feature store');
+  }
+
+  if (!logger) {
+    throw new Error('You must configure the client with a logger');
+  }
```
tanderson-ld, line 31: I think the logger is optional on most SDKs.
```
   * If you do not set this property, the SDK uses {@link basicLogger} with a
   * minimum level of `info`.
   */
  logger?: LDLogger;
```

packages/sdk/fastly/src/utils/validateOptions.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const validateOptions = (sdkKey: string, options: LDOptionsInternal) => {
+  const { eventsBackendName, featureStore, logger, sendEvents, ...rest } = options;
+  if (!sdkKey) {
+    throw new Error('You must configure the client with a client key');
+  }
+
+  if (!featureStore || typeof featureStore !== 'object' || !featureStore.get) {
+    throw new Error('You must configure the client with a feature store');
+  }
```
tanderson-ld, line 27: Here is the logic used by the type validators for the feature store.  https://github.com/launchdarkly/js-core/blob/main/packages/shared/common/src/validators.ts#L22

tanderson-ld (commented):
General question: why are there a handful of files that are copied from common and not used as dependencies?

kinyoklion:
> General question: why are there a handful of files that are copied from common and not used as dependencies?

@tanderson-ld This is because the common edge package has dependencies which are not suitable (and shouldn't have been included originally). 

Basically node:events is used in the common edge package and should not have been. The correct way to fix this is to remove that dependency and migrate it into the edge SDKs that can use it, or major version them and just remove it entirely. It isn't actually useful functionality in the edge.

ldhenry:
> > General question: why are there a handful of files that are copied from common and not used as dependencies?
> 
> @tanderson-ld This is because the common edge package has dependencies which are not suitable (and shouldn't have been included originally).
> 
> Basically node:events is used in the common edge package and should not have been. The correct way to fix this is to remove that dependency and migrate it into the edge SDKs that can use it, or major version them and just remove it entirely. It isn't actually useful functionality in the edge.

IMO a major version bump that removes node:events seems reasonable to me. Unfortunately I don't think I'll have the bandwidth for this before I go on parental leave. Do you think this is a blocker for the Fastly SDK?

packages/sdk/fastly/src/createPlatformInfo.ts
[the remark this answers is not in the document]
ldhenry: I think Edge is probably enough of a distinction here. I'm happy to change it but I don't see any future client SDK for Fastly (or other edge providers). I was going off of [Cloudflare](https://github.com/launchdarkly/js-core/blob/67bde4afbda9ff3bd005eea4fa1b45638843662d/packages/sdk/cloudflare/src/createPlatformInfo.ts#L17), and [Vercel](https://github.com/launchdarkly/js-core/blob/67bde4afbda9ff3bd005eea4fa1b45638843662d/packages/sdk/vercel/src/createPlatformInfo.ts#L14).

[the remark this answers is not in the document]
kinyoklion: Yeah, I think you can think of `Edge` as an equivalent level to `Client` or `Server`. It theoretically could be possible to make a 'Client' SDK for edge, but it wouldn't be very useful.

packages/sdk/fastly/src/utils/validateOptions.ts
ldhenry: We are always setting a logger [here](https://github.com/launchdarkly/js-core/blob/58c7cd25072fa5e7434d6c1a6ad802ae70894c4f/packages/sdk/fastly/src/index.ts#L16) if not specified so the additional validation in validateOptions is maybe a little redundant.

packages/sdk/fastly/src/utils/validateOptions.ts
ldhenry: Thanks. I started down this road and it seems promising. I have to sign off soon but I should have your suggestion in first thing tomorrow.

packages/sdk/fastly/example/src/index.ts
```diff
@@ -71,6 +71,32 @@ async function handleRequest(event: FetchEvent) {
...
+      return runEvaluationsInSeries(count - 1, results);
+    };
+
+    await runEvaluationsInSeries(35, flagValues);
+
+    return new Response(JSON.stringify(flagValues, undefined, 2), {
+      status: 200,
+      headers: new Headers({ 'Content-Type': 'application/json' }),
+    });
+  }
```
tanderson-ld, line 98: Was this for local testing?

packages/sdk/fastly/src/api/EdgeFeatureStore.ts
```diff
@@ -14,6 +14,7 @@ export interface EdgeProvider {
 
 export class EdgeFeatureStore implements LDFeatureStore {
   private readonly _rootKey: string;
+  private _kvData: string | null | undefined;
```
tanderson-ld, line 17: I believe only undefined is necessary for this situation.  Is there another API that has both null and undefined that is causing you to use both here?

packages/sdk/fastly/src/api/EdgeFeatureStore.ts
```diff
@@ -24,6 +25,18 @@ export class EdgeFeatureStore implements LDFeatureStore {
     this._rootKey = `LD-Env-${sdkKey}`;
   }
 
+  /**
+   * This function is used to lazy load the KV data from the edge provider. This is necessary because Fastly Compute
+   * has a limit of 32 backend requests (including requests to fetch the KV data).
+   * https://docs.fastly.com/products/compute-resource-limits
+   */
```
tanderson-ld, line 32: Do you want to describe in the readme or public APIs that data won't receive updates during the runtime of the fastly process/execution/request handler lifecycle?

(I think I have that right)

packages/sdk/fastly/example/src/index.ts
ldhenry: Yep good catch

packages/sdk/fastly/src/api/EdgeFeatureStore.ts
ldhenry: The Fastly KV fetch method returns `null` when the KV item does not exist and we have a `null` check in the `initialized` method. However, I don't think `undefined` is needed so I removed that.

packages/sdk/fastly/src/api/EdgeFeatureStore.ts
ldhenry: Thanks for the feedback. I just added a Usage Notes section to the README that contains this.

PR #725, opened 2025-01-07
chore: preparation work for FDv2 feature store and persistence support

tanderson-ld:
Splitting these changes up into multiple PRs to keep them manageable and because we can't merge directly to main as `applyChanges` is dependent on transaction support in the PersistentDataStore.

Existing tests are exercising applyChanges, but through init, upsert, and delete.  Plan to add applyChanges tests in ta/fdv2-temporary-holding after transaction support is added to PersistentDataStore in a subsequent PR as part of SDK-1029

### DIFF
--- a/packages/shared/sdk-server/src/api/subsystems/LDFeatureStore.ts
+++ b/packages/shared/sdk-server/src/api/subsystems/LDFeatureStore.ts
@@ -137,6 +137,13 @@ export interface LDFeatureStore {
    */
   upsert(kind: DataKind, data: LDKeyedFeatureStoreItem, callback: () => void): void;
 
+  applyChanges(
+    basis: boolean,
+    data: LDFeatureStoreDataStorage,
+    selector: String | undefined,
+    callback: () => void,
+  ): void;
+
   /**
    * Tests whether the store is initialized.
    *
--- a/packages/shared/sdk-server/src/api/subsystems/LDFeatureStore.ts
+++ b/packages/shared/sdk-server/src/api/subsystems/LDFeatureStore.ts
@@ -137,6 +137,16 @@ export interface LDFeatureStore {
    */
   upsert(kind: DataKind, data: LDKeyedFeatureStoreItem, callback: () => void): void;
 
+  /**
+   * @param basis If true, completely overwrites the current contents of the data store
+   * with the provided data.  If false, upserts the items in the provided data.  Upserts
+   * are made only if provided items have newer versions than existing items.
+   * @param data An object in which each key is the "namespace" of a collection (e.g. `"features"`) and
+   * the value is an object that maps keys to entities. The actual type of this parameter is
+   * `interfaces.FullDataSet<VersionedData>`.
+   * @param selector TODO
+   * @param callback Will be called after the changes are applied.
+   */
   applyChanges(
     basis: boolean,
     data: LDFeatureStoreDataStorage,
--- a/packages/shared/sdk-server/src/api/subsystems/LDFeatureStore.ts
+++ b/packages/shared/sdk-server/src/api/subsystems/LDFeatureStore.ts
@@ -138,13 +138,16 @@ export interface LDFeatureStore {
   upsert(kind: DataKind, data: LDKeyedFeatureStoreItem, callback: () => void): void;
 
   /**
+   * Applies the provided data onto the existing data, replacing all data or upserting depending
+   * on the basis parameter.
+   *
    * @param basis If true, completely overwrites the current contents of the data store
    * with the provided data.  If false, upserts the items in the provided data.  Upserts
    * are made only if provided items have newer versions than existing items.
    * @param data An object in which each key is the "namespace" of a collection (e.g. `"features"`) and
    * the value is an object that maps keys to entities. The actual type of this parameter is
    * `interfaces.FullDataSet<VersionedData>`.
-   * @param selector TODO
+   * @param selector opaque string that uniquely identifies the state that contains the changes
    * @param callback Will be called after the changes are applied.
    */
   applyChanges(
--- a/packages/shared/sdk-server/src/data_sources/DataSourceUpdates.ts
+++ b/packages/shared/sdk-server/src/data_sources/DataSourceUpdates.ts
@@ -67,14 +67,38 @@ export default class DataSourceUpdates implements LDDataSourceUpdates {
   ) {}
 
   init(allData: LDFeatureStoreDataStorage, callback: () => void): void {
+    this.applyChanges(true, allData, undefined, callback); // basis is true for init. selector is undefined for FDv1 init
+  }
+
+  upsert(kind: DataKind, data: LDKeyedFeatureStoreItem, callback: () => void): void {
+    this.applyChanges(
+      false, // basis is false for upserts
+      {
+        [kind.namespace]: {
+          [data.key]: data,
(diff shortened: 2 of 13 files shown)


packages/shared/sdk-server/src/data_sources/DataSourceUpdates.ts
```diff
@@ -101,55 +132,28 @@ export default class DataSourceUpdates implements LDDataSourceUpdates {
...
-        });
-
-        callback?.();
-      });
-    };
-    if (checkForChanges) {
-      this._featureStore.get(kind, key, doUpsert);
-    } else {
-      doUpsert();
-    }
```
tanderson-ld, line 152: For reviewers: upsert and init have been combined and the distinction is now driven by the bool `basis`.

packages/shared/sdk-server/src/data_sources/DataSourceUpdates.ts
```diff
@@ -101,55 +132,28 @@ export default class DataSourceUpdates implements LDDataSourceUpdates {
...
       this._featureStore.all(VersionedDataKinds.Features, (oldFlags) => {
         this._featureStore.all(VersionedDataKinds.Segments, (oldSegments) => {
-          const oldData = {
+          oldData = {
             [VersionedDataKinds.Features.namespace]: oldFlags,
             [VersionedDataKinds.Segments.namespace]: oldSegments,
           };
-          doInit(oldData);
         });
       });
```
tanderson-ld, line 153: For reviewers: These `all` calls feel a bit off, but without a refactoring of layer responsibility, I don't see a change that is worth the effort.  Thoughts?

packages/shared/sdk-server/src/store/InMemoryFeatureStore.ts
```diff
@@ -12,22 +12,6 @@ export default class InMemoryFeatureStore implements LDFeatureStore {
...
-    }
-    if (Object.hasOwnProperty.call(items, key)) {
-      const old = items[key];
-      if (!old || old.version < item.version) {
-        items[key] = item;
-      }
-    } else {
-      items[key] = item;
-    }
-  }
```
tanderson-ld, line 29: For reviewers: This logic has moved into apply changes.

packages/shared/sdk-server/src/store/PersistentDataStoreWrapper.ts
```diff
@@ -252,6 +252,27 @@ export default class PersistentDataStoreWrapper implements LDFeatureStore {
...
+    } else {
+      Object.entries(_data).forEach(([namespace, items]) => {
+        Object.keys(items || {}).forEach((key) => {
+          const item = items[key];
+          this.upsert({ namespace }, { key, ...item }, () => {});
+        });
+      });
+      _callback();
+    }
+  }
```
tanderson-ld, line 274: For reviewers: This code is not shippable (why this PR is going to a temp branch) due to the intermediate states potentially being invalid.

packages/shared/sdk-server/src/data_sources/DataSourceUpdates.ts
kinyoklion: I think something of this sort will generally be required. Though they could be combined potentially into some "get all for kinds", but I don't see a huge amount of value in that. 

With FDv2 they should at least always be in memory.

packages/shared/sdk-server/src/store/InMemoryFeatureStore.ts
```diff
@@ -53,19 +37,66 @@ export default class InMemoryFeatureStore implements LDFeatureStore {
...
+    selector: String | undefined, // TODO handle selector
+    callback: () => void,
+  ): void {
+    if (basis) {
+      this._initCalled = true;
+      this._allData = data;
+    } else {
+      Object.entries(data).forEach(([namespace, items]) => {
+        Object.keys(items || {}).forEach((key) => {
+          let existingItems = this._allData[namespace];
```
kinyoklion, line 82: We should make a shallow copy of the existing data, update the shallow data, and then update the reference to the new data. Otherwise partially applied visibility is available.

kinyoklion (changes requested):

packages/shared/sdk-server/src/store/InMemoryFeatureStore.ts
tanderson-ld: Good catch.

packages/shared/sdk-server/src/api/subsystems/LDFeatureStore.ts
```diff
@@ -137,6 +137,23 @@ export interface LDFeatureStore {
...
   upsert(kind: DataKind, data: LDKeyedFeatureStoreItem, callback: () => void): void;
 
+  /**
+   * @param basis If true, completely overwrites the current contents of the data store
+   * with the provided data.  If false, upserts the items in the provided data.  Upserts
+   * are made only if provided items have newer versions than existing items.
+   * @param data An object in which each key is the "namespace" of a collection (e.g. `"features"`) and
+   * the value is an object that maps keys to entities. The actual type of this parameter is
+   * `interfaces.FullDataSet<VersionedData>`.
+   * @param selector TODO
```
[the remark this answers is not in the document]
tanderson-ld: Replaced TODO with comment

packages/shared/sdk-server/src/api/subsystems/LDFeatureStore.ts
```diff
@@ -137,6 +137,23 @@ export interface LDFeatureStore {
    */
   upsert(kind: DataKind, data: LDKeyedFeatureStoreItem, callback: () => void): void;
 
+  /**
```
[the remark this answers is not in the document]
tanderson-ld: Done

packages/shared/sdk-server/src/store/InMemoryFeatureStore.ts
tanderson-ld: Done

packages/shared/sdk-server/src/store/InMemoryFeatureStore.ts
```diff
@@ -53,19 +37,74 @@ export default class InMemoryFeatureStore implements LDFeatureStore {
...
+        Object.keys(items || {}).forEach((key) => {
+          let existingItems = tempData[namespace];
+          if (!existingItems) {
+            existingItems = {};
+            tempData[namespace] = existingItems;
+          }
+          const item = items[key];
+          if (Object.hasOwnProperty.call(existingItems, key)) {
+            const old = existingItems[key];
+            if (!old || old.version < item.version) {
```
kinyoklion, line 96: I don't like this being on the FDv2 path. Version checking  is V1, V2 we should apply whatever we are told. Likely it won't actually cause a problem. I can approve this for now, but would like to note it in a task or something and consider coming back to it.

kinyoklion (approved):

packages/shared/sdk-server/src/store/InMemoryFeatureStore.ts
tanderson-ld: That's an interesting point.  I will add a story and note it.

tanderson-ld: Added TODO, done.

PR #729, opened 2025-01-14
feat: Vendor TraceKit

kinyoklion:
This PR vendors tracekit to alleviate some issues with ESM compatibility.
Currently the typescript conversion is minimal. It adds the types which are largely compatible and leaves most of it reasonably unchanged.

This PR does not port the tests from tracekit to this repository. A subsequent PR will do this.

BEGIN_COMMIT_OVERRIDE
feat: Vendor TraceKit
feat: Export browser-telemetry initialization method.
END_COMMIT_OVERRIDE

Jira: EMSR-14
Jira: EMSR-15

### DIFF
--- a/packages/telemetry/browser-telemetry/src/vendor/TraceKit.ts
+++ b/packages/telemetry/browser-telemetry/src/vendor/TraceKit.ts
@@ -1,484 +1,252 @@
-let TraceKit = {};
-
 /**
  * https://github.com/csnover/TraceKit
  * @license MIT
  * @namespace TraceKit
  */
-(function(window, undefined) {
-    if (!window) {
-        return;
+
+/**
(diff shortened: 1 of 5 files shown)


packages/telemetry/browser-telemetry/src/vendor/TraceKit.ts
kinyoklion, line 1: Aside from type annotations the only functional changes in this file are additional conversion checks to match the expected types in the original code.

Additionally the module support has been removed, and assigning to window has been removed, and a function was added to get the tracekit instance versus exporting it.

The problem with modules was that they are imported immutably, so we could not do things like adjust the context lines which required assignment to an import.

tanderson-ld (approved):

PR #733, opened 2025-01-15
feat: Add support for breadcrumb filtering.

kinyoklion:
This PR adds the ability to specify a filter that runs when adding any breadcrumb. The filter can return a modified breadcrumb, or return undefined to filter the breadcrumb entirely.

This could be used to redact any information which a customer doesn't want in a bread crumb.

### DIFF
--- a/packages/telemetry/browser-telemetry/src/api/Options.ts
+++ b/packages/telemetry/browser-telemetry/src/api/Options.ts
@@ -1,3 +1,4 @@
+import { Breadcrumb } from './Breadcrumb';
 import { Collector } from './Collector';
 
 /**
@@ -22,7 +23,17 @@ export interface UrlFilter {
   (url: string): string;
 }
 
-export interface HttpBreadCrumbOptions {
+/**
+ * Interface for breadcrumb filters.
+ *
+ * Given a breadcrumb the filter may return a modified breadcrumb or undefined to
+ * exclude the breadcrumb.
+ */
+export interface BreadcrumbFilter {
+  (breadcrumb: Breadcrumb): Breadcrumb | undefined;
+}
+
+export interface HttpBreadcrumbOptions {
   /**
    * If fetch should be instrumented and breadcrumbs included for fetch requests.
    *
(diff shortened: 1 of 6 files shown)


packages/telemetry/browser-telemetry/src/collectors/http/fetchDecorator.ts
```diff
@@ -77,7 +77,8 @@ export default function decorateFetch(callback: (breadcrumb: HttpBreadcrumb) =>
       return response;
     });
   }
-  wrapper.prototype = originalFetch.prototype;
+
+  wrapper.prototype = originalFetch?.prototype;
```
kinyoklion, line 81: Small robustness improvement exposed by running tests in VSCode.

packages/telemetry/browser-telemetry/src/options.ts
```diff
@@ -163,6 +170,11 @@ export default function parse(options: Options, logger?: MinLogger): ParsedOptio
         checkBasic('boolean', 'breadcrumbs.keyboardInput', logger),
       ),
       http: parseHttp(options.breadcrumbs?.http, defaults.breadcrumbs.http, logger),
+      breadcrumbFilters: itemOrDefault(
+        options.breadcrumbs?.breadcrumbFilters,
+        defaults.breadcrumbs.breadcrumbFilters,
+        checkBasic('array', 'breadcrumbs.breadcrumbFilters', logger),
```
kinyoklion, line 176: Cursory validation. Basically makes sure it is an array. If the function types are wrong, then that will be handled during the filtering.

packages/telemetry/browser-telemetry/src/api/Options.ts
```diff
@@ -131,7 +142,38 @@ export interface Options {
...
+     *       return;
+     *     }
+     *    return breadcrumb;
+     *   }
+     * ]
+     * ```
+     *
+     * If you want to redact or modify URLs in breadcrumbs, then a urlFilter should be used.
+     *
+     * If any breadcrumFilters throw an exception while processing a breadcrumb, then that breadcrumb will be excluded.
```
stasquatch, line 171:
```suggestion
     * If any breadcrumbFilters throw an exception while processing a breadcrumb, then that breadcrumb will be excluded.
```

stasquatch (approved):

PR #736, opened 2025-01-16
feat: Add basic logging support for browser-telemetry.

kinyoklion:
Adds the ability to specify a logger and to also use the base SDK logger when possible. The base SDK of a 3.x version does not expose the logger, but 4.x will. 

There is a minimal replication of the warning level of the logging interface to allow compatibility with both the 3.x SDK and the 4.x SDK.

Prefixing is done at message time instead of as part of the logger. Potentially it could be ideal to do it in the logger, but this approach makes it clear that the interpolation supported by most browser loggers will not come into play. (Where if the logger does this prefixing it is either complex or you lose the sprintf style formatting support.)

Some earlier log messages has been added, but a logger instance was not yet available.


packages/telemetry/browser-telemetry/src/BrowserTelemetryImpl.ts
```diff
@@ -127,16 +130,28 @@ export default class BrowserTelemetryImpl implements BrowserTelemetry {
...
+    this._setLogger();
     this._pendingEvents.forEach((event) => {
       this._client?.track(event.type, event.data);
     });
     this._pendingEvents = [];
   }
 
+  private _setLogger() {
+    this._logger =
+      this._options.logger ?? ((this._client as any)?.logger as MinLogger) ?? fallbackLogger;
```
kinyoklion, line 152: So, basically in the 4.x SDK there is a `logger` property, but in 3.x there is not. 

Additionally error monitoring should always be initialized as early as possible, which is before the SDK (there are also some inter-dependencies that require this order). As a result we need some intermediate logging.

packages/telemetry/browser-telemetry/src/BrowserTelemetryImpl.ts
```diff
@@ -153,7 +171,14 @@ export default class BrowserTelemetryImpl implements BrowserTelemetry {
     if (this._client === undefined) {
       this._pendingEvents.push({ type, data: event });
       if (this._pendingEvents.length > this._maxPendingEvents) {
-        // TODO: Log when pending events must be dropped. (SDK-915)
+        if (!this._eventsDropped) {
```
kinyoklion, line 174: Generally whenever a max is hit it will be hit many times. This will just log once to avoid creating log spam. This situation will generally represent misconfiguration where the client is not being registered.

Or something terribly wrong in the users application.

packages/telemetry/browser-telemetry/src/BrowserTelemetryImpl.ts
```diff
@@ -59,21 +60,6 @@ function applyBreadcrumbFilter(
   return breadcrumb === undefined ? undefined : filter(breadcrumb);
 }
 
-function applyBreadcrumbFilters(
```
kinyoklion, line 62: This moved to be a private member of the BrowserTelemetryImpl for access to a logger.

packages/telemetry/browser-telemetry/src/BrowserTelemetryImpl.ts
```diff
@@ -212,8 +238,27 @@ export default class BrowserTelemetryImpl implements BrowserTelemetry {
...
+    filters: BreadcrumbFilter[],
+  ): Breadcrumb | undefined {
+    try {
+      return filters.reduce(
+        (breadcrumbToFilter: Breadcrumb | undefined, filter: BreadcrumbFilter) =>
+          applyBreadcrumbFilter(breadcrumbToFilter, filter),
+        breadcrumb,
+      );
+    } catch (e) {
+      if (!this._breadcrumbFilterError) {
```
kinyoklion, line 255: If the filter was wrong once, then it will likely remain wrong, so we don't want to log spam in that situation.

packages/telemetry/browser-telemetry/src/BrowserTelemetryImpl.ts
```diff
@@ -175,7 +197,14 @@ export default class BrowserTelemetryImpl implements BrowserTelemetry {
     if (this._client === undefined) {
       this._pendingEvents.push({ type, data: event });
       if (this._pendingEvents.length > this._maxPendingEvents) {
-        // TODO: Log when pending events must be dropped. (SDK-915)
+        if (!this._eventsDropped) {
+          this._eventsDropped = true;
+          this._logger.warn(
+            prefixLog(
+              `Maximim pending events reached. Old events will be dropped until the SDK client is registered.`,
```
tanderson-ld, line 204: nit: maximum, and in related test

tanderson-ld (approved):

packages/telemetry/browser-telemetry/src/BrowserTelemetryImpl.ts
kinyoklion: Ty. I failed at typing.

PR #743, opened 2025-01-22
feat: Add the ability to filter errors.

kinyoklion:
This PR adds support for error level filtering. This is the highest level filter type capable of filtering on any data.

Other filters are provided to simplify the implementation of filtering across all error types. Breadcrumb filters could be implemented in terms of an error filter, but they would have higher complexity and different performance characteristics. For example with a breadcrumb filter we filter that breadcrumb regardless of how many events it may appear in, versus having to filter that same breadcrumb each time an event is captures.

Custom url filters operate at the HTTP capture level similarly reducing the frequency of redaction.

### DIFF
--- a/packages/telemetry/browser-telemetry/__tests__/options.test.ts
+++ b/packages/telemetry/browser-telemetry/__tests__/options.test.ts
@@ -1,4 +1,5 @@
 import { Breadcrumb } from '../src/api/Breadcrumb';
+import { ErrorData } from '../src/api/ErrorData';
 import ErrorCollector from '../src/collectors/error';
 import parse, { defaultOptions } from '../src/options';
 
@@ -16,7 +17,8 @@ it('handles an empty configuration', () => {
 });
 
 it('can set all options at once', () => {
-  const filter = (breadcrumb: Breadcrumb) => breadcrumb;
+  const breadcrumbFilter = (breadcrumb: Breadcrumb) => breadcrumb;
+  const errorFilter = (error: ErrorData) => error;
   const outOptions = parse({
     maxPendingEvents: 1,
     breadcrumbs: {
@@ -24,9 +26,10 @@ it('can set all options at once', () => {
       click: false,
       evaluations: false,
       flagChange: false,
-      filters: [filter],
+      filters: [breadcrumbFilter],
     },
     collectors: [new ErrorCollector(), new ErrorCollector()],
+    errorFilters: [errorFilter],
   });
   expect(outOptions).toEqual({
     maxPendingEvents: 1,
@@ -41,7 +44,7 @@ it('can set all options at once', () => {
         instrumentFetch: true,
         instrumentXhr: true,
       },
-      filters: expect.arrayContaining([filter]),
+      filters: expect.arrayContaining([breadcrumbFilter]),
     },
     stack: {
       source: {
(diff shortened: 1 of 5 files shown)


packages/telemetry/browser-telemetry/__tests__/options.test.ts
```diff
@@ -51,6 +54,7 @@ it('can set all options at once', () => {
       },
     },
     collectors: [new ErrorCollector(), new ErrorCollector()],
+    errorFilters: expect.arrayContaining([errorFilter]),
```
kinyoklion, line 57: The breadcrumbs filter is scoped to breadcrumbs, so it can just be called `filters`, but this is top level configuration so I think it needs `error` prepended.

packages/telemetry/browser-telemetry/src/BrowserTelemetryImpl.ts
```diff
@@ -198,8 +197,18 @@ export default class BrowserTelemetryImpl implements BrowserTelemetry {
    * @param event The event data.
    */
   private _capture(type: string, event: EventData) {
+    const filteredEvent = this._applyFilters(event, this._options.errorFilters, (e: unknown) => {
+      if (!this._errorFilterError) {
+        this._errorFilterError = true;
+        this._logger.warn(prefixLog(`Error applying error filters: ${e}`));
```
stasquatch, line 203: Does `e` need to be json stringified in order to print the value instead of `[Object object]` or the type?

kinyoklion: Generally no, as long as it is actually an error or a string. The default formatting for the error type will use the name and message (https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-error.prototype.tostring). 

Sometimes it is good to have custom error formatting when you aren't certain it will be an error.

There is some additional complexity generally with errors as they tend to either be self-referential, or have serialization which results in strange field output.

Just logging an error will typically use the message.
```
> console.log(`Error ${new Error("potato")}`)
Error Error: potato
```

JSON.stringify an error.
```
JSON.stringify(new Error("potato"))
'{}'
```

stasquatch: Ah! Gotcha, thanks for explaining.

stasquatch (approved):

packages/telemetry/browser-telemetry/src/BrowserTelemetryImpl.ts
kinyoklion: I will make a task to see if I can safely add some meta-data about the specific filter. If nothing else potentially the index.

stasquatch: is `e` always unknown here or can it be typed to an `Error` if it's always called from a try-catch?

kinyoklion: I could add a type guard using instanceof, but the typing of the catch is any/unknown depending on ts version. You technically can do really annoying things like `throw 10`.

stasquatch: Oh dang, TIL. Makes sense why you've implemented it as an `unknown` type.

PR #745, opened 2025-01-22
chore: Add browser-telemetry to release please.

kinyoklion:
Add release-please support for browser-telemetry and address a number of minor issues.

BEGIN_COMMIT_OVERRIDE
chore: Add browser-telemetry to release please.
fix: Export BrowserTelemetry, BrowserTelemetryInspector, and ImplementsCrumb.
chore: Update documentation comment links that were invalid.
feat: Rename initializeTelemetryInstance to initTelemetryInstance for consistency with initTelemetry. 
END_COMMIT_OVERRIDE

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -26,6 +26,7 @@ jobs:
       package-react-universal-release: ${{ steps.release.outputs['packages/sdk/react-universal--release_created'] }}
       package-browser-released: ${{ steps.release.outputs['packages/sdk/browser--release_created'] }}
       package-server-ai-released: ${{ steps.release.outputs['packages/sdk/server-ai--release_created'] }}
+      package-browser-telemetry-released: ${{ steps.release.outputs['packages/telemetry/browser-telemetry--release_created'] }}
     steps:
       - uses: googleapis/release-please-action@v4
         id: release
@@ -397,4 +398,24 @@ jobs:
         uses: ./actions/full-release
         with:
           workspace_path: packages/sdk/server-ai
+          aws_assume_role: ${{ vars.AWS_ROLE_ARN }}
+
+  release-browser-telemetry:
+    runs-on: ubuntu-latest
+    needs: ['release-please', 'release-browser']
+    permissions:
+      id-token: write
+      contents: write
+    if: ${{ always() && !failure() && !cancelled() && needs.release-please.outputs.package-browser-telemetry-released == 'true' }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20.x
+          registry-url: 'https://registry.npmjs.org'
+      - id: release-browser-telemetry
+        name: Full release of packages/telemetry/browser-telemetry
+        uses: ./actions/full-release
+        with:
+          workspace_path: packages/telemetry/browser-telemetry
           aws_assume_role: ${{ vars.AWS_ROLE_ARN }}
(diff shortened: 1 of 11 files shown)


.github/workflows/release-please.yml
```diff
@@ -398,3 +399,23 @@ jobs:
         with:
           workspace_path: packages/sdk/server-ai
           aws_assume_role: ${{ vars.AWS_ROLE_ARN }}
+
+  release-browser-telemetry:
+    runs-on: ubuntu-latest
+    needs: ['release-please', 'release-browser']
```
kinyoklion, line 405: Currently we are using the browser package for some type imports.

.github/workflows/release-please.yml
```diff
@@ -398,3 +399,23 @@ jobs:
...
+    permissions:
+      id-token: write
+      contents: write
+    if: ${{ always() && !failure() && !cancelled() && needs.release-please.outputs.package-browser-telemetry-released == 'true' }}
```
kinyoklion, line 409: This incantation allows us to depend on 'release-browser' without requiring that the browser package is being released.

Basically if release-browser is going to run, and it fails, cancel this step.
If it is ran, and it is a success, then run this step.
If it is not ran, but we need to release because we have out own changes, then run this step.

packages/telemetry/browser-telemetry/src/api/Breadcrumb.ts
```diff
@@ -68,7 +68,7 @@ export interface Breadcrumb {
 /**
  * Utility type which allows for easy extension of base breadcrumb type.
  */
-type ImplementsCrumb<U extends Breadcrumb> = U;
+export type ImplementsCrumb<U extends Breadcrumb> = U;
```
kinyoklion, line 71: This needs to be exported for documentation generation purposes, but it isn't expected that an application developer would use it.

packages/telemetry/browser-telemetry/src/api/Options.ts
```diff
@@ -212,8 +212,10 @@ export interface Options {
    *
    * Can be used to redact or modify error data.
    *
-   * For filtering breadcrumbs or URLs in error data, see {@link breadcrumbs.filters} and
-   * {@link breadcrumbs.http.customUrlFilter}.
+   * If any filter throws an exception, then the error data will be discarded.
+   *
+   * For filtering breadcrumbs or URLs in error data, refer to the `breadcrumbs.filters` option in {@link breadcrumbs} and
```
kinyoklion, line 217: Just some changes to get documentation links working.

release-please-config.json
```diff
@@ -65,6 +65,9 @@
           "jsonpath": "$.dependencies['@launchdarkly/server-sdk-ai']"
         }
       ]
+    },
+    "packages/telemetry/browser-telemetry": {
+      "bump-minor-pre-major": true
```
kinyoklion, line 70: Prior to 1.0 we want to increment the minor.

0.1.0 -> 0.2.0 etc.

When we do the 1.0 release this will be removed.

.github/workflows/release-please.yml
tanderson-ld: Consider adding this comment to the file for posterity.

tanderson-ld (approved):

.github/workflows/release-please.yml
kinyoklion: This pattern is used extensively for this project. We probably need a centralized explanation of how to do this.

PR #747, opened 2025-01-22
feat: adds fdv2 transactional persistent store

tanderson-ld:
**Related issues**

SDK-1029

Adds TransactionalPersistentStore containing memory store and persistence store.  Swaps on basis being provided as necessary.

Planning to add tests on holding branch (PR target branch).

### DIFF
--- a/packages/shared/sdk-server/src/store/InMemoryFeatureStore.ts
+++ b/packages/shared/sdk-server/src/store/InMemoryFeatureStore.ts
@@ -18,11 +18,11 @@ export default class InMemoryFeatureStore implements LDFeatureStore {
       if (Object.prototype.hasOwnProperty.call(items, key)) {
         const item = items[key];
         if (item && !item.deleted) {
-          return callback?.(item);
+          callback?.(item);
         }
       }
     }
-    return callback?.(null);
+    callback?.(null);
   }
 
   all(kind: DataKind, callback: (res: LDFeatureStoreKindData) => void): void {
@@ -110,7 +110,7 @@ export default class InMemoryFeatureStore implements LDFeatureStore {
   }
 
   initialized(callback: (isInitialized: boolean) => void): void {
-    return callback?.(this._initCalled);
+    callback?.(this._initCalled);
   }
 
   /* eslint-disable class-methods-use-this */
--- a/packages/shared/sdk-server/src/store/InMemoryFeatureStore.ts
+++ b/packages/shared/sdk-server/src/store/InMemoryFeatureStore.ts
@@ -110,7 +110,7 @@ export default class InMemoryFeatureStore implements LDFeatureStore {
   }
 
   initialized(callback: (isInitialized: boolean) => void): void {
-    callback?.(this._initCalled);
+    return callback?.(this._initCalled);
   }
 
   /* eslint-disable class-methods-use-this */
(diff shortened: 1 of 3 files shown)


packages/shared/sdk-server/src/store/InMemoryFeatureStore.ts
```diff
@@ -18,11 +18,11 @@ export default class InMemoryFeatureStore implements LDFeatureStore {
       if (Object.prototype.hasOwnProperty.call(items, key)) {
         const item = items[key];
         if (item && !item.deleted) {
-          return callback?.(item);
+          callback?.(item);
         }
       }
     }
-    return callback?.(null);
+    callback?.(null);
```
tanderson-ld, line 25: For reviewers: Not sure why these returns were here

packages/shared/sdk-server/src/store/PersistentDataStoreWrapper.ts
```diff
@@ -123,32 +123,35 @@ export default class PersistentDataStoreWrapper implements LDFeatureStore {
...
   init(allData: LDFeatureStoreDataStorage, callback: () => void): void {
     this._queue.enqueue((cb) => {
-      const afterStoreInit = () => {
-        this._isInitialized = true;
-        if (this._itemCache) {
-          this._itemCache.clear();
-          this._allItemsCache!.clear();
+      this._internalInit(allData, cb);
+    }, callback);
+  }
```
tanderson-ld, line 128: For reviewers: Adding queue protection for applyChanges down below.  To re-use functionality, moved queue entry into public functions and moved actual functionality into internal functions.  This should help future developers maintain the queue separation with less risk of multi entrant deadlock.

packages/shared/sdk-server/src/store/PersistentDataStoreWrapper.ts
```diff
@@ -218,59 +221,73 @@ export default class PersistentDataStoreWrapper implements LDFeatureStore {
...
+                this._internalUpsert({ namespace }, { key, ...item }, resolve); // callback intentionally not passed
+              }),
+            );
+          });
         });
-      });
-      _callback();
-    }
+        // invoke callback after all operations complete
+        Promise.all(promises).then(cb);
```
tanderson-ld, line 288: For reviewers: Please advise if there is a more efficient way to achieve combining callbacks.

tanderson-ld: Need to update so that each upsert request goes through the queue.  Underlying core implementations do not support parallel upsert.

tanderson-ld: Done

packages/shared/sdk-server/src/store/InMemoryFeatureStore.ts
kinyoklion: For something like this it is usually good to return whenever you make a callback, to ensure you don't execute more core unintentionally.

kinyoklion: Most importantly you don't want to execute the callback twice. Which removing the first return does here. It will execute for finding the item, then execute again for null. Returning whenever you execute a callback ensures you don't do this.

kinyoklion (changes requested):
Continuing to review, but there is a relevant comment to address.

packages/shared/sdk-server/src/store/InMemoryFeatureStore.ts
kinyoklion: You can read more here: https://github.com/eslint-community/eslint-plugin-n/blob/master/docs/rules/callback-return.md

I thought I had this rule enabled, but it appears some lint changes may be needed. Because the rule was removed from base eslint I guess? https://eslint.org/docs/latest/rules/callback-return

tanderson-ld: Reverted

kinyoklion (approved):

Issue #757, opened 2025-01-24
react-native iOS settings may not be defined with new architecture


quinnturner:
**Describe the bug**

For `settings.AppleLocale`, `settings` can be `null`/`undefined` using the new arch, so I think we also need a check _a.settings?.AppleLocale

**To reproduce**

When I loaded my application using React 0.77 with the new architecture on an iOS simulator with the latest iOS version `settings` was `undefined` or `null`

**Expected behavior**

Check for nullability on `settings`

**SDK version**
Latest

kinyoklion:
Hey @quinnturner, do you know if it is ever non-null (for instance on a real device)? I can approve your PR for now, but my concern here is that the code may actually need re-structured to make this call at a different time, or with a different API. So I will likely leave this issue open even after merging the PR until that behavior can be validated.

kinyoklion:
Ok. Just be aware that if you are using auto-environment attributes for targeting, that this value could be missing with this change.

kinyoklion:
Filed internally as _SDK-1084_

PR #758, opened 2025-01-24
fix(react-native): check for nullability in SettingsManager?.settings

quinnturner:
This check seems necessary for the new architecture (on React Native 0.77) to prevent an immediate crash.

Closes https://github.com/launchdarkly/js-core/issues/757


mobinni:
We ran into this as well, it would be great to get this in upstream

kinyoklion (approved):

PR #760, opened 2025-01-24
feat: Add cacheTtlMs option

keelerm84:
When the SDK retrieves data from the EdgeKV, it does so using a single
sub-request per call to `variation`, `variationDetail`, or
`allFlagsState`. The problem is that Akamai imposes a limit of 4
sub-requests per handler event.

If a customer evaluates more than 4 distinct flags during the handling
of a single event, subsequent flag lookups would fail. To combat this,
we now cache the flag values for a specified amount of time.

### DIFF
--- a/packages/sdk/svelte/svelte.config.js
+++ b/packages/sdk/svelte/svelte.config.js
@@ -11,8 +11,8 @@ const config = {
     // adapter-auto only supports some environments, see https://kit.svelte.dev/docs/adapter-auto for a list.
     // If your environment is not supported or you settled on a specific environment, switch out the adapter.
     // See https://kit.svelte.dev/docs/adapters for more information about adapters.
-    adapter: adapter()
-  }
+    adapter: adapter(),
+  },
 };
 
 export default config;
--- a/packages/sdk/svelte/svelte.config.js
+++ b/packages/sdk/svelte/svelte.config.js
@@ -11,8 +11,8 @@ const config = {
     // adapter-auto only supports some environments, see https://kit.svelte.dev/docs/adapter-auto for a list.
     // If your environment is not supported or you settled on a specific environment, switch out the adapter.
     // See https://kit.svelte.dev/docs/adapters for more information about adapters.
-    adapter: adapter(),
-  },
+    adapter: adapter()
+  }
 };
 
 export default config;
--- a/packages/shared/akamai-edgeworker-sdk/__tests__/featureStore/cacheableStore.test.ts
+++ b/packages/shared/akamai-edgeworker-sdk/__tests__/featureStore/cacheableStore.test.ts
@@ -66,21 +66,23 @@ describe('given a mock edge provider with test data', () => {
     });
 
     it('caches expires after duration', async () => {
+      jest.spyOn(Date, 'now').mockImplementation(() => 0);
       const cacheProvider = new CacheableStoreProvider(mockEdgeProvider, 'rootKey', 50);
       await cacheProvider.get('rootKey');
       await cacheProvider.get('rootKey');
       expect(mockGet).toHaveBeenCalledTimes(1);
 
-      await jest.advanceTimersByTimeAsync(20);
+      jest.spyOn(Date, 'now').mockImplementation(() => 20);
       await cacheProvider.get('rootKey');
       expect(mockGet).toHaveBeenCalledTimes(1);
 
-      await jest.advanceTimersByTimeAsync(30);
+      jest.spyOn(Date, 'now').mockImplementation(() => 50);
       await cacheProvider.get('rootKey');
       expect(mockGet).toHaveBeenCalledTimes(2);
     });
 
     it('prefetch respects cache TTL', async () => {
+      jest.spyOn(Date, 'now').mockImplementation(() => 0);
       const cacheProvider = new CacheableStoreProvider(mockEdgeProvider, 'rootKey', 50);
       await cacheProvider.get('rootKey');
       await cacheProvider.get('rootKey');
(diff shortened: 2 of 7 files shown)


packages/shared/akamai-edgeworker-sdk/src/featureStore/cacheableStoreProvider.ts
```diff
@@ -2,14 +2,13 @@ import { EdgeProvider } from '.';
 
 /**
  * Wraps around an edge provider to cache a copy of the sdk payload locally an explicit request is made to refetch data from the origin.
- * The wrapper is neccessary to ensure that we dont make redundant sub-requests from Akamai to fetch an entire environment payload.
+ * The wrapper is necessary to ensure that we don't make redundant sub-requests from Akamai to fetch an entire environment payload.
```
kinyoklion, line 5: We should maybe note that, at the time of writing, there is a 4 request sub-limit.

I think we also want to update the docs to indicate using the SDK uses one of their sub-requests. In case they already need to make their own sub requests for other reasons.

packages/sdk/svelte/svelte.config.js
kinyoklion, line 1: We'll want to revert this file.

packages/shared/akamai-edgeworker-sdk/__tests__/featureStore/cacheableStore.test.ts
```diff
@@ -0,0 +1,96 @@
+import { EdgeProvider } from '../../src/featureStore';
+import CacheableStoreProvider from '../../src/featureStore/cacheableStoreProvider';
+import * as testData from '../testData.json';
+
+describe('CacheableStoreProvider', () => {
```
kinyoklion, line 5:
```suggestion
describe('given a mock edge provider with test data', () => {
```

packages/shared/akamai-edgeworker-sdk/__tests__/featureStore/cacheableStore.test.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  describe('with infinite cache ttl', () => {
+    it('caches initial request', async () => {
+      const cacheProvider = new CacheableStoreProvider(mockEdgeProvider, 'rootKey', 0);
+      await cacheProvider.get('rootKey');
+      await cacheProvider.get('rootKey');
+      expect(mockGet).toHaveBeenCalledTimes(1);
+    });
+
+    it('prefetch does not reset', async () => {
```
kinyoklion, line 47:
```suggestion
    it('does not reset on prefetch', async () => {
```

I know I have not gone through and refactored all the tests in the project, but I try to write these such that `it` is part of what you read.

packages/shared/akamai-edgeworker-sdk/__tests__/featureStore/cacheableStore.test.ts
```diff
@@ -0,0 +1,96 @@
...
+    });
+
+    it('caches expires after duration', async () => {
+      const cacheProvider = new CacheableStoreProvider(mockEdgeProvider, 'rootKey', 50);
+      await cacheProvider.get('rootKey');
+      await cacheProvider.get('rootKey');
+      expect(mockGet).toHaveBeenCalledTimes(1);
+
+      // eslint-disable-next-line no-promise-executor-return
+      await new Promise((resolve) => setTimeout(resolve, 60));
```
kinyoklion, line 74: Instead of depending on time we instead should use fake timers.

Setup
```
jest.useFakeTimers();
```

If we want all pending timers to execute, then we use:
```
await jest.runAllTimersAsync()'
```
Alternatively you can advance them by a specific amount.
```
await jest.advanceTimersByTimeAsync(100); // advances by 100ms, things scheduled for less than that should execute.
```

packages/shared/akamai-edgeworker-sdk/src/featureStore/cacheableStoreProvider.ts
```diff
@@ -1,15 +1,34 @@
...
+ * - 10 for enterprise
  */
 export default class CacheableStoreProvider implements EdgeProvider {
   cache: string | null | undefined;
+  cachedAt: number | undefined;
 
   constructor(
     private readonly _edgeProvider: EdgeProvider,
     private readonly _rootKey: string,
+    private readonly _cacheTtlMs: number | undefined = undefined,
```
kinyoklion, line 31:
```suggestion
    private readonly _cacheTtlMs?: number,
```

packages/shared/akamai-edgeworker-sdk/src/featureStore/cacheableStoreProvider.ts
```diff
@@ -18,22 +37,50 @@ export default class CacheableStoreProvider implements EdgeProvider {
...
+      }
+      this.cachedAt = Date.now();
     }
 
     return this.cache;
   }
 
   /**
-   * Invalidates cache and fetch environment payload data from origin. The result of this data is cached in memory.
+   * Fetches environment payload data from the origin in according with the caching configuration.
```
kinyoklion, line 52:
```suggestion
   * Fetches environment payload data from the origin in accordance with the caching configuration.
```

kinyoklion (changes requested):

packages/sdk/svelte/svelte.config.js
keelerm84: Done.

packages/shared/akamai-edgeworker-sdk/__tests__/featureStore/cacheableStore.test.ts
keelerm84: Done.

kinyoklion (approved):

kinyoklion (approved):

Issue #777, opened 2025-02-12
LDClient.flush() does not reject if there is an error sending analytics


hysmio:
**Describe the bug**
`LDClient.flush()` does not reject like it states if the underlying `EventProcessor.flush()` call throws, irrespective of callback. 

if a callback is provided, the callback will receive the error, the promise will not reject as intended. If no callback is provided, the promise will still not reject, which is unintended as per the JSDoc.

**To reproduce**
1. Call `await LDClient.flush()` with an explicit `eventsUri` that has an error status code, like 503.
2. It will not throw an error
3. Call it again but with a callback `await LDClient.flush((error, success) => console.error(success, error));`, you will get a console error output.

**Expected behavior**
The promise on `await LDClient.flush()` should reject if there is an error.

**Logs**
If applicable, add any log output related to your problem.

**SDK version**
```
@launchdarkly/node-server-sdk@9.7.3
@launchdarkly/js-server-sdk-common@2.11.0
@launchdarkly/js-sdk-common@2.13.0
```

**Language version, developer tools**
```
node - v20.11.1
```

**OS/platform**
```
Windows 11
WSL - Ubuntu 24.04
```

**Solution**
Added some comments to explain what the solution I think could be, **also found that the `callback` is called with an error and then immediately again without one, so that should also be fixed.**

```ts
# LDClient.flush
  async flush(callback?: (err: Error | null, res: boolean) => void): Promise<void> {
    try {
      await this._eventProcessor.flush();
    } catch (err) {
      if (!callback) { // If callback isn't set, throw the error to the caller.
        throw err;
      }
      // also this will currently callback twice regardless, there should probably be a return after this call
      callback(err as Error, false);
    }
    callback?.(null, true);
  }
```ts

kinyoklion:
@hysmio Thank you for the report.

_Filed internally as SDK-777-

kinyoklion:
@hysmio I am interested in the use case for the flush error handling. The callback error needs resolved for sure, and I will be making an issue to handle it correctly.

I am hesitant to actually reject the flush promise. First because it has not been rejected in several years and as a result people are likely not handling it correctly.

Generally speaking manually flushing the client isn't often needed, aside from short running instance (start, do some work, flush and close).

If the flush fails, the events are already gone by that point, and flushing again would either have no events to flush (as they were discarded from the last flush), or would flush new accumulated events since the last time flush was called.

So I mainly want to validate that the thought isn't to retry, as the events would already be gone.

[...]

PR #779, opened 2025-02-14
fix: Fix issue where flush callback could be called twice.

kinyoklion:
Relates to #777


packages/shared/sdk-server/src/LDClientImpl.ts
kinyoklion, line 1: Based on our interface documentation there is another issue with this implementation. We say the function rejects, but it doesn't actually. I am more concerned about changing that as it is likely to introduce problems. We may instead consider changing the API documentation and suggesting using the callback if the error is required.

Additionally you cannot retry if it fails, at least not the same payload, as those events will have been discarded by that time.

PR #783, opened 2025-02-18
fix: Fix issue processing URLs for fetch and XHR requests.

kinyoklion:
A URL filter was throwing an exception. This fixes the filter to prevent that exception, but it also makes the fetch/xhr wrappers more resilient to ensure we do not prevent requests from functioning when a filter fails.

### DIFF
--- a/packages/telemetry/browser-telemetry/src/BrowserTelemetryImpl.ts
+++ b/packages/telemetry/browser-telemetry/src/BrowserTelemetryImpl.ts
@@ -115,8 +115,6 @@ export default class BrowserTelemetryImpl implements BrowserTelemetry {
   private _registrationComplete: boolean = false;
 
   // Used to ensure we only log the event dropped message once.
-  private _clientRegistered: boolean = false;
-  // Used to ensure we only log the event dropped message once.
   private _eventsDropped: boolean = false;
   // Used to ensure we only log the breadcrumb filter error once.
   private _breadcrumbFilterError: boolean = false;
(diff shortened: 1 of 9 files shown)


packages/telemetry/browser-telemetry/src/BrowserTelemetryImpl.ts
```diff
@@ -133,6 +131,10 @@ export default class BrowserTelemetryImpl implements BrowserTelemetry {
     this._maxPendingEvents = _options.maxPendingEvents;
     this._maxBreadcrumbs = _options.breadcrumbs.maxBreadcrumbs;
 
+    // Set the initial logger, it may be replaced when the client is registered.
```
kinyoklion, line 134: Logger can potentially be used earlier in the process now.

packages/telemetry/browser-telemetry/src/filters/defaultUrlFilter.ts
```diff
@@ -14,18 +14,27 @@ function authorityUrlFilter(url: string): string {
   // This will work in browser environments, but in the future we may want to consider an approach
   // which doesn't rely on the browser's URL parsing. This is because other environments we may
   // want to target, such as ReactNative, may not have as robust URL parsing.
-  const urlObj = new URL(url);
```
kinyoklion, line 17: This was failing with relative URLs.

stasquatch (approved):

louis-launchdarkly (approved):

PR #786, opened 2025-02-21
feat!: Replace prefetch behavior with simple TTL cache

keelerm84:
Previously, the LDClient would issue a call to prime the store with
data, which would be retained for the lifetime of the variation or all
flags call.

This priming call is being removed in favor of a simple TTL cache. The
cache will be populated on the initial usage of the SDK, and then
periodically as it is detected to be expired.

The TTL can be configured with:

- Positive value representing the time to cache the value
- 0 to cache the value indefinitely. This allows a customer to
  initialize the SDK within an EdgeWorker handler, and get a "snapshot"
  of the view for the lifetime of the SDK.
- Negative value representing no cache. This value is highly discouraged
  as usage restrictions in Akamai make it ineffective.


packages/shared/akamai-edgeworker-sdk/src/featureStore/index.ts
```diff
@@ -99,6 +93,30 @@ export class EdgeFeatureStore implements LDFeatureStore {
     }
   }
 
+  // This method is used to retrieve the environment payload from the edge
```
kinyoklion, line 96: ``/** doc comment`

packages/shared/akamai-edgeworker-sdk/src/featureStore/index.ts
```diff
@@ -99,6 +93,30 @@ export class EdgeFeatureStore implements LDFeatureStore {
...
+  // provider. It will cache the payload for the duration of the cacheTtlMs.
+  private async _getStorePayload(): Promise<ReturnType<typeof deserializePoll>> {
+    let item = this._cache.get();
+    if (item !== undefined) {
+      return item;
+    }
+
+    const i = await this._edgeProvider.get(this._rootKey);
```
kinyoklion, line 104: Change from i

tanderson-ld (approved):

kinyoklion (approved):

PR #810, opened 2025-03-25
feat: Support inline context for custom and migration events

abarker-launchdarkly:
**Requirements**

- Support inlining context for custom events.  The logic for this needs to happen in EventProcessor which is shared between server and client sdks in js-core.  It has been verified with product analytics that this is ok for the client sdks.
- Support inlining context for migration events.  This causes a change in the public interface for `LDClient.trackMigration` where the `LDMigrationOpEvent` parameter has been modified.  `contextKeys` has been made optional and deprecated in favor of `context` and all internal uses of `LDMigrationOpEvent` have been changed to use `context`.

BEGIN_COMMIT_OVERRIDE
feat: Support inline context for custom and migration events
fix: Deprecate LDMigrationOpEvent.contextKeys in favor of LDMigrationOpEvent.context
END_COMMIT_OVERRIDE

### DIFF
--- a/packages/sdk/browser/contract-tests/suppressions.txt
+++ b/packages/sdk/browser/contract-tests/suppressions.txt
@@ -4,3 +4,15 @@ streaming/requests/URL path is computed correctly/no environment filter/base URI
 streaming/requests/context properties/single kind minimal/REPORT
 streaming/requests/context properties/single kind with all attributes/REPORT
 streaming/requests/context properties/multi-kind/REPORT
+tags/stream requests/{"applicationId":null,"applicationVersion":"._-abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ012345678"}
+tags/stream requests/{"applicationId":null,"applicationVersion":"________________________________________________________________"}
+tags/stream requests/{"applicationId":"","applicationVersion":"._-abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ012345678"}
+tags/stream requests/{"applicationId":"","applicationVersion":"________________________________________________________________"}
+tags/stream requests/{"applicationId":"._-abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ012345678","applicationVersion":null}
+tags/stream requests/{"applicationId":"._-abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ012345678","applicationVersion":""}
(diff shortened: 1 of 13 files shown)


packages/shared/common/src/internal/events/EventProcessor.ts
```diff
@@ -83,9 +83,10 @@ interface PageviewOutputEvent {
  */
 type DiagnosticEvent = any;
 
-interface MigrationOutputEvent extends Omit<InputMigrationEvent, 'samplingRatio'> {
+interface MigrationOutputEvent extends Omit<InputMigrationEvent, 'samplingRatio' | 'context'> {
   // Make the sampling ratio optional so we can omit it when it is one.
   samplingRatio?: number;
+  context?: FilteredContext;
```
kinyoklion, line 89: We may want a comment here. 

I understand we are replacing one context with another, but I am not clear why it would be optional. (After reading forward it looks like maybe for compatibility. Theoretically its `internal` and shouldn't be outside the SDKs. Though we should double check nobody is actually floating its dependency in the repo, as that would be quite bad.)

packages/sdk/browser/contract-tests/suppressions.txt
```diff
@@ -4,3 +4,15 @@ streaming/requests/URL path is computed correctly/no environment filter/base URI
 streaming/requests/context properties/single kind minimal/REPORT
 streaming/requests/context properties/single kind with all attributes/REPORT
 streaming/requests/context properties/multi-kind/REPORT
+tags/stream requests/{"applicationId":null,"applicationVersion":"._-abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ012345678"}
```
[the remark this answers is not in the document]
abarker-launchdarkly: I was debugging this with @keelerm84 and he pointed out that there is a problem that needs addressed and suggested that for now they should be added to the suppression list

packages/shared/common/src/internal/events/EventProcessor.ts
abarker-launchdarkly: The reason this is optional is because we are still supporting contextKeys in the public interface for `LDClient.trackMigration()` so at this point in the code, we could either have `contextKeys` or `context`

packages/sdk/browser/contract-tests/suppressions.txt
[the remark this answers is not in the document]
keelerm84: Ryan, for these specifically I think this is just the fact that you can't affect the headers on streaming requests right?

packages/shared/common/src/internal/events/EventProcessor.ts
kinyoklion: Yeah, I think that is probably worth adding a comment here in the code. As this is in a common package it is pretty easy to lose the relevant context about the specific SDKs..

abarker-launchdarkly: Comment added

kinyoklion: LGTM. Thank you!

PR #814, opened 2025-03-27
feat: Option to use gzip to compress event

abarker-launchdarkly:
**Requirements**

LDConfig option `enableEventCompression` was added for js-server-sdk-common and NodeRequests was modified to support gzip compression of POST bodies if this option is enabled.

### DIFF
--- a/packages/sdk/server-node/src/platform/NodeRequests.ts
+++ b/packages/sdk/server-node/src/platform/NodeRequests.ts
@@ -105,7 +105,7 @@ export default class NodeRequests implements platform.Requests {
 
   private _hasProxyAuth: boolean = false;
 
-  private _enableEventCompression: boolean = false;
+  private _enableBodyCompression: boolean = false;
 
   constructor(
     tlsOptions?: LDTLSOptions,
@@ -116,7 +116,7 @@ export default class NodeRequests implements platform.Requests {
     this._agent = createAgent(tlsOptions, proxyOptions, logger);
     this._hasProxy = !!proxyOptions;
     this._hasProxyAuth = !!proxyOptions?.auth;
-    this._enableEventCompression = !!enableEventCompression;
+    this._enableBodyCompression = !!enableEventCompression;
   }
 
   async fetch(url: string, options: platform.Options = {}): Promise<platform.Response> {
@@ -135,7 +135,7 @@ export default class NodeRequests implements platform.Requests {
     // enableEventCompression config setting is true and the compressBodyIfPossible
     // option is true.
     else if (
-      this._enableEventCompression &&
+      this._enableBodyCompression &&
       !!options.compressBodyIfPossible &&
       options.method?.toLowerCase() === 'post' &&
       options.body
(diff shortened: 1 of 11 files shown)


packages/shared/common/src/api/platform/Requests.ts
```diff
@@ -75,6 +75,11 @@ export interface Options {
   headers?: Record<string, string>;
   method?: string;
   body?: string;
+  /**
+   * Gzip compress the post body only if the underlying SDK framework supports it
+   * and the config option enableEventCompression is set to true.
+   */
+  compressBodyIfPossible?: boolean;
```
tanderson-ld, line 82: I recommend against mentioning the config option here.  This layer shouldn't know about the config option as the platform/Requests interface may be used in other cases in the future.

abarker-launchdarkly: I did this based on recommendation from @kinyoklion because currently the only post that is done is for events... but we wanted to make it flexible in the future if we do a different post for something else we can call call `Reqests.fetch()` and pass `compressBodyIfPossible=false` to override the gzip behavior.

packages/sdk/server-node/src/platform/NodeRequests.ts
```diff
@@ -101,25 +105,44 @@ export default class NodeRequests implements platform.Requests {
...
   private _hasProxyAuth: boolean = false;
 
-  constructor(tlsOptions?: LDTLSOptions, proxyOptions?: LDProxyOptions, logger?: LDLogger) {
+  private _enableEventCompression: boolean = false;
+
+  constructor(
+    tlsOptions?: LDTLSOptions,
+    proxyOptions?: LDProxyOptions,
+    logger?: LDLogger,
+    enableEventCompression?: boolean,
```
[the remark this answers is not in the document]
abarker-launchdarkly: This one is interesting because options are passed to NodeRequests before they are validated, so it will either have the value passed in at client initialization or be undefined (which would be coerced to false in line 119 below).

packages/shared/common/src/api/platform/Requests.ts
kinyoklion: The most important part of the internal interfaces is us knowing the implications. The relationship in this case is a little odd at the moment. Likely we would want the node platform to just support compression, and enableEventCompression to flow into `compressBodyIfPossible` for events.

tanderson-ld (approved):

packages/sdk/server-node/src/platform/NodeRequests.ts
```diff
@@ -101,25 +105,44 @@ export default class NodeRequests implements platform.Requests {
 
   private _hasProxyAuth: boolean = false;
 
-  constructor(tlsOptions?: LDTLSOptions, proxyOptions?: LDProxyOptions, logger?: LDLogger) {
+  private _enableEventCompression: boolean = false;
```
kinyoklion, line 108: We should probably just call this `_enableBodyCompression` at this level.

abarker-launchdarkly: Fixed

PR #821, opened 2025-04-14
feat: Add support for multiple context event summaries.

kinyoklion:


packages/sdk/browser/src/BrowserClient.ts
```diff
@@ -122,7 +122,7 @@ export class BrowserClient extends LDClientImpl implements LDClient {
       clientSideId,
       autoEnvAttributes,
       platform,
-      filterToBaseOptions({ ...options, logger }),
```
kinyoklion, line 125: Unrelated bug, will be fixed before this goes forward.

packages/shared/common/__tests__/internal/events/EventSummarizer.test.ts
```diff
@@ -12,16 +12,16 @@ describe('given an event summarizer', () => {
   });
 
   it('does nothing for an identify event.', () => {
-    const beforeSummary = summarizer.getSummary();
+    const beforeSummary = summarizer.getSummaries()[0];
```
kinyoklion, line 15: These tests are just updated to handle the API difference added by supporting multiple summaries.

packages/shared/common/__tests__/setupCrypto.ts
```diff
@@ -1,14 +1,25 @@
 import { Hasher } from '../src/api';
 
+class MockHasher implements Hasher {
```
kinyoklion, line 3: The mock hasher implementation that was being used for client-side was not sufficiently representative of a real hash.

packages/shared/common/src/AttributeReference.ts
```diff
@@ -144,4 +144,8 @@ export default class AttributeReference {
       this._components.every((value, index) => value === other.getComponent(index))
     );
   }
+
+  public get components() {
```
kinyoklion, line 148: This is only public to within the SDK.

packages/shared/common/src/Context.ts
```diff
@@ -465,4 +466,74 @@ export default class Context {
   public get legacy(): boolean {
     return this._wasLegacy;
   }
+
+  public async hash(crypto: Crypto): Promise<string | undefined> {
+    if (!this.valid) {
+      return undefined;
+    }
+
+    const hasher = crypto.createHash('sha256');
```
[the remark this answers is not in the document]
kinyoklion: If we wanted a deterministic SDK-to-SDK implementation we could spec out the hash order.

Though other SDKs already have hash code implementations: https://github.com/launchdarkly/flutter-client-sdk/blob/016e0880fd1145d75bf8de6acd932e0497bfb768/packages/common/lib/src/ld_context.dart#L91

packages/shared/common/src/api/subsystem/LDEventSummarizer.ts
kinyoklion, line 1: The event summarizer needed an interface to make it swappable for client/server.

packages/shared/common/src/internal/events/EventProcessor.ts
```diff
@@ -143,6 +146,15 @@ export default class EventProcessor implements LDEventProcessor {
       _config.privateAttributes.map((ref) => new AttributeReference(ref)),
     );
 
+    if (summariesPerContext) {
```
kinyoklion, line 149: We theoretically may want to move this up the stack and use dependency injection instead. Though it will require some shuffling of things around.

packages/shared/common/src/Context.ts
```diff
@@ -465,4 +466,74 @@ export default class Context {
...
+        hasher.update(String(target));
+      }
+    }
+
+    if (hasher.digest) {
+      return hasher.digest('hex');
+    }
+
+    // The hasher must have either digest or asyncDigest.
+    const digest = await hasher.asyncDigest!('hex');
```
kinyoklion, line 536: This async digest is going to cause some problems. Because the event has to be enqueued asynchronously, which doesn't fit with our current contract of things.

packages/shared/common/src/Context.ts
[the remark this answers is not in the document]
kinyoklion: I am going to think through this. I think a standard hash code+equality would be better, but it also is going to increase the complication of the collection.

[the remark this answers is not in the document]
kinyoklion: I think hash is fine here. The change of a collision is basically 0 and this is simpler.

packages/shared/common/src/internal/events/MultiEventSummarizer.ts
```diff
@@ -0,0 +1,43 @@
...
+          this._summarizers[hash] = new EventSummarizer(true, this._contextFilter);
+          summarizer = this._summarizers[hash];
+        }
+
+        summarizer.summarizeEvent(event);
+      }
+    })();
+  }
+
+  getSummaries(): SummarizedFlagsEvent[] {
```
kinyoklion, line 38: Moving to a model where getting the summaries automatically clears. It makes it easier to validate there are no race conditions from async interactions.

tanderson-ld (approved):

package.json
```diff
@@ -70,6 +70,7 @@
   "packageManager": "yarn@3.4.1",
   "//": "Pin jsonc-parser because v3.3.0 breaks rollup-plugin-esbuild",
   "resolutions": {
-    "jsonc-parser": "3.2.0"
+    "jsonc-parser": "3.2.0",
+    "parse5": "7.2.1"
```
kinyoklion, line 74: Pin parse5 because it currently has a typescript problem.

kinyoklion:
Closed to improve hash.

PR #823, opened 2025-04-15
feat: Environment ID support for hooks

abarker-launchdarkly:
**Requirements**

If present, `environmentId` is now passed to the HookRunner in the evaluation series.

For streaming connections:
- Response headers are now attached to the `open` event (https://github.com/launchdarkly/js-eventsource/pull/33)
- `StreamingProcessor` passes these headers to the stream listeners via `processJson`.  The listener for the PUT event extracts the `environmentId` from the headers and passes this as initialization metadata to the underlying feature store.

For polling connections:
- `PollingProcessor` retrieves  the response headers via the underlying `Requestor` and extracts the `environmentId` from the headers and passes this as initialization metadata to the underlying feature store.

LDClient will then call `getInitMetaData()` on the underlying feature store (if the feature store supports it) when executing a hook and pass `environmentId` in the execution series data.

Currently only `InMemoryFeatureStore` has been modified to support initialization metadata.  This functionality can be added to other feature stores by modifying `init()` to accept the optional `initMetadata` parameter and implementing the optional `getInitMetadata()` method.

### DIFF
--- a/packages/shared/sdk-server/src/LDClientImpl.ts
+++ b/packages/shared/sdk-server/src/LDClientImpl.ts
@@ -344,6 +344,7 @@ export default class LDClientImpl implements LDClient {
               },
             );
           }),
+        this._featureStore.getInitMetaData?.()?.environmentId,
       )
       .then((detail) => {
         callback?.(null, detail.value);
@@ -375,6 +376,7 @@ export default class LDClientImpl implements LDClient {
             },
           );
         }),
+      this._featureStore.getInitMetaData?.()?.environmentId,
     );
   }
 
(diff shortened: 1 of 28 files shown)


packages/shared/sdk-server/src/LDClientImpl.ts
```diff
@@ -615,6 +620,7 @@ export default class LDClientImpl implements LDClient {
       defaultValue,
       MIGRATION_VARIATION_METHOD_NAME,
       () => this._migrationVariationInternal(key, context, defaultValue),
+      this._featureStore.getInitMetaData?.()?.environmentId,
```
kinyoklion, line 623: `?.()?.`
Feels like it is getting close to an emoticon for a boxer.

PR #827, opened 2025-04-17
feat: Add hook support for the track series.

kinyoklion:
Adds support for the track series to the client-side hook implementation.


packages/shared/sdk-client/__tests__/hooks/HookRunner.test.ts
```diff
@@ -301,4 +303,161 @@ describe('given a hook runner and test hook', () => {
...
+
+    emptyHookRunner.afterTrack({
+      key: 'test',
+      context: { kind: 'user', key: 'user-123' },
+    });
+
+    expect(logger.error).not.toHaveBeenCalled();
+  });
+
+  it('executes hook stages in the specified order', () => {
```
kinyoklion, line 360: This test was missing from the client-side implementation previously. Added it and extended it for track.

tanderson-ld (approved):

PR #833, opened 2025-04-25
chore: adds LDTransactionalFeatureStore, LDTransactionalDataSourceUpdates, and FDv2 DataSource impls.

tanderson-ld:
**Requirements**

This PR moves changes for FDv2 from the temporary holding branch to main. All picked code has been previously reviewed and was picked from the temporary holding branch using file pick commands.

Resolved conflicts with @abarker-launchdarkly's recent initMetadata changes.  Please verify those conflict resolutions @abarker-launchdarkly .


packages/shared/sdk-server/src/api/subsystems/LDFeatureStore.ts
```diff
@@ -144,6 +144,28 @@ export interface LDFeatureStore {
...
+   * @param basis If true, completely overwrites the current contents of the data store
+   * with the provided data.  If false, upserts the items in the provided data.  Upserts
+   * are made only if provided items have newer versions than existing items.
+   * @param data An object in which each key is the "namespace" of a collection (e.g. `"features"`) and
+   * the value is an object that maps keys to entities. The actual type of this parameter is
+   * `interfaces.FullDataSet<VersionedData>`.
+   * @param selector opaque string that uniquely identifies the state that contains the changes
+   * @param callback Will be called after the changes are applied.
+   */
+  applyChanges(
```
kinyoklion, line 161: This is a breaking change, which would require the SDKs be major versioned.

It is possible to do this as an optional method, but require that method to support FDv2.

This interface is both exported and expected to be implemented by external packages. So, for example, node server would need major versioned and then the persistence packages would need major versioned with that new version constraint.

kinyoklion (changes requested):

packages/shared/sdk-server/src/api/subsystems/LDFeatureStore.ts
kinyoklion: Also, the persistent store implementations don't really need to implement this if it is handled at the wrapper.

So I would expect either two interfaces, or one interface with an optional method addition.

The two interfaces would allow for internal type safety, as types used with V2 would need to implement it.

tanderson-ld: Updated to have LDTransactionalFeatureStore and LDTransactionalDataSourceUpdates interfaces which contain the additional transactional method.  Going to optimize the interface declaration to use `Partial`.

packages/shared/sdk-server/src/api/subsystems/LDDataSourceUpdates.ts
```diff
@@ -47,4 +47,23 @@ export interface LDDataSourceUpdates {
...
+   * with the provided data.  If false, upserts the items in the provided data.  Upserts
+   * are made only if provided items have newer versions than existing items.
+   * @param data An object in which each key is the "namespace" of a collection (e.g. `"features"`) and
+   * the value is an object that maps keys to entities. The actual type of this parameter is
+   * `interfaces.FullDataSet<VersionedData>`.
+   * @param callback Will be called after the changes are applied.
+   * @param initMetadata Optional metadata to initialize the data source with.
+   * @param selector opaque string that uniquely identifies the state that contains the changes
+   */
+  applyChanges(
```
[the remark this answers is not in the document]
tanderson-ld: Updated to have LDTransactionalFeatureStore and LDTransactionalDataSourceUpdates interfaces which contain the additional transactional method.  Going to optimize the interface declaration to use `Partial`.

packages/shared/sdk-server/__tests__/data_sources/PollingProcessorFDv2.test.ts
```diff
@@ -0,0 +1,233 @@
+import { DataSourceErrorKind, LDPollingError, subsystem } from '../../src';
+import PollingProcessorFDv2 from '../../src/data_sources/PollingProcessorFDv2';
+import Requestor from '../../src/data_sources/Requestor';
+import TestLogger, { LogLevel } from '../Logger';
+
+describe('given an event processor', () => {
```
kinyoklion, line 6:
```suggestion
describe('given a polling processor', () => {
```

packages/shared/sdk-server/__tests__/data_sources/PollingProcessorFDv2.test.ts
```diff
@@ -0,0 +1,233 @@
...
+  afterEach(() => {
+    processor.stop();
+    jest.resetAllMocks();
+  });
+
+  it('polls repeatedly', (done) => {
+    requestor.requestAllData = jest.fn((cb) => cb(undefined, jsonData));
+
+    processor.start(mockDataCallback, mockStatusCallback);
+    setTimeout(() => {
```
kinyoklion, line 142: I would instead set this up to await a couple calls, then let the test timeout indicate failure. Potentially using the async queue and awaiting a couple `take` calls. This gives it a total of 5 seconds to happen, the test timeout, but doesn't cause it to require at least 500ms. Also this test may flake under high load as is.

packages/shared/sdk-server/src/api/subsystems/LDTransactionalDataSourceUpdates.ts
```diff
@@ -0,0 +1,30 @@
...
+   * `interfaces.FullDataSet<VersionedData>`.
+   * @param callback Will be called after the changes are applied.
+   * @param initMetadata Optional metadata to initialize the data source with.
+   * @param selector opaque string that uniquely identifies the state that contains the changes
+   */
+  applyChanges(
+    basis: boolean,
+    data: LDFeatureStoreDataStorage,
+    callback: () => void,
+    initMetadata?: InitMetadata,
```
kinyoklion, line 27: Did you need the type alias for some reason?

packages/shared/sdk-server/src/data_sources/OneShotInitializerFDv2.ts
```diff
@@ -0,0 +1,100 @@
...
+        this._logger?.debug(`${err} - Body follows: ${body}`);
+        statusCallback(
+          subsystemCommon.DataSourceState.Closed,
+          new LDPollingError(DataSourceErrorKind.InvalidData, 'Malformed data in polling response'),
+        );
+      }
+    });
+  }
+
+  stop() {
```
kinyoklion, line 97: I am not sure about this. Do all the callbacks get disconnected from this, in the situation we don't want to use it anymore?

It is a little different for client-side, but I saw this assumption is causing problems in the client-side SDK where a poll completes after stop and the data propagates.

In this case maybe the unintended propagation would be fine? But maybe it could cause an unexpected basis to apply a delta to?

packages/shared/sdk-server/src/data_sources/PollingProcessorFDv2.ts
```diff
@@ -0,0 +1,155 @@
...
+    statusCallback: (status: subsystemCommon.DataSourceState, err?: any) => void,
+  ) {
+    if (this._stopped) {
+      return;
+    }
+
+    const startTime = Date.now();
+    this._logger?.debug('Polling LaunchDarkly for feature flag updates');
+    this._requestor.requestAllData((err, body) => {
+      const elapsed = Date.now() - startTime;
```
kinyoklion, line 44: We likely want a _stopped check inside the callback, so we do not apply data/status from a stopped data source.

packages/shared/sdk-server/src/data_sources/TransactionalDataSourceUpdates.ts
```diff
@@ -0,0 +1,193 @@
...
+  rules?: [
+    {
+      // The shape of rules are different between flags and segments, but
+      // both have clauses of the same shape.
+      clauses?: Clause[];
+    },
+  ];
+}
+
+function computeDependencies(namespace: string, item: LDFeatureStoreItem) {
```
kinyoklion, line 34: This code was just moved, correct?

kinyoklion: Or is it a duplicate?

kinyoklion (changes requested):
I have some concern that ordering is much more important now, so we should maybe consider if we need a queue.

In the old stuff you would have a stream and you know you will get puts and then later patches and the way the event source works we really aren't going to break that. Polling included the entire payload, so even http requests completing out of order wasn't a problem.

Now data could come from many sources that are less coordinated.

packages/shared/sdk-server/src/api/subsystems/LDTransactionalDataSourceUpdates.ts
kinyoklion: Ah, I don't think it is harmful in this case, but generally I want to check that all the types in an interface are exported. This alias isn't, but you can always use the original in this case.

Sometimes we end up accidentally exporting interfaces that don't use public types and then people have a hard time using them.

packages/shared/sdk-server/src/data_sources/OneShotInitializerFDv2.ts
tanderson-ld: Added `_stopped` to guard response processing code path.

packages/shared/sdk-server/src/data_sources/PollingProcessorFDv2.ts
tanderson-ld: Good catch.

packages/shared/sdk-server/src/data_sources/TransactionalDataSourceUpdates.ts
tanderson-ld: Unintentional duplicate.  Removed and pointed to existing impl.

packages/shared/sdk-server/__tests__/data_sources/PollingProcessorFDv2.test.ts
tanderson-ld: Updated all but the last test, which still has a vacuous test case.

packages/shared/sdk-server/src/data_sources/DataSourceUpdates.ts
```diff
@@ -31,7 +29,7 @@ interface TypeWithRuleClauses {
   ];
 }
 
-function computeDependencies(namespace: string, item: LDFeatureStoreItem) {
+export function computeDependencies(namespace: string, item: LDFeatureStoreItem) {
```
kinyoklion, line 32: suggestion (non-blocking): I would consider just putting this in `computeDependencies.ts`

kinyoklion (approved):

PR #834, opened 2025-04-25
feat: Add client-side SDK plugin support.

kinyoklion:
Adds plugin support for react native and browser.

Types are common, as well as most functionality, but SDK specific implementation is required. This is because the leaf-node SDKs have different types for the LDClient. Also the SDK needs control over the ordering of operations that happen during construction in order to register plugins at the correct point.

There is a bit more complexity in this implementation than would be required for most SDKs. Some refactoring of the base client implementation, potentially to use composition instead of inheritance, may allow for some better sharing here in the long-term. 

Initially I tried using generics in the base implementation, to account for the different client implementations, but this introduced a lot of changes to all the layers and didn't play well with configuration, which also then needs to be generic. 

The approach I took was to have the configuration in the individual SDKs and then use generics for utility methods to collect hooks and register plugins. In the long-term this would also benefit from the base implementation using generics for the configuration, which would allow a shared configuration and validation.

### DIFF
--- a/packages/sdk/browser/__tests__/BrowserClient.mocks.ts
+++ b/packages/sdk/browser/__tests__/BrowserClient.mocks.ts
@@ -7,6 +7,7 @@ import {
   SdkData,
 } from '@launchdarkly/js-client-sdk-common';
 
+import { BrowserOptions } from '../src/options';
 import { MockHasher } from './MockHasher';
 
 function mockResponse(value: string, statusCode: number) {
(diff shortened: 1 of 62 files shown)


packages/sdk/browser/__tests__/BrowserClient.mocks.ts
kinyoklion, line 1: Extracted from existing client tests to be re-used for plugin tests.

kinyoklion:
I noticed a problem with wrapperName/wrapperVersion in the SDKs. So I will make that a PR and it will be a prerequisite of this PR.

Vadman97 (approved):

PR #842, opened 2025-05-01
feat: Add support for an identify queue.

kinyoklion:


packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -255,47 +272,69 @@ export default class LDClientImpl implements LDClient {
...
-      (res) => {
-        afterIdentify({ status: 'completed' });
-        return res;
-      },
-      (e) => {
-        afterIdentify({ status: 'error' });
-        throw e;
-      },
-    );
+        return identifyPromise.then(
```
kinyoklion, line 309: In general I try to avoid promise chaining, but in this PR it is used to make it clear where things are happening synchronously. We synchronously return this promise that has some chaining that will resolve at some point. We don't actually want to await any of the process inside this method.

packages/shared/sdk-client/src/async/AsyncTaskQueue.ts
kinyoklion, line 1: Changed to a factory to save a few bytes.

kinyoklion:
> @launchdarkly/js-client-sdk size report This is the brotli compressed size of the ESM build. Compressed size: 21625 bytes Compressed size limit: 25000 Uncompressed size: 74100 bytes

After all the changes it only ended up being 625 bytes over between the hooks, plugins, identify queue, and per-context summary events.

kinyoklion:
Added @keelerm84 for thoughts on the before/after approach to the queue to handle synchronization with before/after identify. Not as an exclusive reviewer.

kinyoklion:
Looking into a problem with change notifications and shedding. It seems like a shed identify causes notifications to stop.

kinyoklion:
> Looking into a problem with change notifications and shedding. It seems like a shed identify causes notifications to stop.
This may just be the contexts I was using not having changes.

packages/shared/sdk-client/src/api/LDClient.ts
```diff
@@ -329,3 +334,39 @@ export interface LDClient {
...
+   * @param context
+   *    The LDContext object.
+   * @param identifyOptions
+   *    Optional configuration. Please see {@link LDIdentifyOptions}.
+   * @returns
+   *    A promise which resolves to an object containing the result of the identify operation.
+   *    The promise returned from this method will not be rejected.
+   */
+  identifyResult(
+    pristineContext: LDContext,
```
tanderson-ld, line 369: The word 'pristine' here might need a bit of explanation.

packages/shared/sdk-client/src/api/LDIdentifyOptions.ts
```diff
@@ -33,4 +33,15 @@ export interface LDIdentifyOptions {
    * not be required.
    */
   noTimeout?: boolean;
+
+  /**
+   * If true, the identify operation will be sheddable. This means that if multiple identify operations are done, without
```
tanderson-ld, line 38:
```suggestion
   * If true, the identify operation will be sheddable. This means that if multiple identify operations are started without
```

The word 'done' could be confused with finished.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -255,47 +278,86 @@ export default class LDClientImpl implements LDClient {
...
+    const callSitePromise = this._identifyQueue
+      .execute(
+        {
+          before: async () => {
+            let context = await ensureKey(pristineContext, this.platform);
+            if (this.autoEnvAttributes === AutoEnvAttributes.Enabled) {
+              context = await addAutoEnv(context, this.platform, this._config);
+            }
+            const checkedContext = Context.fromLDContext(context);
+            if (checkedContext.valid) {
```
tanderson-ld, line 290: I'm unsure if context validation here should prevent the before identify calls.  I could go either way, but not validation seems more logically sound.

tanderson-ld (approved):

packages/shared/sdk-client/src/LDClientImpl.ts
kinyoklion: Commenting for posterity.

The problem is that we cannot safely represent the invalid context with the current interface, so we don't execute before or after.

Secondary to that is that the information conveyed is not very useful in that case, it is basically that an attempt was made to identify with an invalid context.

PR #849, opened 2025-05-12
chore: now utilizes FDv2 basis param and supports FDv1 fallback

tanderson-ld:


packages/sdk/server-node/package.json
```diff
@@ -47,7 +47,7 @@
   "dependencies": {
     "@launchdarkly/js-server-sdk-common": "2.15.1",
     "https-proxy-agent": "^5.0.1",
-    "launchdarkly-eventsource": "2.1.0"
+    "launchdarkly-eventsource": "2.2.0"
```
tanderson-ld, line 50: For reviewers: This includes error header support needed for FDv2 fallback.

packages/shared/common/__tests__/internal/fdv2/PayloadStreamReader.test.ts
```diff
@@ -25,7 +25,7 @@ it('it sets basis to true when intent code is xfer-full', () => {
   });
 
   mockStream.simulateEvent('server-intent', {
-    data: '{"payloads": [{"code": "xfer-full", "id": "mockId"}]}',
+    data: '{"payloads": [{"intentCode": "xfer-full", "id": "mockId"}]}',
```
tanderson-ld, line 28: FDv2 spec change.

tanderson-ld:
Do not merge into target branch until target merges to main.

packages/shared/common/src/api/platform/Requests.ts
```diff
@@ -134,4 +134,5 @@ export interface Requests {
 export interface HttpErrorResponse {
   message: string;
   status?: number;
+  headers?: Record<string, string>;
```
kinyoklion, line 137: In the case of multiple values for a header is the second string some delimited form?

packages/shared/sdk-server/src/store/TransactionalFeatureStore.ts
```diff
@@ -70,31 +70,37 @@ export default class TransactionalFeatureStore implements LDTransactionalFeature
...
-                this._nonTransPersistenceStore.upsert({ namespace }, { key, ...item }, resolve);
-              }),
-            );
+          this._nonTransPersistenceStore.init(data, callback);
+        } else {
+          const promises: Promise<void>[] = [];
+          Object.entries(data).forEach(([namespace, items]) => {
+            Object.keys(items || {}).forEach((key) => {
+              const item = items[key];
+              promises.push(
```
kinyoklion, line 91: We will want to do some validation tests with real stores for this behavior.

packages/shared/common/src/api/platform/Requests.ts
tanderson-ld: Discussed in slack, going to leave use of Record<string, string> to not undo event source changes that already released.

packages/shared/sdk-server/src/store/TransactionalFeatureStore.ts
tanderson-ld: Revised to iterate using reduce since underlying persistent store wrapper was queueing anyways and I don't want this layer to assume the underlying store will enforce queueing.

packages/shared/common/src/api/platform/Requests.ts
tanderson-ld: It will follow the existing event source response header type, which I believe is delimited.

packages/shared/sdk-server/src/data_sources/PollingProcessorFDv2.ts
```diff
@@ -24,15 +25,24 @@ export default class PollingProcessorFDv2 implements subsystemCommon.DataSource
 
   private _statusCallback?: (status: subsystemCommon.DataSourceState, err?: any) => void;
 
+  /**
+   * @param _requestor to fetch flags from cloud services
```
kinyoklion, line 29:
```suggestion
   * @param _requestor to fetch flags
```

packages/shared/sdk-server/src/data_sources/PollingProcessorFDv2.ts
```diff
@@ -70,81 +88,151 @@ export default class PollingProcessorFDv2 implements subsystemCommon.DataSource
...
       }
-
-      try {
-        const parsed = JSON.parse(body) as internal.FDv2EventsCollection;
-        const payloadProcessor = new internal.PayloadProcessor(
-          {
-            flag: (flag: Flag) => {
-              processFlag(flag);
-              return flag;
+      if (body) {
```
kinyoklion, line 95: I am not sure I understand the `!body` path.

packages/shared/sdk-server/src/data_sources/PollingProcessorFDv2.ts
```diff
@@ -70,81 +88,151 @@ export default class PollingProcessorFDv2 implements subsystemCommon.DataSource
...
+    return [
+      {
+        key: 'basis',
+        value: selector,
+      },
+    ];
+  }
+
+  // helper function to transform FDv1 response data into events the PayloadProcessor can parse
+  private _processFDv1FlagsAndSegments(
```
kinyoklion, line 170: I feel like this should maybe just be in its own file.

kinyoklion: Or, if it doesn't use `this` at least hoisted out of the class to a free function.

packages/shared/sdk-server/src/data_sources/PollingProcessorFDv2.ts
```diff
@@ -70,81 +88,151 @@ export default class PollingProcessorFDv2 implements subsystemCommon.DataSource
...
 
       // schedule poll
       this._timeoutHandle = setTimeout(() => {
-        this._poll(dataCallback, statusCallback);
+        this._poll(dataCallback, statusCallback, selectorGetter);
       }, sleepFor);
+    }, this._selectorAsQueryParams(selectorGetter?.()));
+  }
+
+  private _selectorAsQueryParams(selector: string | undefined): { key: string; value: string }[] {
```
kinyoklion, line 156: I think this should probably be hoisted out of the class.

kinyoklion (approved):

PR #859, opened 2025-05-15
feat: Add support for per-context summary events.

kinyoklion:
Adds support for per-context summary events. Per-context summary events allow for enhanced analytics to know which contexts evaluate which flags, even when those flags are not part of an experiment.


packages/shared/common/__tests__/json/testdata/input/structures.json
kinyoklion, line 1: JSON files intentionally do not have newlines at the end for test purposes.

packages/shared/common/__tests__/json/canonicalize.test.ts
kinyoklion, line 1: We have a local canonicalization implementation, but we use standardized test data. Our implementation handles cycles.

.github/workflows/browser.yml
```diff
@@ -41,4 +41,4 @@ jobs:
           target_file: 'packages/sdk/browser/dist/index.js'
           package_name: '@launchdarkly/js-client-sdk'
           pr_number: ${{ github.event.number }}
-          size_limit: 21000
+          size_limit: 25000
```
kinyoklion, line 44: Both this PR and the client-side identify PR need a little breathing room. This PR exceeded the limit by 174 bytes.

tanderson-ld (approved):

Issue #868, opened 2025-05-22
[React Native] Inconsistent withReasons + initialConnectionMode behavior


artyorsh:
**Describe the bug**

`variationDetail` functions return `LDEvaluationDetailTyped` with reason = null, when `initialConnectionMode` = `polling`, regardless of `withReasons` being set to true.

**To reproduce**

Init the client with `withReasons: true` and `initialConnectionMode: 'polling'`, evaluate a flag, see the `reason` is null.

```ts
const options: LDOptions = {
  withReasons: true,
  initialConnectionMode: 'polling',
};

const ldClient = new ReactNativeLDClient(
  'mobile-key-123',
  AutoEnvAttributes.Enabled,
  options,
);

const variation: LDEvaluationDetailTyped<boolean> = ldClient.boolVariationDetail(flag, false);

console.log(variation.reason);
```

**Expected behavior**
If `withReasons` is set to true, variationDetail functions should return non-nullable reasons regardless of `initialConnectionMode` setting.

**SDK version**
@launchdarkly/react-native-client-sdk@10.10.1 (also seen in v10.9.2)

**Language version, developer tools**
Irrelevant.

**OS/platform**
iOS 18.0.0, Android 15 (API 35)

For instance, Ubuntu 16.04, Windows 10, or Android 4.0.3. If your code is running in a browser, please also include the browser type and version.

PR #869, opened 2025-05-22
fix: Requestor initialized with wrong withReasons and useReport config

artyorsh:
**Requirements**

**Related issues**

https://github.com/launchdarkly/js-core/issues/868

Provides Requestor configuration in correct order.

None.

**Additional context**

Fixes https://github.com/launchdarkly/js-core/issues/868

Edit: in functions with many arguments, consider arguments with an object, so that it's more clear what values are assigned to which keys.

### DIFF
--- a/packages/sdk/react-native/__tests__/MobileDataManager.test.ts
+++ b/packages/sdk/react-native/__tests__/MobileDataManager.test.ts
@@ -103,7 +103,9 @@ describe('given a MobileDataManager with mocked dependencies', () => {
           close: eventSourceCloseMethod,
         })),
         fetch: mockedFetch,
-        getEventSourceCapabilities: jest.fn(),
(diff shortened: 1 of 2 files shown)


kinyoklion:
@artyorsh Thank you for finding this! I will do some validation and then get it released.

artyorsh:
Ouch :) Why it fails?

I unfortunately don't have more time for the issue, hope it is incorrectly implemented test and there is enough context already.

kinyoklion:
> Ouch :) Why it fails?
> 
> I unfortunately don't have more time for the issue, hope it is incorrectly implemented test and there is enough context already.

Some of the builds require permissions that are not compatible with forks. We probably should disable those steps, but I will check the builds to make sure everything is fine.

kinyoklion (approved):

PR #895, opened 2025-07-15
feat: Adding Vercel AI SDK mapper

abarker-launchdarkly:


packages/sdk/server-ai/src/api/config/LDAIConfigTracker.ts
```diff
@@ -133,6 +133,55 @@ export interface LDAIConfigTracker {
...
+   * This function will track the duration of the operation, the token usage, and the success or error status.
+   *
+   * If the provided function throws, then this method will also throw.
+   * In the case the provided function throws, this function will record the duration and an error.
+   * A failed operation will not have any token usage data.
+   *
+   * @param func Function which executes the operation.
+   * @returns The result of the operation.
+   */
+  trackVercelAISDKStreamTextMetrics<
```
tracisiebel, line 172:
```suggestion
  trackVercelAIStreamTextMetrics
```
idk if we need the SDK part?

tracisiebel:
Do you have an example of how someone using the Vercel AI SDK would use this wrapper? If not - no worries - we can come up with something.

PR #903, opened 2025-07-31
feat: Improve KV data caching and deserialization

abarker-launchdarkly:
**Requirements**

Improve KV data caching and deserialization

### DIFF
--- a/packages/sdk/fastly/src/api/EdgeFeatureStore.ts
+++ b/packages/sdk/fastly/src/api/EdgeFeatureStore.ts
@@ -14,7 +14,7 @@ export interface EdgeProvider {
 
 export class EdgeFeatureStore implements LDFeatureStore {
   private readonly _rootKey: string;
-  private _kvData: string | null = null;
+  private _deserializedPromise: Promise<LDFeatureStoreDataStorage | null> | null = null;
 
   constructor(
     private readonly _edgeProvider: EdgeProvider,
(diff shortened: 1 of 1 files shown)


packages/sdk/fastly/src/api/EdgeFeatureStore.ts
```diff
@@ -106,13 +121,14 @@ export class EdgeFeatureStore implements LDFeatureStore {
...
-    const config = await this._getKVData();
-    const result = config !== null;
+    const deserialized = await this._getKVData();
+    const result = deserialized !== null;
     this._logger.debug(`Is ${this._rootKey} initialized? ${result}`);
     callback(result);
   }
 
   init(allData: LDFeatureStoreDataStorage, callback: () => void): void {
+    this._getKVData();
```
abarker-launchdarkly, line 131: For reviewers: I'm not entirely sure if this is necessary but wanted to play it safe.

packages/sdk/fastly/src/api/EdgeFeatureStore.ts
```diff
@@ -30,11 +30,36 @@ export class EdgeFeatureStore implements LDFeatureStore {
...
+        };
+        this._logger.debug('Successfully cached deserialized data');
+
+        return deserializedData;
+      })();
+    } else {
+      this._logger.debug('Using cached deserialized data');
     }
-    return this._kvData;
+    return this._deserializedPromise;
```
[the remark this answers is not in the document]
kinyoklion: I do think we need something to be a bit more robust in case of failure. Deserialization failure may not be recoverable, but does the KV api have any recoverable errors?

packages/sdk/fastly/src/api/EdgeFeatureStore.ts
kinyoklion: I would follow what we are doing with daemon mode, which I think it immediately initialized. As waitForInitialization would generally apply to cases where the evaluation result would be different if you waited versus not waiting. We are initialized from that perspective, but we haven't cached, which is the same as daemon mode.

Though it may be easier to explain to customers.

ldhenry (commented):
This looks good to me but I'll defer to @kinyoklion for final approval.

kinyoklion (approved):

Issue #907, opened 2025-08-18
EdgeFeatureStore incurs significant overhead


dferber90:
**Describe the bug**
EdgeFeatureStore's incur significant overhead which slows down feature flag evaluations.

At the moment the `@launchdarkly/js-server-sdk-common-edge` expects the bootstrap payload it receives from the feature store to be a string, which it then parses.

But some stores are already supply an object instead of a string.

These stores currently need to first call `JSON.stringify` only for the `@launchdarkly/js-server-sdk-common-edge` to then `JSON.parse` that string into an object again.

`JSON.stringify` and especially `JSON.parse` are very slow operations.

For example, this leads to evaluating 70 feature flags taking over 150ms (excluding any network requests). After patching the issue this goes down to less than 20ms.

**Details**

The issue is that @launchdarkly/vercel-server-sdk has [this line](https://github.com/launchdarkly/js-core/blob/7f0a54c4d880c8bc784fee0d4a2bc9155e96c1b7/packages/sdk/vercel/src/index.ts#L56) where it JSON.stringify’s the value read from our Edge Config store.

```
  const edgeProvider: EdgeProvider = {
    get: async (rootKey: string) => {
      const json = await edgeConfig.get(rootKey);
      return json ? JSON.stringify(json) : null;
    },
  };
```

It then later parses it again when calling `deserializePoll` [here](https://github.com/launchdarkly/js-core/blob/7f0a54c4d880c8bc784fee0d4a2bc9155e96c1b7/packages/shared/sdk-server-edge/src/api/EdgeFeatureStore.ts#L99), which internally calls `tryParse` which then calls `JSON.parse`.

**Expected behavior**
The `@launchdarkly/js-server-sdk-common-edge` should accept a plain object instead of having to be passed a string.

**SDK version**
- `@launchdarkly/js-server-sdk-common-edge` 2.6.8
- `@launchdarkly/vercel-server-sdk` 1.3.32

**Language version, developer tools**
TypeScript, JavaScript

**OS/platform**
Unrelated

**Additional context**
See the shared Slack channel you have with Vercel, where I posted about this as well.

PR #914, opened 2025-08-27
fix: Improve serialization/deserialization overhead with EdgeFeatureStore

abarker-launchdarkly:
**Requirements**

**Related issues**

https://github.com/launchdarkly/js-core/issues/907

Vercel's edge config already returns the JSON.parsed payload but it was being JSON.stringified and JSON.parsed an additional time which was adding overhead.  EdgeFeatureStore has been improved to be able to handle either stringified or parsed data so that this extra serialization/deserialization step doesn't need to occur.

### DIFF
--- a/packages/shared/sdk-server/src/store/serialization.ts
+++ b/packages/shared/sdk-server/src/store/serialization.ts
@@ -240,6 +240,21 @@ function tryParse(data: string): any {
 /**
  * @internal
  */
+export function reviveFullPayload(payload: FlagsAndSegments): FlagsAndSegments {
+  Object.values(payload?.flags || []).forEach((flag) => {
+    processFlag(flag);
+  });
+
+  Object.values(payload?.segments || []).forEach((segment) => {
+    processSegment(segment);
+  });
+
+  return payload;
+}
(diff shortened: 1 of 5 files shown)


packages/shared/sdk-server/src/store/serialization.ts
```diff
@@ -237,6 +237,27 @@ function tryParse(data: string): any {
...
+ * This function is intended for usage inside LaunchDarkly SDKs.
+ * This function should NOT be used by customer applications.
+ * This function may be changed or removed without a major version.
+ *
+ * @param payload Payload data from launchdarkly.
+ * @returns The revived and processed data.
+ */
+export function reviveFullPayload(payload: Record<string, any>): FlagsAndSegments {
+  const flagsAndSegments = payload as FlagsAndSegments;
+  Object.values(flagsAndSegments?.flags || []).forEach((flag) => {
```
tanderson-ld, line 250: Consider a for loop instead of forEach to avoid async risk in the future.  Though other iterations in this file are forEach, so maybe not worth the inconsistency.

PR #918, opened 2025-09-05
fix: improving payload processing overhead when payload does not change

abarker-launchdarkly:
**Requirements**

**Related issues**
https://github.com/launchdarkly/js-core/issues/907

Vercel's Edge Config SDK returns the same object reference if the payload has not changed.  A custom implementation of EdgeFeatureStore was added to take advantage of this and not re-reprocess the KV payload if it has not changed.

### DIFF
--- a/packages/sdk/vercel/__tests__/api/EdgeFeatureStore.test.ts
+++ b/packages/sdk/vercel/__tests__/api/EdgeFeatureStore.test.ts
@@ -1,0 +1,182 @@
+import { AsyncStoreFacade, LDFeatureStore } from '@launchdarkly/js-server-sdk-common-edge';
+import * as edgeExports from '@launchdarkly/js-server-sdk-common-edge';
+
+import { EdgeFeatureStore } from '../../src/api/EdgeFeatureStore';
+import mockEdgeProvider from '../utils/mockEdgeProvider';
+import * as testData from './testData.json';
+
+describe('EdgeFeatureStore', () => {
+  const sdkKey = 'sdkKey';
+  const kvKey = `LD-Env-${sdkKey}`;
+  const mockLogger = {
+    error: jest.fn(),
+    warn: jest.fn(),
+    info: jest.fn(),
+    debug: jest.fn(),
+  };
+  const mockGet = mockEdgeProvider.get as jest.Mock;
+  let featureStore: LDFeatureStore;
+  let asyncFeatureStore: AsyncStoreFacade;
+
+  beforeEach(() => {
+    featureStore = new EdgeFeatureStore(mockEdgeProvider, sdkKey, 'MockEdgeProvider', mockLogger);
+    asyncFeatureStore = new AsyncStoreFacade(featureStore);
+    mockGet.mockImplementation(() => Promise.resolve(testData));
+  });
+
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  describe('get', () => {
+    test('get flag', async () => {
+      const flag = await asyncFeatureStore.get({ namespace: 'features' }, 'testFlag1');
+
+      expect(mockGet).toHaveBeenCalledWith(kvKey);
+      expect(flag).toMatchObject(testData.flags.testFlag1);
+    });
+
+    test('invalid flag key', async () => {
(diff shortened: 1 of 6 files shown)


packages/sdk/vercel/__tests__/api/EdgeFeatureStore.test.ts
abarker-launchdarkly, line 1: For reviewers: This file is copied from js-server-sdk-common-edge and mostly unchanged except for the addition of the "get multiple flags" tests.

packages/sdk/vercel/__tests__/api/testData.json
abarker-launchdarkly, line 1: For reviewers: This file is copied exactly from js-server-sdk-common-edge

packages/sdk/vercel/src/index.ts
```diff
@@ -12,18 +12,19 @@ import type { EdgeConfigClient } from '@vercel/edge-config';
...
 
+import type { EdgeProvider } from './api';
+import { EdgeFeatureStore } from './api';
 import createPlatformInfo from './createPlatformInfo';
 
 export * from '@launchdarkly/js-server-sdk-common-edge';
 
-export type { LDClient };
+export { EdgeFeatureStore };
+export type { EdgeProvider, LDClient };
```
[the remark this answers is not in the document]
kinyoklion: Yes, there were some mistakes in the general export chain. Basically everything from `api` at each level is supposed to be exported ultimately. Unfortunately it technically requires major versions throughout the repository.

packages/sdk/vercel/src/api/EdgeFeatureStore.ts
```diff
@@ -0,0 +1,135 @@
...
+    kind: DataKind,
+    dataKey: string,
+    callback: (res: LDFeatureStoreItem | null) => void,
+  ): Promise<void> {
+    const { namespace } = kind;
+    const kindKey = namespace === 'features' ? 'flags' : namespace;
+    this._logger.debug(`Requesting ${dataKey} from ${this._rootKey}.${kindKey}`);
+
+    try {
+      const storePayload = await this._getStorePayload();
```
kinyoklion, line 38: Probably not applicable here, but `await` often ends up being much slower in the highly parallel case. Which is why the stores and the evaluator support end-to-end use of callbacks instead of just async await.

Just this one is potentially a small hit.

The original case this became a problem was the startup of some high traffic applications that would end up with like 1000 evals in parallel. With all promises the tasks interleaved and it took as long to get a single result as get all the results. Versus the callbacks where they resolve in approximately order because they end up executing synchronously.

packages/sdk/vercel/src/api/EdgeFeatureStore.ts
```diff
@@ -0,0 +1,135 @@
...
+  /**
+   * This method is used to retrieve the environment payload from the edge
+   * provider.
+   */
+  private async _getStorePayload(): Promise<
+    Exclude<ReturnType<typeof reviveFullPayload>, undefined>
+  > {
+    // Vercel Edge Config will return the same object reference if
+    // the payload has not changed.
+    const providerData = await this._edgeProvider.get(this._rootKey);
```
kinyoklion, line 88: I know the expectation here, but we should also validate that this function internally handles this well. In that if we start like 100 reads at the same time, when await a promise array, do we get the same reference for all of them. Basically the general problem of internally setting synchronous state.

If it doesn't then we probably should have an async queue like we use in persistent stores. Maybe long-term we should consider that anyway.

kinyoklion (approved):

packages/sdk/vercel/src/api/EdgeFeatureStore.ts
abarker-launchdarkly: Verified the behavior of the library and it will return the same reference in this case

abarker-launchdarkly: Discussed this and for now I'm leaving as-is since it follows the pattern from EdgeFeatureStore in js-server-sdk-common-edge.  However, this could be a consideration for change in the future if further performance enhancements are required.

PR #923, opened 2025-09-10
fix: app crashing due to undefined settings constant accessor

CalhamJNorthway:
The purpose of this PR is to fix a crash I found yesterday caused by an accessor that is undefined when newer versions of react native have Fabric enabled.

Our Build Setup:
`react-native: v0.80.0
`@launchdarkly/react-native-client-sdk: v10.1.5`

**Requirements**
- [x] add null coalescing operation to use `getConstants` to prevent undefined `settings` object
- [x] still default to old accessor if not undefined 

**Related issues**
I've not seen any related issues yet regarding your repo, though I did find this one [Stack Overflow post](https://stackoverflow.com/questions/79526197/cannot-read-property-applelocale-of-undefined).

The solution I've provided is to use a null coalescing check to fallback to the new `getConstants` method on native modules if the `settings` object is undefined.

I searched to see if there are preferred alternatives and did not come up with any other solutions. I am open to feedback on that though :)

### DIFF
--- a/packages/sdk/react-native/src/platform/locale.ts
+++ b/packages/sdk/react-native/src/platform/locale.ts
@@ -1,12 +1,15 @@
 import { NativeModules, Platform } from 'react-native';
-
 /**
- * Ripped from:
- * https://dev.to/medaimane/localization-and-internationalization-in-react-native-reaching-global-audiences-3acj
+ * Apps opted into Fabric (the new architecture of React Native) 
+ * may not have access to the SettingsManager.settings.AppleLocale property.
+ * It is now common to use the `getConstants` method to access these constant properties with Fabric enabled apps.
  */
-const locale =
-  Platform.OS === 'ios'
-    ? NativeModules.SettingsManager?.settings?.AppleLocale // iOS
-    : NativeModules.I18nManager?.localeIdentifier;
+const getAppleLocale = () => {
+  const settings = NativeModules.SettingsManager?.settings ?? NativeModules.SettingsManager?.getConstants()?.settings;
+  return settings?.AppleLocale;
+}
 
-export default locale;
+export default Platform.select({
+    ios: getAppleLocale(),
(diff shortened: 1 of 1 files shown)


CalhamJNorthway:
> @CalhamJNorthway Thank you for the PR!
> 
> What was the crash you are seeing? While I think that we need an update to reliably be accessing the locale, I don't see anything that was doing anything other than forwarding along the undefined value. Settings was undefined, but the app locale as already being conditionally accessed from it.
> 
> `NativeModules.SettingsManager?.settings?.AppleLocale`
> 
> Though there was a previous version where it looked like this: `NativeModules.SettingsManager?.settings.AppleLocale`
> 
> Which would have crashed.
> 
> Thank you, Ryan

Hey @kinyoklion, thanks for the prompt reply. I can add a little more detail, the issue was actually with the launch darkly client not instantiating correctly when the locale was undefined. You end up getting an error

`Error initializing LaunchDarkly: TypeError: Cannot read property 'ldApplication' of undefined`

Which after some further digging, I'm assuming was caused by the error below

`Error initializing LaunchDarkly: TypeError: Cannot read property 'AppleLocale' of undefined`

If you would like me to reproduce it with some screenshots I can do that tomorrow and post them here as a reply!

packages/sdk/react-native/src/platform/locale.ts
```diff
@@ -1,12 +1,15 @@
...
-    : NativeModules.I18nManager?.localeIdentifier;
+const getAppleLocale = () => {
+  const settings = NativeModules.SettingsManager?.settings ?? NativeModules.SettingsManager?.getConstants()?.settings;
+  return settings?.AppleLocale;
+}
 
-export default locale;
+export default Platform.select({
+    ios: getAppleLocale(),
+    android: NativeModules.I18nManager?.localeIdentifier,
```
joker23: change case to `default` to preserve original fallback behavior

PR #965, opened 2025-10-24
fix: Fix default configs always being disabled

jsonbailey:
> [!NOTE]
> Convert defaults to LD flag values (with _ldMeta.enabled) before calling variation, so configs/agents respect default enabled; update tests accordingly.
> 
> - **Server AI SDK**:
>   - **Default handling**: Add `_toLDFlagValue` to convert `defaultValue` into LD flag shape with `_ldMeta.enabled`, and use it in `_evaluate` so `enabled` mirrors the provided default when no flag metadata is present.
>   - **Refactor**: Rename local `provider` to `configProvider` in agent/config assembly to avoid shadowing.
> - **Tests**:
>   - Update expectations to assert `enabled` equals the provided default and that `variation` is called with the LD-shaped default (including `_ldMeta.enabled`).
>   - Adjust mocks for both config and agent paths to return LD-shaped values.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4be19fa2747aeebcad9b9f7eba87d8bf6548da7a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/sdk/server-ai/src/LDAIClientImpl.ts
+++ b/packages/sdk/server-ai/src/LDAIClientImpl.ts
@@ -69,7 +69,7 @@ export class LDAIClientImpl implements LDAIClient {
     return Mustache.render(template, variables, undefined, { escape: (item: any) => item });
   }
 
-  private _toLDFlagValue(defaultValue: LDAIDefaults | LDAIAgentDefaults): {
+  private static _toLDFlagValue(defaultValue: LDAIDefaults | LDAIAgentDefaults): {
     _ldMeta: { enabled: boolean };
     model?: LDModelConfig;
     messages?: LDMessage[];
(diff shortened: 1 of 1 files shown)


packages/sdk/server-ai/src/LDAIClientImpl.ts
```diff
@@ -69,12 +69,31 @@ export class LDAIClientImpl implements LDAIClient {
...
+    instructions?: string;
+  } {
+    return {
+      _ldMeta: { enabled: defaultValue.enabled ?? false },
+      model: defaultValue.model,
+      messages: 'messages' in defaultValue ? defaultValue.messages : undefined,
+      provider: defaultValue.provider,
+      instructions: 'instructions' in defaultValue ? defaultValue.instructions : undefined,
+    };
+  }
```
tanderson-ld: This could be static.  Does it need to be defined on LDAIClientImpl?

jsonbailey: Ideally this would be defined in api/config as a utility method but this could create a circular dependency. I wanted to avoid any breaking changes to the SDK for this. 

Current flow:

- api/agents → imports from api/config (line 1 in LDAIAgent.ts)
- api/config → would need to import from api/agents (for LDAIAgentDefaults)

This creates a cycle:

- api/agents → api/config → api/agents ❌

PR #1000, opened 2025-11-14
feat!: Only support identify with result.

kinyoklion:
Originally the SDK API had an `identify` method which didn't return any result. We added an `identifyResult` which is a safer API. 

We have decided that we shouldn't support the less-safe API and instead should have a single identify method. So the regular identify method will now have the behavior of `identifyResult`.

This moves the browser client to use a PIMPL implementation to decouple its interface of the inheritance hierarchy. Which allows for identify to have a completely different signature than the base implementation. It also allows us to retain identify in RN until we are ready to do a major version.

For the PIMPL implementation I decided to use a factory instead of a class. So tests needed updated to use that factory.

I also re-organized plugin registration so the better encapsulate PIMPL implementation would be passed to register.

The "compat" adapter needed to be updated to use the new API. For now it supports using the old interface, but we may want to consider removing it if we determine it isn't safe.

---

> [!NOTE]
> Replaces the BrowserClient class with a makeClient factory using a PIMPL impl and changes identify to return LDIdentifyResult (replacing identifyResult); updates plugin registration, compat, initialization, and tests.
> 
> - **API**
>   - `identify` now returns `LDIdentifyResult` (with statuses like `completed/shed/timeout/error`) and accepts browser-specific options; `identifyResult` is removed from the public interface.
>   - Introduce factory `makeClient` wrapping an internal `BrowserClientImpl` (PIMPL); `initialize` now uses `makeClient`.
>   - `LDClient` type updated to expose the new `identify` signature and omit the old one.
> - **Plugins**
>   - Refactor plugin registration to register against the public PIMPL client; hook retrieval unchanged; environment metadata still forwarded.
> - **Compat**
>   - `LDClientCompatImpl` updated to use `makeClient` and new `identify` result handling; passes `sheddable: false`, propagates `bootstrap/hash`, and maps failures/timeouts.
> - **Tests**
>   - Migrate tests to `makeClient`; adjust expectations for `identify` return values and shedding behavior; verify plugin registration and environment metadata.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4df7885d3a0902794ef57cd44655ac3371c2c0ba. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


kinyoklion:
bugbot review

PR #1019, opened 2025-12-01
chore: add an example app for browser sdk

joker23:
This PR will add a new browser SDK example that works with the new 4.x implementation.

I think this also addressed sdk-744

---

> [!NOTE]
> Adds a new browser SDK example app with build/run tooling and includes it in workspaces.
> 
> - **Example app** (`packages/sdk/browser/example`):
>   - **App code**: Adds `index.html`, `index.css`, and `src/app.ts` demonstrating `initialize`, `identify`, `variation`, and `change`/`error` listeners with simple UI updates.
>   - **Build/run**: Adds `package.json` scripts (`start`, `build`), `tsconfig.json`, and `tsdown.config.ts` that substitutes `LD_CLIENT_SIDE_ID`/`LD_FLAG_KEY` from `.env.template` into `dist/app.js`.
>   - **Docs**: Adds `README.md` with setup instructions.
> - **Monorepo**:
>   - Adds `packages/sdk/browser/example` to root `package.json` `workspaces`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4ba5b0b6287e7e18d57c4ca00f15b7c6f754585d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/sdk/browser/example/src/app.ts
```diff
@@ -0,0 +1,46 @@
+import { initialize } from '@launchdarkly/js-client-sdk';
+
+// Set clientSideID to your LaunchDarkly client-side ID
+const clientSideID = '';
```
kinyoklion: If we could set this up to work with a .env, or regular environment variable, then that would be good. 

Generally speaking this approach results in accidental commits of tokens. Client-side IDs are not secret, so it isn't a big issue, but still nice to not have to worry about it. 

If you do use a .env, then you can make sure it is in the git ignore.

kinyoklion: (We can do the same thing with the flagKey, for convenience.)

joker23: yea for this kind of example, I think we can inject those values during build time... I'll refactor this example, once we get the waitForInit PR in.

packages/sdk/browser/example/src/app.ts
```diff
@@ -0,0 +1,55 @@
...
+    ldclient.on('change', () => {
+      render();
+    });
+
+    const {status} = await ldclient.identify(context);
+    if (status === 'completed') {
+      render();
+    } else if (status === 'error') {
+      div.replaceChild(document.createTextNode('Error identifying client'), div.firstChild as Node);
+    }
```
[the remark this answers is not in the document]
joker23: yea I can get rid of the try catch now that we aren't throwing anything and I'll add a status box to track errors.

PR #1024, opened 2025-12-03
fix: bootstrapped client can do evaluation when not ready

joker23:
This commit will allow variation calls to use bootstrapped values during the client initialization process.

The bootstrap configuration will only be done once per client instance as we expect this fall back will only be active for a brief amount of time.

---

> [!NOTE]
> Enables pre-initialization flag evaluation by applying bootstrap flags once during identify, with a new shared setBootstrap hook.
> 
> - **Browser SDK (`packages/sdk/browser/src/BrowserClient.ts`)**
>   - Apply bootstrap during `identifyResult`: read flags via `readFlagsFromBootstrap`, call `setBootstrap(context, data)`, with a once-only guard (`_bootstrapAttempted`) and error logging.
> - **Shared SDK Client (`packages/shared/sdk-client/src/LDClientImpl.ts`)**
>   - Add protected `setBootstrap(pristineContext, newFlags)` to set context and initialize flags via `FlagManager.setBootstrap`.
>   - Import `ItemDescriptor` for bootstrap flag descriptors.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 15e59c706af8f08ffc3b7ff40be42ca96c2fe143. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/akamai-base.yml
+++ b/.github/workflows/akamai-base.yml
@@ -14,8 +14,8 @@ jobs:
   build-test-akamai:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6
-      - uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # v6
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+      - uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6.1.0
       - id: shared
         name: Shared CI Steps
         uses: ./actions/ci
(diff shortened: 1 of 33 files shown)


packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -360,6 +361,19 @@ export default class LDClientImpl implements LDClient, LDClientIdentifyResult {
...
+  /**
+   * Exposes the bootstrap functionality to the derived classes. This function is mainly used to help set
+   * flag values for bootstrapped clients before the initialization is complete.
+   *
+   * @param pristineContext The LDContext object to be identified.
+   * @param newFlags The new flags to set.
+   */
+  protected setBootstrap(pristineContext: LDContext, newFlags: { [key: string]: ItemDescriptor }) {
+    this._uncheckedContext = pristineContext;
+    this._checkedContext = Context.fromLDContext(this._uncheckedContext);
```
kinyoklion: I assume this will still get ultimately set by the normal code path, which would then handle any normal context processing? Do we need to be setting context information at all? I think in the previous SDK bootstrap would fail to send events before init complete, so this may be better for some cases.

joker23:
replaced by https://github.com/launchdarkly/js-core/pull/1036

PR #1036, opened 2025-12-12
feat: allow clients to evaluate bootstrapped flags when not ready

joker23:
**Related issues**

sdk-1653
sdk-1376
sdk-1663
sdk-1681

- Introduced `ActiveContextTracker` to manage the current active context and its serialized state.
- Updated `LDClientImpl` to utilize the new context tracker for identifying and evaluating flags.
- Added logic in `BrowserClientImpl` to read flags from bootstrap data during the initial identification process.
- Added a new `presetFlags` function in client sdk common that allow flagstore to take in contexless data before initialization
- Bootstrapped data will first `preset` the flag store so they can be evaluated before a full context is made
- Added logic to suppress event creation if context is not validated (while the client only has preset data)

**Additional context**
- eventually I would like to consolidate the context tracker with the tracking logic we have for waitForInitialization and have a general state tracker for client sdk common.
- I am also open to suggestions on distinguishing between "no context" state and "pre initialized" state. Though, it seems like, at the moment, those 2 states are handled in the same way.

Supersedes https://github.com/launchdarkly/js-core/pull/1024

---

> [!NOTE]
> Enable flag evaluation from bootstrap data before identify completes and refactor context handling via ActiveContextTracker with event suppression when no valid context.
> 
> - **Browser SDK**:
>   - Read bootstrap data in `BrowserClient.identifyResult` and `preset` flags for immediate evaluation; track first-identify via `_identifyAttempted`.
>   - Add test verifying flags evaluate from bootstrap before identify completes in `packages/sdk/browser/__tests__/BrowserClient.test.ts`.
> - **Core SDK (shared)**:
>   - Introduce `ActiveContextTracker` (`src/context/createActiveContextTracker.ts`) and update `LDClientImpl` to use it for getting/setting context and coordinating identify promises.
>   - Allow evaluations without an active context (log warning) and suppress event creation when context is absent; add `presetFlags` to `LDClientImpl`.
> - **Flag management**:
>   - Add `presetFlags` to `FlagManager`/`DefaultFlagManager` to initialize in-memory flags without persistence.
>   - Update `FlagUpdater` to track active `Context` (not just key) and adjust `init/initCached/upsert` and change callbacks accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b7b4bf8c29d22ec8a30d0ff948471dae7e6ed9fc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/shared/sdk-client/src/context/createActiveContextTracker.ts
+++ b/packages/shared/sdk-client/src/context/createActiveContextTracker.ts
@@ -5,17 +5,14 @@ import { Context, LDContext } from '@launchdarkly/js-sdk-common';
  * used by the client.
  */
 export interface ActiveContextTracker {
-  _pristineContext?: LDContext;
-  _context?: Context;
-
   /**
-   * Set the active context and pristine context. This will only be called when the passed in context
+   * Set the active context and unwrapped context. This will only be called when the passed in context
    * is checked and valid.
    *
-   * @param pristineContext - The pristine context, which is the context as it was passed in to the SDK.
+   * @param unwrappedContext - The unwrapped context, which is the context as it was passed in to the SDK.
    * @param context - The active context, which is the context as it was checked and validated.
    */
-  set(pristineContext: LDContext, context: Context): void;
+  set(unwrappedContext: LDContext, context: Context): void;
 
   /**
    * Get the active context.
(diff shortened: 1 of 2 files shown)


packages/shared/sdk-client/src/context/createActiveContextTracker.ts
```diff
@@ -0,0 +1,92 @@
...
+   * @returns The active context or undefined if it has not been set.
+   */
+  getContext(): Context | undefined;
+
+  /**
+   * Get the pristine context.
+   *
+   * @returns The pristine context or undefined if it has not been set.
+   */
+  getPristineContext(): LDContext | undefined;
```
kinyoklion: Looks like this context isn't really "pristine" as it has the generated keys at some point. I think we probably need to be a little more elaborate in either description or name.

kinyoklion (approved):

PR #1061, opened 2026-01-14
fix: evaluation detail type should be optional

joker23:
- Updates the common client typing so evaluation detail properties are optional (better compat)
- Removes the browser sdk compat module
- Moved the non-optional evaluation detail type to RN only (to avoid breaking changes)

---

> [!NOTE]
> Aligns evaluation detail typings and removes deprecated browser compat code.
> 
> - **Types:** Make `reason` optional in `LDEvaluationDetail` and `LDEvaluationDetailTyped` in `shared/sdk-client` for better compatibility
> - **React Native:** Add RN-specific `LDEvaluationDetail`/`LDEvaluationDetailTyped` (nullable `reason`) and re-export; update `useTypedVariationDetail` to use these types
> - **Browser SDK:** Remove the entire `src/compat` module, related tests, export entries, and build entry; delete upgrade docs; simplify `package.json` exports and `tsup` config
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8c8a043684813a89f114f6136562e2acbe82e0ac. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/sdk/react-native/src/hooks/variation/LDEvaluationDetail.ts
+++ b/packages/sdk/react-native/src/hooks/variation/LDEvaluationDetail.ts
@@ -1,16 +1,35 @@
-import { LDEvaluationDetailTyped as CommonLDEvaluationDetailTyped, LDEvaluationReason, } from '@launchdarkly/js-client-sdk-common';
+import {
+  LDEvaluationDetail as CommonDetail,
+  LDEvaluationDetailTyped as CommonDetailTyped,
+  LDEvaluationReason,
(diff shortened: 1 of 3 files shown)


packages/sdk/browser/src/compat/legacyTypes.ts
joker23: If we want to move forward with this idea... I'll try to minimize this file to only the relevant types needed for compat

PR #1064, opened 2026-01-16
fix: evaluation detail type should be optional

joker23:
This PR will:
- Updates the common client typing so evaluation detail properties are optional (better compat)
- Moved the non-optional evaluation detail type to RN only (to avoid breaking changes)

---

> [!NOTE]
> **Type updates (shared client)**
> 
> - `LDEvaluationDetail` and `LDEvaluationDetailTyped` now have optional `reason` (omit when absent) in `packages/shared/sdk-client/src/api/LDEvaluationDetail.ts`.
> - `createSuccessEvaluationDetail` emits `reason` only when provided; deleted/missing flags still return `variationIndex: null` and `value: undefined` (`evaluation/evaluationDetail.ts`, `LDClientImpl.ts`).
> - Tests updated to stop expecting `reason: null` and to accept details without `reason` where not present.
> 
> **React Native compatibility**
> 
> - Introduces RN-specific `LDEvaluationDetail`/`LDEvaluationDetailTyped` with `reason: LDEvaluationReason | null` in `hooks/variation/LDEvaluationDetail.ts`; re-exported from `src/index.ts`.
> - `useTypedVariationDetail` now normalizes `detail.reason ?? null` to satisfy RN nullable contract across all typed variation detail paths.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 509511637b1c59503bf50ffdced9066cc7ce109d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/shared/sdk-client/src/api/LDEvaluationDetail.ts
```diff
@@ -20,7 +18,7 @@ export type LDEvaluationDetail = Omit<CommonDetail, 'reason'> & {
   /**
    * An optional object describing the main factor that influenced the flag evaluation value.
    */
-  reason: LDEvaluationReason | null;
+  reason?: LDEvaluationReason | null;
```
kinyoklion: I would like to make `null` not a possibility. How problematic do we think that would be?

Can RN just ?? `null` it?

joker23: For RN, I think that could work, but that might cause problems if anyone is importing the typedef? I guess we can export the bad typedef and use the correct one in the internally?
> too paranoid?

PR #1244, opened 2026-04-03
feat: adding start() method to common client sdk package

joker23:
This PR will consolidate the `start()` method into the shared client sdk package.

To do this we also:
- introduce an internal option called `requiresStart` which should be `true` for new SDKs (and only RN has it default `false`)
- `requiresStart` is, effectively, a feature flag for whether the SDK is using the new `create` + `start` initialization pattern
- consolidated the unit tests for the `start` method into client common
- moved `start` method out of browser and electron sdks

---

> [!NOTE]
> **Medium Risk**
> Touches core initialization/identify behavior across multiple SDKs and changes when bootstrap parsing and goal tracking occur, which could affect startup sequencing and existing integrations if edge cases regress.
> 
> **Overview**
> Moves the `start()` implementation (including bootstrap flag preloading and init promise caching/timeout behavior) from the Browser and Electron SDKs into the shared `packages/shared/sdk-client` `LDClientImpl`.
> 
> Introduces `LDStartOptions` in the shared API and adds internal options `requiresStart` and `initialContext` so platform SDKs can enforce the new create+`start()` initialization pattern; `identifyResult()` now blocks pre-`start()` calls when `requiresStart` is enabled and updates the logged error message.
> 
> Updates Browser/Electron wrappers and tests accordingly (Browser/Electron remove their custom `start()` logic, Browser ensures goal tracking only starts after `start()`), and adds a consolidated `LDClientImpl.start()` test suite in the shared package.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5c9e61f90da54f819f108e846a4726f7b04cffc0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---

<a href="https://app.devin.ai/review/launchdarkly/js-core/pull/1244" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/packages/sdk/browser/__tests__/BrowserClient.test.ts
+++ b/packages/sdk/browser/__tests__/BrowserClient.test.ts
@@ -246,31 +246,6 @@ describe('given a mock platform for a BrowserClient', () => {
     expect(client.getContext()).toEqual({ kind: 'user', key: 'bob' });
   });
 
-  it('parses bootstrap data when start is called with bootstrap', async () => {
-    const client = makeClient(
-      'client-side-id',
-      { kind: 'user', key: 'bob' },
-      AutoEnvAttributes.Disabled,
-      {
-        streaming: false,
-        logger,
-        diagnosticOptOut: true,
-      },
-      platform,
-    );
-
-    await client.start({
-      identifyOptions: {
-        bootstrap: goodBootstrapDataWithReasons,
-      },
-    });
-
-    // Verify that bootstrap data was parsed and flags are available.
-    expect(client.allFlags().killswitch).toBe(true);
-    expect(client.allFlags()['string-flag']).toBe('is bob');
-    expect(client.allFlags().cat).toBe(false);
-  });
-
   it('uses the latest bootstrap data when identify is called with new bootstrap data', async () => {
     const initialBootstrapData = {
       'string-flag': 'is bob',
--- a/packages/sdk/browser/__tests__/BrowserClient.test.ts
+++ b/packages/sdk/browser/__tests__/BrowserClient.test.ts
@@ -1,8 +1,4 @@
-import {
-  AutoEnvAttributes,
-  LDLogger,
-  LDSingleKindContext,
-} from '@launchdarkly/js-client-sdk-common';
+import { AutoEnvAttributes, LDLogger } from '@launchdarkly/js-client-sdk-common';
 
 import { makeClient } from '../src/BrowserClient';
 import { makeBasicPlatform } from './BrowserClient.mocks';
@@ -181,7 +177,7 @@ describe('given a mock platform for a BrowserClient', () => {
     });
   });
 
-  it('can use bootstrap data', async () => {
+  it('can use bootstrap data via start()', async () => {
     const client = makeClient(
       'client-side-id',
       { kind: 'user', key: 'bob' },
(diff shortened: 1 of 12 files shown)


joker23:
@cursor review

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -302,8 +381,21 @@ export default class LDClientImpl implements LDClient, LDClientIdentifyResult {
     pristineContext: LDContext,
     identifyOptions?: LDIdentifyOptions,
   ): Promise<LDIdentifyResult> {
-    const identifyTimeout = identifyOptions?.timeout ?? DEFAULT_IDENTIFY_TIMEOUT_SECONDS;
-    const noTimeout = identifyOptions?.timeout === undefined && identifyOptions?.noTimeout === true;
+    if (this._requiresStart && !this.startPromise) {
+      this.logger.error(
+        'Client must be started before it can identify a context, did you forget to call start()?',
```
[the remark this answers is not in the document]
kinyoklion: Though something simpler than "subsequent" might be better.

I'm basically after 2 things:
1. Don't assume/imply they knew the correct operation.
2. If possible be clear that identify isn't for the first context. Only when changing contexts is required.

packages/sdk/browser/__tests__/BrowserClient.test.ts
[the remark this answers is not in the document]
joker23: okay, yea I think we should move the test refactors to a different PR and discuss there.

packages/sdk/browser/__tests__/BrowserClient.test.ts
```diff
@@ -246,10 +246,7 @@ describe('given a mock platform for a BrowserClient', () => {
     expect(client.getContext()).toEqual({ kind: 'user', key: 'bob' });
   });
 
-  it('parses bootstrap data only once when using start()', async () => {
-    const bootstrapModule = await import('@launchdarkly/js-client-sdk-common');
-    const readFlagsFromBootstrapSpy = jest.spyOn(bootstrapModule, 'readFlagsFromBootstrap');
-
+  it('parses bootstrap data when start is called with bootstrap', async () => {
```
kinyoklion: This makes me assume it parses it more than once? Which actually I think was already true? Once synchronously and again during the identify process? But what behavior change is this test change for?

joker23: This test should have been moved to the shared instance. I think the reason the test changed in this way is that there is no longer a way to test that behavior on this level.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -302,8 +381,21 @@ export default class LDClientImpl implements LDClient, LDClientIdentifyResult {
...
-    const noTimeout = identifyOptions?.timeout === undefined && identifyOptions?.noTimeout === true;
+    if (this._requiresStart && !this.startPromise) {
+      this.logger.error(
+        'The client must be started before a context can be identified. Call start() prior to identifying a context.',
+      );
+      return { status: 'error', error: new Error('Identify called before start') };
+    }
+
+    let effectiveOptions = identifyOptions;
+    if (this._requiresStart && identifyOptions?.sheddable === undefined) {
```
kinyoklion: I don't know that I like coupling of the default shedding behavior to the _requiresStart behavior.

packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -260,6 +270,75 @@ export default class LDClientImpl implements LDClient, LDClientIdentifyResult {
     this._flagManager.presetFlags(newFlags);
   }
 
+  /**
+   * Sets the initial context for the client. This must be called before `start()`.
+   * @param context The initial context.
+   */
+  setInitialContext(context: LDContext): void {
```
kinyoklion: Needs to be protected. React directly exposes the inheritance.

kinyoklion (changes requested):

packages/shared/sdk-client/src/LDClientImpl.ts
kinyoklion: This may present somewhat of a problem to the approach. So maybe we can figure out another approach.

packages/shared/sdk-client/src/LDClientImpl.ts
joker23: okay - yea I got a little too ambitious with trying to change the common default option to be `sheddable`.

joker23: so for this one, I decided to revert to the old defaults (`sheddable` is `false`) and have the browser and electron override that (which is what they did before)

packages/shared/sdk-client/src/LDClientImpl.ts
joker23: for this, I decided to get rid of this function altogether and introduced a new internal option that is `initialContext`

kinyoklion (approved):

PR #1289, opened 2026-04-17
feat: common identify now reports identify status

joker23:
This PR will consolidate the identify implementation in the sdk common level.

sdk-2078
sdk-2081
sdk-2082

The change here is that identify will always return a promise that resolves to a result. This pattern allows for better identify status tracking.

A few caveats:
- we will continue having react native conform to the old identify function signiture
- we have another work item to assess removing the proxy implementation that browser and electron sdk uses to compat the identify function

---

> [!NOTE]
> **Medium Risk**
> Changes the public `identify()` contract across the shared client and browser/electron SDKs from throwing/void to returning `LDIdentifyResult`, requiring downstream callers to handle `{status}` outcomes; behavior differences (timeouts/errors no longer reject) could break assumptions.
> 
> **Overview**
> **Unifies `identify()` to always return an `LDIdentifyResult`** (completed/error/timeout/shed) and guarantees the Promise does not reject, removing the separate `identifyResult()` API and updating the shared `LDClient` type accordingly.
> 
> Browser and Electron clients are updated to call/forward the new `identify()` (including IPC handlers and PIMPL facades) while still defaulting `sheddable: true` in their overrides; related tests are rewritten to assert returned status objects instead of promise rejections.
> 
> React Native preserves backward compatibility by exporting an RN-specific `LDClient` type with `identify(): Promise<void>` semantics and overriding `ReactNativeLDClient.identify()` to throw on error/timeout (and explicitly disallowing `start()`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b88b80557cdf91dbe82570ece663a3bcc18d3f49. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/sdk/react-native/src/ReactNativeLDClient.ts
+++ b/packages/sdk/react-native/src/ReactNativeLDClient.ts
@@ -231,15 +231,6 @@ export default class ReactNativeLDClient extends LDClientImpl {
     );
   }
 
-  /**
-   * Identifies a context to LaunchDarkly.
-   *
-   * This override preserves backward compatibility by throwing on error or timeout,
-   * matching the behavior consumers expect from the React Native SDK.
-   *
-   * @param context The LDContext object.
-   * @param identifyOptions Optional configuration. See {@link LDIdentifyOptions}.
-   */
   override async identify(
     context: LDContext,
     identifyOptions?: LDIdentifyOptions,
--- a/packages/sdk/react-native/src/ReactNativeLDClient.ts
+++ b/packages/sdk/react-native/src/ReactNativeLDClient.ts
@@ -20,6 +20,7 @@ import {
   type LDIdentifyResult,
   LDPluginEnvironmentMetadata,
   LDTimeoutError,
+  type LDWaitForInitializationResult,
   MOBILE_DATA_SYSTEM_DEFAULTS,
   MOBILE_TRANSITION_TABLE,
   mobileFdv1Endpoints,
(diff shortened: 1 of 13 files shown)


packages/sdk/react-native/src/ReactNativeLDClient.ts
```diff
@@ -227,6 +231,32 @@ export default class ReactNativeLDClient extends LDClientImpl {
...
+   * This override preserves backward compatibility by throwing on error or timeout,
+   * matching the behavior consumers expect from the React Native SDK.
+   *
+   * @param context The LDContext object.
+   * @param identifyOptions Optional configuration. See {@link LDIdentifyOptions}.
+   */
+  override async identify(
+    context: LDContext,
+    identifyOptions?: LDIdentifyOptions,
+  ): Promise<LDIdentifyResult> {
```
joker23, line 238: I think should be okay because I don't think anyone is checking for `void` returns?

joker23:
@cursor review

joker23:
@cursor review

joker23:
@cursor review

joker23:
@cursor review

packages/sdk/react-native/src/ReactNativeLDClient.ts
```diff
@@ -227,6 +232,33 @@ export default class ReactNativeLDClient extends LDClientImpl {
...
+    if (result.status === 'error') {
+      throw result.error;
+    } else if (result.status === 'timeout') {
+      const timeoutError = new LDTimeoutError(
+        `identify timed out after ${result.timeout} seconds.`,
+      );
+      this.logger.error(timeoutError.message);
+      throw timeoutError;
+    }
+    return result;
```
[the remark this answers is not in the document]
joker23: This is intentional

packages/sdk/react-native/src/ReactNativeLDClient.ts
kinyoklion: It doesn't make me particularly comfortable to actually be returning something.

joker23: yea, I don't think we could really do this step comfortably until we major version RN... which is fine... I'll move this back to draft or close, otherwise the changes seem fairly trivial.

PR #1304, opened 2026-04-21
fix: FDv2 Only -- Adjust the behavior of initialization when only cache initializers are available.

kinyoklion:
The main goal of this is to attach meta-data to cache initializer factories to allow for making some additional runtime decisions with regard to initialization.

In doing so I am changing the factory types from functions into an interface. Which will make this kind of change easier in the future. So the majority of actual change is from that adjustment.

---

> [!NOTE]
> **Medium Risk**
> Touches FDv2 data-source orchestration and factory typings; mistakes could change initialization resolution/rejection and status reporting (especially offline/cache-only scenarios). Scope is contained to FDv2 pipeline and is covered by expanded tests.
> 
> **Overview**
> **FDv2 initialization now treats cache-only configurations as valid even on cache miss.** `FDv2DataSource` detects when *all* initializers are cache-marked and there are no synchronizers, and in that case resolves initialization (optionally asserting `VALID`) instead of rejecting with “all sources exhausted”; it also avoids overwriting previously-reported error status and guards against `close()` races during the exhaustion path.
> 
> **Factory types were refactored from function types to `{ create(...) }` interfaces** (`InitializerFactory` adds optional `isCache`), requiring updates across `SourceManager`, `SourceFactoryProvider`, `FDv2DataManagerBase` (FDv1 fallback factory), and tests; `createCacheInitializerFactory` now returns an `isCache`-marked factory. CI package-size threshold was bumped from `38000` to `39000`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1104dc5d6c432ad90721877dedec2cddc4b26052. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/shared/sdk-client/src/datasource/fdv2/FDv2DataSource.ts
```diff
@@ -184,7 +195,8 @@ export function createFDv2DataSource(config: FDv2DataSourceConfig): FDv2DataSour
     }
 
     // All initializers exhausted.
-    if (dataReceived) {
+    if (dataReceived || cacheOnlyDataSystem) {
```
kinyoklion: Everything is basically to accomplish this.

kinyoklion:
bugbot review

kinyoklion:
bugbot review

kinyoklion:
bugbot review

packages/shared/sdk-client/src/datasource/fdv2/FDv2DataSource.ts
```diff
@@ -183,8 +200,30 @@ export function createFDv2DataSource(config: FDv2DataSourceConfig): FDv2DataSour
...
+    if (cacheOnlyDataSystem) {
+      // Cache-only data system with no synchronizer to produce a VALID
+      // status on its own. On a cache miss with no errors, nothing else
+      // has asserted VALID yet, so do it here. Skip the update if:
+      //   - dataReceived (cache hit): applyChangeSet already asserted VALID.
+      //   - errorReportedDuringInit: reportError set an error status that
+      //     must not be silently overwritten.
+      if (!dataReceived && !errorReportedDuringInit) {
+        statusManager.requestStateUpdate('VALID');
+      }
```
tanderson-ld, line 221: Do you want to report valid even if max cached contexts is 0 leading to cache miss in all cases?

tanderson-ld (approved):

joker23 (approved):

packages/shared/sdk-client/src/datasource/fdv2/FDv2DataSource.ts
kinyoklion: Unless I am mid-understanding I would not expect any difference between reasons the cache missed. It just missed.

tanderson-ld: You're correct, I thought about the interaction of the condition incorrectly.  All good.

PR #1312, opened 2026-04-22
fix: migrate anonymous context namespace to general namespace

joker23:
This PR will add logic to start migrating the old namespaces generated by `namespaceForAnonymousGeneratedContextKey` to the new general namespaces that are generated by `namespaceForGeneratedContextKey`.

We've deprecated the specific anonymous namespace for context keys for a while now, but never implemented a way to transition users to the new namespaces.

---

> [!NOTE]
> **Medium Risk**
> Changes how anonymous context keys are read/written in persistent storage by introducing an automatic migration path, which could affect identity continuity if storage behaviors differ across platforms. The migration includes a verification step before clearing legacy data, reducing but not eliminating risk of edge-case storage failures.
> 
> **Overview**
> Anonymous contexts now look up generated keys in the general `ContextKeys` namespace, with an automatic fallback migration from the legacy `AnonymousKeys` namespace.
> 
> `getOrGenerateKey` accepts an optional legacy storage key: if the new key is missing but a legacy key exists, it copies the value, verifies the write, and only then clears the legacy entry; otherwise it generates a new UUID as before. Tests were added/updated to cover migration, no-op/failed writes, and the “new key already exists” path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 35de7b0d0072a4ea500db3d7fbfe82dfcc8e621c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/shared/sdk-client/__tests__/context/ensureKey.test.ts
+++ b/packages/shared/sdk-client/__tests__/context/ensureKey.test.ts
@@ -102,4 +102,41 @@ describe('ensureKey', () => {
     const c = await ensureKey(context, mockPlatform);
     expect(c.key).toEqual('random1');
   });
+
+  it('should migrate anonymous key from legacy namespace', async () => {
+    const stored: Record<string, string> = {
+      LaunchDarkly_AnonymousKeys_org: 'migrated-key',
+    };
+    (mockPlatform.storage.get as jest.Mock).mockImplementation(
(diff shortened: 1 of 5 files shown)


joker23:
@cursor review

kinyoklion (approved):

PR #1315, opened 2026-04-23
chore: replace `prettier` with `@stylistic/eslint`

joker23:
This PR replacees `prettier` with `@stylistic/eslint-plugin` which seems to pull in less deps and is more flexible (less opinionated). This PR is mostly done for security reasons as `prettier` has history of supply chain attacks.

We went with this alternative because it consolidates styling back to linting (old school eslint) and we can configure the style in the same eslint configurations which is nice. More importantly, `@stylistic` feels like a more secure choice:

1. They provide provenance with their npm package (which is not the case with `prettier`
2. They do not have as many downstream dependencies

**NOTE** There will be a separate PR stacked on this one that will address all of the style violations that are caused by the change in configuration. We are also seeing that `prettier` actually hid some actual potential bugs in the code.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes lint/style tooling and `check` scripts across many workspaces, which can introduce new CI failures or formatting diffs even though it does not affect runtime code paths.
> 
> **Overview**
> Replaces the repo’s Prettier-based formatting workflow with ESLint-driven styling via `@stylistic/eslint-plugin`, and adds `eslint-plugin-simple-import-sort` for deterministic import ordering.
> 
> Removes Prettier config files (`.prettierrc`, `.prettierignore`) and drops `prettier`/`eslint-plugin-prettier`/`eslint-config-prettier`/`prettier-plugin-sort-imports` dependencies from the root and many package workspaces, updating package scripts to rely on `eslint --fix` (new/standardized `lint:fix`) and adjusting `check` pipelines accordingly.
> 
> Updates `.eslintrc.js` to include stylistic rule configuration, import-sort rules, and expanded ignore patterns (e.g. `node_modules`, `next-env.d.ts`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ee407aa05490aa05a797df48400054148a868423. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1316, opened 2026-04-23
refactor: fix lint issues

joker23:
This PR is the result of running `yarn lint:fix` with the configurations in https://github.com/launchdarkly/js-core/tree/skz/sdk-2234/remove-prettier

---

<a href="https://app.devin.ai/review/launchdarkly/js-core/pull/1316" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

---

> [!NOTE]
> **Medium Risk**
> Mostly mechanical lint/prettier autofixes (import/export ordering, ternary formatting, string literal normalization), but it also changes a few public entrypoints/exports ordering and adds `variationDetail` to `sdk-client`’s `LDClientImpl`, which could affect consumers relying on exact module shape or missing APIs.
> 
> **Overview**
> Applies repo-wide `yarn lint:fix` output: normalizes import ordering (including `type`-only imports), reformats conditional/ternary expressions, and replaces many template literals with plain string literals.
> 
> Also adjusts several package entrypoints to reorder/re-export symbols consistently (e.g. `index.ts` files across browser/electron/react-native/server/edge/ai providers), and updates a few runtime import positions (notably Shopify Oxygen timer polyfill).
> 
> Adds `variationDetail()` to `packages/shared/sdk-client`’s `LDClientImpl`, enabling detail-returning evaluations alongside existing `variation()`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 84065c1318cfe25afe64fc63dc2d02ab9e59475d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/sdk/server-node/src/platform/NodeRequests.ts
```diff
@@ -132,16 +132,15 @@ export default class NodeRequests implements platform.Requests {
     // Note this does not affect SSE as the event source is not using this fetch implementation.
     if (options.method?.toLowerCase() === 'get') {
       headers['accept-encoding'] = 'gzip';
-    }
```
kinyoklion, line 135: I cannot say I like this.

packages/sdk/shopify-oxygen/src/index.ts
```diff
@@ -1,9 +1,10 @@
+// Polyfill timer functions for Shopify Oxygen runtime
```
kinyoklion, line 1: Is this one only ESM so we can know import order doesn't matter? This seems like it would be a safe one.

I think we should call out any location where import order matters, but I am a bit paranoid because import order 100% matters for the observability stuff.

kinyoklion (approved):

packages/sdk/shopify-oxygen/src/index.ts
joker23: So I think this particular change is fine, the rule is in https://github.com/launchdarkly/js-core/pull/1315/changes#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5R59-R66

referencing https://github.com/lydell/eslint-plugin-simple-import-sort#custom-grouping

this makes sure the side-effect imports are done first which I generally agree with. I think if we need to customize import ordering beyond this then having the ignore comment is a good way to highlight (😄) that

PR #1330, opened 2026-04-28
feat: fall back to model.parameters.tools when root tools absent

jsonbailey:
## Summary

- Adds `_resolveTools()` to `LDAIConfigUtils` — checks root-level `tools` first, falls back to `model.parameters.tools` for older AI configs
- Root-level `tools` always takes priority when present
- `model.parameters.tools` as an array or non-object is silently ignored (may be user-defined custom param)
- Entries missing a `name` string are silently skipped; valid entries in the same map are still returned

## Background

The LaunchDarkly AI Config API now sends tools in two places: updated configs send them at both root and `model.parameters.tools`; older configs only send them in `model.parameters.tools`. Previously the SDK only read root-level, leaving older configs with no tools.

## Test plan
- [ ] Existing tools tests pass
- [ ] Tools parsed from `model.parameters.tools` when root key is absent
- [ ] Root `tools` wins when both locations are present
- [ ] `model.parameters.tools` as array → `undefined`
- [ ] Entries missing `name` skipped; valid entries in same map returned

Jira: AIC-1935

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Changes how `tools` are derived for agent/completion configs by adding a fallback to `model.parameters.tools` and new validation/logging, which could alter runtime tool availability for existing flags.
> 
> **Overview**
> Adds a new tool-resolution path for AI configs: `tools` are now taken from the root-level `tools` map when present, otherwise parsed from `model.parameters.tools` to support older flag shapes.
> 
> Tool parsing now validates that tools are object maps (not arrays/invalid types), normalizes entries (including defaulting missing `name` to the map key), and emits warnings via an optionally passed logger. Tests were expanded to cover the fallback, precedence rules, and invalid-shape handling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit edcee1df2c618a5484d8c3adeadeb170d6f355ea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/sdk/server-ai/src/api/config/LDAIConfigUtils.ts
+++ b/packages/sdk/server-ai/src/api/config/LDAIConfigUtils.ts
@@ -143,6 +143,41 @@ export class LDAIConfigUtils {
     }
   }
 
+  private static _resolveTools(
+    flagValue: LDAIConfigFlagValue,
+  ): { [toolName: string]: LDTool } | undefined {
+    if (flagValue.tools !== undefined) {
+      return flagValue.tools;
+    }
+
+    const rawTools = flagValue.model?.parameters?.['tools'];
+
+    if (rawTools === null || rawTools === undefined || Array.isArray(rawTools)) {
+      return undefined;
+    }
+
+    if (typeof rawTools !== 'object') {
+      return undefined;
+    }
+
+    const toolsMap = rawTools as { [key: string]: unknown };
+    const result: { [toolName: string]: LDTool } = {};
+
+    for (const [toolName, toolValue] of Object.entries(toolsMap)) {
+      if (
+        toolValue === null ||
+        typeof toolValue !== 'object' ||
+        Array.isArray(toolValue) ||
+        typeof (toolValue as { name?: unknown }).name !== 'string'
+      ) {
+        return undefined;
+      }
+      result[toolName] = toolValue as LDTool;
+    }
+
+    return result;
+  }
+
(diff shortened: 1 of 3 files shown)


packages/sdk/server-ai/src/api/config/LDAIConfigUtils.ts
```diff
@@ -143,6 +146,62 @@ export class LDAIConfigUtils {
...
+        logger?.warn(`LaunchDarkly AI: Skipping tool "${toolName}": expected an object`);
+        continue;
+      }
+      const toolObj = toolValue as { [key: string]: unknown };
+      result[toolName] = {
+        name: typeof toolObj['name'] === 'string' ? toolObj['name'] : toolName,
+        description:
+          typeof toolObj['description'] === 'string' ? toolObj['description'] : undefined,
+        type: typeof toolObj['type'] === 'string' ? toolObj['type'] : undefined,
+        parameters: toolObj['parameters'] as LDTool['parameters'],
```
joker23: nit: you may want to check if `toolObj['type']` is an `object` first?
> nit because this is internal and shouldn't run into bad base.

joker23 (approved):
1 nit comment

joker23 (approved):

PR #1331, opened 2026-04-28
feat: add Evaluator class for judge orchestration

jsonbailey:
## Summary
- Adds `Evaluator` class that wraps judges and `JudgeConfiguration`
- `Evaluator.noop()` static factory returns a no-op evaluator (resolves to `[]`)
- `evaluate(input, output)` runs all configured judges in parallel; missing judge key → warning + skip (not error)
- `Evaluator` does NOT call `tracker.trackJudgeResult` — that belongs in the managed layer
- Attaches `Evaluator` to `LDAICompletionConfig` and `LDAIAgentConfig` (populated in `createChat`/`createAgent`)

## Test plan
- [ ] All 188 existing tests pass
- [ ] New `Evaluator.test.ts` covers noop(), judge evaluation, missing judge warns+skips, error handling, no tracker calls

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Introduces new judge orchestration and changes how `createChat`/`createJudge` construct and wire judges, which can affect when/if evaluations run and how sampling is applied. Main risk is behavioral drift in judge execution paths due to the new evaluator attachment and updated `Judge` sampling semantics.
> 
> **Overview**
> Adds an internal `Evaluator` abstraction that runs all configured `Judge` instances in parallel and returns an array of `LDJudgeResult`s (with a `noop()` mode that always returns `[]`).
> 
> Refactors `LDAIClientImpl.createChat` to build and attach an `Evaluator` onto the returned `LDAICompletionConfig` (and updates config types to carry `evaluator`), instead of returning a populated judges map to `TrackedChat`; judge instance creation is centralized via a new `_createJudgeInstance` helper.
> 
> Updates `Judge` to support a constructor-level default sampling rate (exposed via `sampleRate`) and changes `evaluate`/`evaluateMessages` so `samplingRate` is optional and falls back to the constructor default. Adds/updates Jest coverage for `Evaluator`, the new `Judge` sampling behavior, and the updated `createJudge` constructor arguments.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 12f365734306eec1f38acf8530fb043f83c9454c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---

<a href="https://app.devin.ai/review/launchdarkly/js-core/pull/1331" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/sdk/server-ai/src/LDAIClientImpl.ts
+++ b/packages/sdk/server-ai/src/LDAIClientImpl.ts
@@ -22,6 +22,7 @@ import {
 } from './api/config';
 import { LDAIConfigFlagValue, LDAIConfigUtils } from './api/config/LDAIConfigUtils';
 import { AgentGraphDefinition, LDAgentGraphFlagValue, LDGraphTracker } from './api/graph';
+import { Evaluator } from './api/judge/Evaluator';
 import { Judge } from './api/judge/Judge';
 import { LDAIClient } from './api/LDAIClient';
 import { AIProviderFactory, SupportedAIProvider } from './api/providers';
@@ -168,6 +169,26 @@ export class LDAIClientImpl implements LDAIClient {
     });
 
     return judges;
+  }
+
+  private async _buildEvaluator(
+    judgeConfigs: LDJudge[],
+    context: LDContext,
+    variables?: Record<string, unknown>,
+    defaultAiProvider?: SupportedAIProvider,
+  ): Promise<Evaluator> {
+    if (judgeConfigs.length === 0) {
+      return Evaluator.noop();
+    }
+
+    const judgesRecord = await this._initializeJudges(
+      judgeConfigs,
+      context,
+      variables,
+      defaultAiProvider,
+    );
+    const judgesMap = new Map<string, Judge>(Object.entries(judgesRecord));
+    return new Evaluator(judgesMap, { judges: judgeConfigs }, this._logger);
   }
 
   private async _completionConfig(
@@ -318,14 +339,23 @@ export class LDAIClientImpl implements LDAIClient {
       return undefined;
     }
 
-    const judges = await this._initializeJudges(
+    const evaluator = await this._buildEvaluator(
       config.judgeConfiguration?.judges ?? [],
       context,
       variables,
       defaultAiProvider,
     );
 
-    return new TrackedChat(config, provider, judges, this._logger);
+    // Attach the evaluator to the config for use by the managed layer
+    const configWithEvaluator: LDAICompletionConfig = { ...config, evaluator };
+
+    // Build the legacy judges record for TrackedChat backward compat
+    const judges: Record<string, Judge> = {};
+    evaluator.judges.forEach((judge, k) => {
+      judges[k] = judge;
+    });
+
+    return new TrackedChat(configWithEvaluator, provider, judges, this._logger);
   }
 
   async createJudge(
--- a/packages/sdk/server-ai/src/LDAIClientImpl.ts
+++ b/packages/sdk/server-ai/src/LDAIClientImpl.ts
@@ -147,9 +147,7 @@ export class LDAIClientImpl implements LDAIClient {
     context: LDContext,
     variables?: Record<string, unknown>,
     defaultAiProvider?: SupportedAIProvider,
-  ): Promise<Record<string, Judge>> {
-    const judges: Record<string, Judge> = {};
-
+  ): Promise<Map<string, Judge>> {
     const judgePromises = judgeConfigs.map(async (judgeConfig) => {
       const judge = await this.createJudge(
         judgeConfig.key,
@@ -158,13 +156,14 @@ export class LDAIClientImpl implements LDAIClient {
         variables,
         defaultAiProvider,
       );
-      return judge ? { key: judgeConfig.key, judge } : null;
+      return judge ? ([judgeConfig.key, judge] as const) : null;
     });
 
     const results = await Promise.all(judgePromises);
-    results.forEach((result) => {
-      if (result) {
-        judges[result.key] = result.judge;
+    const judges = new Map<string, Judge>();
+    results.forEach((entry) => {
+      if (entry) {
+        judges.set(entry[0], entry[1]);
       }
     });
 
@@ -181,13 +180,12 @@ export class LDAIClientImpl implements LDAIClient {
       return Evaluator.noop();
     }
 
-    const judgesRecord = await this._initializeJudges(
+    const judgesMap = await this._initializeJudges(
       judgeConfigs,
       context,
       variables,
       defaultAiProvider,
     );
-    const judgesMap = new Map<string, Judge>(Object.entries(judgesRecord));
     return new Evaluator(judgesMap, { judges: judgeConfigs }, this._logger);
   }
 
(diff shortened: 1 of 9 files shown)


packages/sdk/server-ai/src/api/judge/Evaluator.ts
```diff
@@ -0,0 +1,80 @@
...
+ * `evaluate` method that runs all configured judges against a given input/output pair.
+ *
+ * The `Evaluator` is responsible only for running judges and returning results.
+ * It does NOT call `tracker.trackJudgeResult` — that is the responsibility of
+ * the managed layer (e.g., ManagedModel, ManagedAgent).
+ */
+export class Evaluator {
+  constructor(
+    readonly judges: Map<string, Judge>,
+    readonly judgeConfiguration: LDJudgeConfiguration,
```
joker23: I'm a bit confused, it looks like `LDJudgeConfiguration` already has an array of judges so why are we asking for both a map and configuration? Or is this a foreign key like situation?

packages/sdk/server-ai/src/LDAIClientImpl.ts
```diff
@@ -170,6 +171,26 @@ export class LDAIClientImpl implements LDAIClient {
     return judges;
   }
 
+  private async _buildEvaluator(
```
joker23, line 145: so does this mean that the `Evaluator` class is an internal implementation?

packages/sdk/server-ai/src/LDAIClientImpl.ts
```diff
@@ -170,6 +171,26 @@ export class LDAIClientImpl implements LDAIClient {
...
+    judgeConfigs: LDJudge[],
+    context: LDContext,
+    variables?: Record<string, unknown>,
+    defaultAiProvider?: SupportedAIProvider,
+  ): Promise<Evaluator> {
+    if (judgeConfigs.length === 0) {
+      return Evaluator.noop();
+    }
+
+    const judgesRecord = await this._initializeJudges(
```
joker23: Could this return a `Map<string, Judge>` directly? Or do we need to deep copy?

packages/sdk/server-ai/src/api/judge/Evaluator.ts
```diff
@@ -0,0 +1,80 @@
...
+        const result: LDJudgeResult = {
+          success: false,
+          sampled: true,
+          errorMessage: err instanceof Error ? err.message : 'Unknown error',
+        };
+        return result;
+      }
+    });
+
+    const settled = await Promise.allSettled(evaluationPromises);
```
joker23: Since we are doing a `try...catch` above, I wonder if we ever go into a rejection? if not the we can probably just use `Promise.all` (and I guess we will need to filter out `null` return case if judge is not in the map)

packages/sdk/server-ai/src/api/judge/Evaluator.ts
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Evaluates the given input/output pair using all configured judges.
+   * Missing judge instances are logged as warnings and skipped (not errors).
+   *
+   * @param input The input that was provided to the AI model.
+   * @param output The output produced by the AI model.
+   * @returns A promise resolving to an array of judge evaluation results.
+   */
+  async evaluate(input: string, output: string): Promise<LDJudgeResult[]> {
+    if (this.judgeConfiguration.judges.length === 0) {
```
joker23: nit: we want this case to do the same thing as the `noop` case right? if so then we may want to reference the same return value in case we decide to be more fancy with `noop` case.

packages/sdk/server-ai/src/api/config/types.ts
```diff
@@ -239,6 +245,11 @@ export interface LDAICompletionConfig extends LDAIConfig {
    * Root-level tools map keyed by tool name. Distinct from model.parameters.tools[].
    */
   tools?: { [toolName: string]: LDTool };
+  /**
+   * Evaluator for this completion config. Populated by createChat/createModel.
+   * Internal; not part of the flag value shape.
```
joker23: You can use TSDoc `@internal` marker and that should exclude this from the API docs

packages/sdk/server-ai/src/api/judge/Evaluator.ts
jsonbailey: The judges are the actual implementations of a Judge. Judges can be attached to multiple different ai configs via the judge configuration. The configuration holds the sample rates specific for the judges for the specific ai config. In practice I'm not sure we actually share implementations of judges so maybe it would be better to include the sample rate for each judge.

packages/sdk/server-ai/src/api/judge/Evaluator.ts
jsonbailey: The noop is a static builder for the evaluator which returns an evaluator. This is what an evaluator instance returns. I might not be following what you mean but we can chat about it.

packages/sdk/server-ai/src/api/judge/Evaluator.ts
joker23: ahh okay, I think that makes sense to me. So this is all part of the sdk initialization?

packages/sdk/server-ai/src/LDAIClientImpl.ts
```diff
@@ -318,14 +318,17 @@ export class LDAIClientImpl implements LDAIClient {
...
       context,
       variables,
       defaultAiProvider,
     );
 
-    return new TrackedChat(config, provider, judges, this._logger);
+    // Attach the evaluator to the config for use by the managed layer
+    const configWithEvaluator: LDAICompletionConfig = { ...config, evaluator };
+
+    return new TrackedChat(configWithEvaluator, provider, {}, this._logger);
```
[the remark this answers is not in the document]
jsonbailey: Acknowledged — this is intentional for this PR. `TrackedChat.invoke()` is already `@deprecated`; the correct evaluation path going forward is `run()`, which PR #1333 (stacked on #1332) wires up properly against the new `Evaluator`. The empty `{}` passed to the legacy `judges` constructor param is an accepted transitional state — callers on the deprecated path get the same "warning + empty results" behavior they'd see if `judgeConfiguration.judges` referenced any unknown key. Resolving.

packages/sdk/server-ai/src/LDAIClientImpl.ts
jsonbailey: For the record — the underlying concern here was addressed by the broader refactor: `_initializeJudges` (which returned a `Record` and required conversion to a `Map`) was deleted entirely. `_buildEvaluator` now constructs `Judge[]` directly via `_createJudgeInstance` and hands that array straight to `new Evaluator(judges, this._logger)`. No deep copies anywhere along the path. Leaving the resolve to you.

joker23 (approved):

packages/sdk/server-ai/src/LDAIClientImpl.ts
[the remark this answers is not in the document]
jsonbailey: This is intentional and will be resolved in a subsequent commit

PR #1342, opened 2026-04-30
fix(server-node): honor x-ld-fd-fallback directive in FDv2 initializer phase

keelerm84:
## Summary

- `OneShotInitializerFDv2` now applies any payload that arrives with `X-LD-FD-Fallback: true` before surfacing `LDFlagDeliveryFallbackError` (previously the directive was wrapped as a generic `LDPollingError` and silently dropped). `Requestor` propagates the directive to its callback so it rides along on 2xx, 3xx, and error responses.
- `CompositeDataSource` reacts to `LDFlagDeliveryFallbackError` immediately: it skips remaining initializers and FDv2 synchronizers, switches to the FDv1 fallback synchronizer without backoff, and transitions to a terminal Closed state when no FDv1 fallback is configured. `PollingProcessorFDv2` and `StreamingProcessorFDv2` honor the directive on success-paths the same way they already did on errors.
- Adds first-class `dataSystem.fdv1Fallback` config (optional `baseUri` + `pollInterval`) and wires it to the SDK's FDv1 fallback synchronizer in `LDClientImpl`. Contract test service declares the `fdv1-fallback` capability and accepts the new config object directly (no longer inferring it from the last synchronizer entry).

Mirrors the Go reference change in [go-server-sdk#365](https://github.com/launchdarkly/go-server-sdk/pull/365) and contract-test wiring in [go-server-sdk#368](https://github.com/launchdarkly/go-server-sdk/pull/368). Spec rationale lives in [sdk-specs#155](https://github.com/launchdarkly/sdk-specs/pull/155); the new harness suite is in [sdk-test-harness#336](https://github.com/launchdarkly/sdk-test-harness/pull/336).

## Test plan

- [x] `js-sdk-common`: 466/466 pass.
- [x] `js-server-sdk-common`: 979 pass / 5 pre-existing skips.
- [x] `node-server-sdk`: 63/63 pass.
- [x] `js-client-sdk-common`: 800/800 pass (regression check; sdk-client uses its own FDv2 implementation).
- [x] New deterministic tests for fallback on initializer success-with-payload, error, malformed body, polling success, streaming `onopen`, composite-level skip-and-switch, and `null` fdv1Fallback config.
- [ ] Contract-test workflow against `feat/fdv2` runs the new "FDv1 Fallback Directive" suite cleanly in CI.

---

> [!NOTE]
> **Medium Risk**
> Changes core data-source failover behavior and request handling to honor the server-directed `x-ld-fd-fallback` directive, which could affect initialization/streaming/polling transitions and retry/backoff behavior if incorrect. Covered by extensive new unit tests, but it still touches critical flag delivery paths.
> 
> **Overview**
> Ensures the server-side Node SDK **immediately honors the server-directed FDv1 fallback directive** (`x-ld-fd-fallback: true`) even during the FDv2 initializer phase, including cases where the directive arrives alongside a valid payload.
> 
> This updates `Requestor` to always surface a `fallbackToFDv1` signal (on success, 304, and error responses), updates FDv2 processors (`OneShotInitializerFDv2`, `PollingProcessorFDv2`, `StreamingProcessorFDv2`) to deliver the directive *atomically* with any payload (via a `fallbackToFDv1` marker) or emit `LDFlagDeliveryFallbackError` on error/parse failures, and updates `CompositeDataSource` to engage FDv1 fallback as a one-way switch (skip remaining initializers/FDv2 syncs, avoid re-engagement loops, bypass backoff only on initial engagement, and terminate if no FDv1 fallback is configured).
> 
> Adds first-class `dataSystem.fdv1Fallback` configuration (optional `baseUri`/`pollInterval`, nullable to opt out), wires it in `LDClientImpl` (including endpoint override support in `Requestor`), and updates contract-test wiring/capabilities plus expands test coverage across all affected components.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fe2770530b7a4adc5726b6038fb9499be1e51e07. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/shared/common/__tests__/subsystem/DataSystem/CompositeDataSource.test.ts
+++ b/packages/shared/common/__tests__/subsystem/DataSystem/CompositeDataSource.test.ts
@@ -650,6 +650,238 @@ it('terminates when FDv1 fallback is requested but no FDv1 fallback synchronizer
   expect(lastCall[0]).toBe(DataSourceState.Closed);
 });
 
+// Regression for the bug where a data source delivered basis-during-init AND THEN emitted
+// LDFlagDeliveryFallbackError as a status callback: the auto-transition on basis would
+// disable the callback handler before the fallback signal was processed, leaving
+// `_syncFactories` pointing at the FDv2 list. The fix is to ride the directive on the data
+// callback itself (via `data.fallbackToFDv1`), so the swap to FDv1 is atomic with the
+// payload application.
+it('switches to FDv1 when an initializer delivers basis with fallbackToFDv1 marker', async () => {
+  const mockInitData = { fallbackToFDv1: true, payload: { key: 'init-payload' } };
+  const mockInitializer = {
+    start: jest
+      .fn()
+      .mockImplementation(
+        (
+          _dataCallback: (basis: boolean, data: any) => void,
+          _statusCallback: (status: DataSourceState, err?: any) => void,
+        ) => {
+          _statusCallback(DataSourceState.Initializing);
+          _statusCallback(DataSourceState.Valid);
+          // Basis arrives with the directive attached. The composite must apply the payload
+          // (via dataCallback) AND swap synchronizer factories to FDv1 atomically; the
+          // FDv2 synchronizer below must NOT be started.
+          _dataCallback(true, mockInitData);
+        },
+      ),
+    stop: jest.fn(),
+  };
+
+  const mockFDv2Synchronizer = {
+    start: jest.fn().mockImplementation(() => {
+      throw new Error('FDv2 synchronizer should not be started after FDv1 fallback');
+    }),
+    stop: jest.fn(),
+  };
+
+  const mockFDv1Data = { key: 'FDv1Data' };
+  const mockFDv1Synchronizer = {
+    start: jest
+      .fn()
+      .mockImplementation(
+        (
+          _dataCallback: (basis: boolean, data: any) => void,
+          _statusCallback: (status: DataSourceState, err?: any) => void,
+        ) => {
+          _statusCallback(DataSourceState.Initializing);
+          _statusCallback(DataSourceState.Valid, null);
+          _dataCallback(true, mockFDv1Data);
+        },
+      ),
+    stop: jest.fn(),
+  };
+
+  const underTest = new CompositeDataSource(
+    [makeDataSourceFactory(mockInitializer)],
+    [makeDataSourceFactory(mockFDv2Synchronizer)],
+    [makeDataSourceFactory(mockFDv1Synchronizer)],
+    undefined,
+    makeTestTransitionConditions(),
+    makeZeroBackoff(),
+  );
+
+  const dataReceived: any[] = [];
+  const statusCallback = jest.fn();
+  await new Promise<void>((resolve) => {
+    const dataCallback = jest.fn((_: boolean, data: any) => {
+      dataReceived.push(data);
+      if (data === mockFDv1Data) {
+        resolve();
+      }
+    });
+    underTest.start(dataCallback, statusCallback);
+  });
+
+  // The composite forwarded the initializer's payload AND then ran the FDv1 synchronizer.
+  // FDv2 synchronizers must not have been started.
+  expect(mockInitializer.start).toHaveBeenCalledTimes(1);
+  expect(mockFDv2Synchronizer.start).not.toHaveBeenCalled();
+  expect(mockFDv1Synchronizer.start).toHaveBeenCalledTimes(1);
+
+  // The init payload was forwarded to the outer dataCallback before FDv1 took over.
+  expect(dataReceived[0]).toBe(mockInitData);
+  expect(dataReceived[1]).toBe(mockFDv1Data);
+});
+
+// Regression for the bug where a synchronizer delivered a payload with fallbackToFDv1 AND
+// THEN emitted an error: the basis-during-init auto-transition (or the error status path)
+// would disable the callback handler before the fallback signal was processed.
+it('switches to FDv1 when a synchronizer delivers a payload with fallbackToFDv1 marker', async () => {
+  const mockSynchronizerData = { fallbackToFDv1: true, payload: { key: 'sync-payload' } };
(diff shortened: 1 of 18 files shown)


packages/shared/sdk-server/src/data_sources/OneShotInitializerFDv2.ts
```diff
@@ -89,11 +115,23 @@ export default class OneShotInitializerFDv2 implements subsystemCommon.DataSourc
 
         payloadProcessor.processEvents(parsed.events);
 
+        // Any accompanying payload has been applied. Honor the fallback directive (if any)
+        // before declaring the initializer Closed so the composite swaps to FDv1.
+        if (fallbackToFDv1) {
+          emitFallback();
+          return;
+        }
```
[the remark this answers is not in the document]
joker23: I _think_ this is valid. We need to be calling this before `dataCallback`

packages/shared/common/src/datasource/CompositeDataSource.ts
```diff
@@ -144,13 +144,30 @@ export class CompositeDataSource implements DataSource {
...
-                  // this error indicates we should fallback to only using FDv1 synchronizers
-                  if (err instanceof LDFlagDeliveryFallbackError) {
-                    this._logger?.debug(`Falling back to FDv1`);
-                    this._syncFactories = this._fdv1Synchronizers;
+                // The FDv1 fallback directive is server-directed and one-way. As soon as we
+                // observe it, end any in-progress initializer phase and replace the FDv2
+                // synchronizers with the configured FDv1 fallback list. If no FDv1 fallback
+                // is configured, the synchronizer list becomes empty and the composite will
+                // terminate via the exhausted-sources path.
+                if (err instanceof LDFlagDeliveryFallbackError) {
```
joker23: I might be misunderstanding the flow here, but if we are only conditioning off of this `LDFlagDeliveryFallbackError` and this error is conditioned off of the presence of the fallback header, wouldn't we go into an infinite loop unless subsequent responses do not have the fallback header?

packages/shared/sdk-server/src/api/options/LDDataSystemOptions.ts
```diff
@@ -68,6 +68,34 @@ export interface LDDataSystemOptions {
...
+export interface FDv1FallbackConfiguration {
+  /**
+   * Override the polling base URI used by the FDv1 Fallback Synchronizer. Defaults to the
+   * SDK's configured polling base URI.
+   */
+  baseUri?: string;
+  /**
+   * The interval between polls, in seconds. Defaults to the SDK's configured pollInterval.
+   */
+  pollInterval?: number;
```
joker23, line 98: we should add a validator for this (eg  https://github.com/launchdarkly/js-core/blob/62bb942cd86c15aca00f82c430739eb74740e04d/packages/shared/sdk-server/src/options/Configuration.ts#L58)

packages/shared/common/src/datasource/CompositeDataSource.ts
```diff
@@ -123,6 +128,31 @@ export class CompositeDataSource implements DataSource {
...
+                  );
+                }
+                this._fdv1FallbackEngaged = true;
+                this._syncFactories = this._fdv1Synchronizers;
+                callbackHandler.disable();
+                this._consumeCancelToken(cancelScheduledTransition);
+                sanitizedStatusCallback(DataSourceState.Interrupted);
+                transitionResolve({ transition: 'switchToSync' });
+                return;
+              }
```
joker23, line 162: nit: this block seems to be the same as:
https://github.com/launchdarkly/js-core/pull/1342/changes/77ee47a151985299a93499f0ac9b2ae2041c530c..8fb496fe1cd4efcde6bc40eb2b662e8c2e612cd4#diff-2bca0f4040acef8d56bcb93808d774c06280128e9ab9f34131088b0b3133b8a0R187-R203

I think it would be easier to maintain if abstracted out the logic.

packages/shared/common/__tests__/subsystem/DataSystem/CompositeDataSource.test.ts
```diff
@@ -650,6 +650,238 @@ it('terminates when FDv1 fallback is requested but no FDv1 fallback synchronizer
...
+  expect(mockFDv1Synchronizer.start).toHaveBeenCalledTimes(1);
+  expect(dataReceived[0]).toBe(mockSynchronizerData);
+  expect(dataReceived[1]).toBe(mockFDv1Data);
+});
+
+// Regression: subsequent fallback signals after the FDv1 synchronizer is engaged must be
+// ignored so the composite does not loop replacing its synchronizer list with itself.
+// This guards against an FDv1 endpoint that erroneously echoes `x-ld-fd-fallback` -- the
+// circular DataSourceList for `_fdv1Synchronizers` would otherwise restart the same
+// FDv1 synchronizer indefinitely after each repeated directive.
```
joker23: nit: I think we should do a round of making the comments more concise. For these test comments in particular, I think we should drop the "regression" tag as there is no regression ... yet :)

packages/shared/sdk-server/src/data_sources/PollingProcessorFDv2.ts
joker23, line 160: would there be a case where we are trying to fallback to fdv1, but this `statuCallback` is triggered before we have a change to set the value in `data`? Same concerns for the one shot initializer. I think that would be rare?

If this happens, then the fallback will attempt to choose a FDv2 processor right?

joker23 (approved):
lgtm

packages/shared/common/__tests__/subsystem/DataSystem/CompositeDataSource.test.ts
keelerm84: Done — dropped the "regression" / "Bug #" framing across all the touched test files in 7e2a58292, and applied the same tightening pass to the new backoff test in fe2770530. Comments now describe the contract being asserted rather than the absent shipped bug.

PR #1373, opened 2026-05-08
chore: init `@launchdarkly/client-testing-plugin`

joker23:
This PR pushes the initial implementation of the client test data plugin

SDK-2124

---

> [!NOTE]
> **Low Risk**
> New dev-only tooling with no production SDK behavior changes; risk is limited to monorepo build/test wiring for the new package.
> 
> **Overview**
> Introduces **`@launchdarkly/client-testing-plugin`**, a new workspace package that exposes **`TestData`** as an SDK plugin for tests and local dev. It drives the existing **`registerDebug` / `LDDebugOverride`** path so flag values can be seeded, updated (with **`change:`** events), removed, or cleared **without network I/O**.
> 
> The package ships with README usage docs, **tsup** build output (ESM/CJS + types), Jest setup (jsdom stubs for **EventSource** and **crypto.subtle**), unit tests against a mock debug override, and browser SDK integration tests. The monorepo wires the package into root **`package.json`** workspaces and **`tsconfig.json`** references; **`.gitignore`** swaps **`.claude/stacks`** for **`.claude/tmp`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c2fa0f634b6376c338d3aa9ccf360591b9db11fc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/tooling/client-testing-plugin/README.md
+++ b/packages/tooling/client-testing-plugin/README.md
@@ -1,0 +1,72 @@
+# LaunchDarkly Client Testing Plugin
+
+A testing plugin for LaunchDarkly client-side JavaScript SDKs. Use it to inject deterministic flag values into a real SDK client during unit tests, integration tests, and local development.
+
+## Install
+
+```bash
+yarn add --dev @launchdarkly/client-testing-plugin @launchdarkly/js-client-sdk
+```
+
+## Usage
+
+```ts
+import { createClient } from '@launchdarkly/js-client-sdk';
+import { TestData } from '@launchdarkly/client-testing-plugin';
+
+// Seed with a base set of flag values.
+const td = new TestData({
+  'new-ui': true,
+  greeting: 'Hello!',
+});
+
+const client = createClient(
+  '<ldClientSideId>', // placeholder -- fill in only for real environments
+  { kind: 'user', key: 'tester' },
+  {
+    plugins: [td],
+    sendEvents: false,
+    streaming: false,
+  },
+);
+
+await client.start({ bootstrap: {} });
+
+client.boolVariation('new-ui', false);          // true
+client.stringVariation('greeting', '(default)'); // 'Hello!'
+
+// Update flags at any time -- the SDK fires change events. Setters chain.
+td.setBool('new-ui', false).setString('greeting', 'Welcome');
+```
+
+### Why these options matter
+
+- **`plugins: [td]`** -- registers the testing plugin so it can inject overrides.
+- **`sendEvents: false`** -- keeps analytics events off in tests.
+- **`streaming: false`** -- prevents the SDK from auto-starting a streaming connection when a `change` listener is registered. Without this, the React SDK provider (and any other code that registers `change` listeners) will trigger a real network call to `clientstream.launchdarkly.com`.
+- **`bootstrap: {}` (passed to `start()`)** -- gives the SDK an empty initial flag set so it does not block on a network identify call. The plugin's overrides are applied immediately afterward.
+
+If you forget any of these, the SDK may attempt to fetch flags from LaunchDarkly during initialization and produce real network traffic, console errors, or stray evaluation events.
+
+## API
+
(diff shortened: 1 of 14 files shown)


joker23:
@cursor review

joker23:
@cursor review

joker23:
@cursor review

joker23:
@cursor review

joker23:
@cursor review

packages/tooling/client-testing-plugin/README.md
```diff
@@ -0,0 +1,72 @@
...
+await client.start({ bootstrap: {} });
+
+client.boolVariation('new-ui', false);          // true
+client.stringVariation('greeting', '(default)'); // 'Hello!'
+
+// Update flags at any time -- the SDK fires change events. Setters chain.
+td.setBool('new-ui', false).setString('greeting', 'Welcome');
+```
+
+### Why these options matter
```
kinyoklion: This is a strange heading.

packages/tooling/client-testing-plugin/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+- **`plugins: [td]`** -- registers the testing plugin so it can inject overrides.
+- **`sendEvents: false`** -- keeps analytics events off in tests.
+- **`streaming: false`** -- prevents the SDK from auto-starting a streaming connection when a `change` listener is registered. Without this, the React SDK provider (and any other code that registers `change` listeners) will trigger a real network call to `clientstream.launchdarkly.com`.
+- **`bootstrap: {}` (passed to `start()`)** -- gives the SDK an empty initial flag set so it does not block on a network identify call. The plugin's overrides are applied immediately afterward.
+
+If you forget any of these, the SDK may attempt to fetch flags from LaunchDarkly during initialization and produce real network traffic, console errors, or stray evaluation events.
```
kinyoklion: I think I would lean toward having a helper function in the test data package?

`const client = createTestClient({kind: 'user', key: 'tester'});`

Maybe optionally allowing options, but making it clear that the specific settings are overwritten.

`const client = createTestClient({kind: 'user', key: 'tester'}, {/* maybe custom options*/}`

Certainly we would support the version you demonstrate as well, but I think most people just need a client.

kinyoklion: Though I think you were trying to make one of these work for all client-side? Makes me a little uncomfortable not having leaf-nodes where we can customize and change decisions if we need to.

packages/tooling/client-testing-plugin/src/TestData.ts
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  /**
+   * @internal
+   *
+   * Shared write path for the typed setters. Stores the value, then fires
+   * `setOverride` unless this is a no-op primitive write (same key, same
+   * primitive value as before). Object/array writes always fire.
+   */
+  private _set(key: string, value: LDFlagValue): this {
```
kinyoklion, line 171: Why do we need this same value check stuff and duplication of the values? There are ways for a real connection to trigger updates even though these fields didn't change.

kinyoklion (approved):

packages/tooling/client-testing-plugin/README.md
joker23: I think providing a wrapped client creation function could help... I think we should still export a more manual way of getting this functionality.

The use case I am thinking about is developers using this plugin for their mocked LD clients. I think that would be a cleaner mock client implementation.

With that said, I think if we provide a per sdk wrapper, then we can make the "manual" path a bit more low level... I'll do a pass on the README now

packages/tooling/client-testing-plugin/src/TestData.ts
joker23: Good point, I will remove this check...

PR #1393, opened 2026-05-26
chore: Add Node platform abstraction for node-client

joker23:
## Summary
- Adds the Node platform abstraction layer for `@launchdarkly/node-client-sdk`: `HeaderWrapper`, `NodeCrypto`, `NodeEncoding`, `NodeInfo`, `NodePlatform`, `NodeRequests`, `NodeResponse`, `NodeStorage`. These implement the `Platform` contract from `@launchdarkly/js-client-sdk-common`.
- Adds runtime dependencies the platform code imports: `@launchdarkly/js-client-sdk-common`, `https-proxy-agent`, `launchdarkly-eventsource`.
- Modernizes `packages/sdk/node-client/tsconfig.json` to `module: ESNext` + `moduleResolution: bundler` (tsup handles emit), adds `esModuleInterop`, `types: ["jest", "node"]`, `skipLibCheck`. Matches the pattern used by `packages/sdk/server-ai`.
- Part of the SDK-2195 stacked migration of `launchdarkly-node-client-sdk` into js-core. Follows PR #1352 (scaffold).

---

> [!NOTE]
> **Medium Risk**
> New TLS, HTTP streaming, and on-disk cache behavior affect runtime networking and local persistence, though it is isolated behind platform interfaces and covered by unit tests; the public client API is not wired yet.
> 
> **Overview**
> Adds the **Node platform layer** for `@launchdarkly/node-client-sdk` so the package can plug into `@launchdarkly/js-client-sdk-common`: `NodePlatform` wires **info**, **crypto**, **encoding**, **file-backed storage**, and **HTTP/HTTPS + EventSource** requests. New `NodeOptions` covers **TLS** (`tlsParams`), optional **POST gzip** for events, and **`localStoragePath`** (default `ldclient-user-cache`).
> 
> `NodeRequests` / `NodeResponse` implement fetch-like behavior (timeouts, gzip on GET, optional body compression, custom CA for HTTPS) via `launchdarkly-eventsource`. `NodeStorage` persists a JSON cache with atomic writes and a process singleton.
> 
> Also adds **Jest** coverage for the platform modules, a **`sdk/node-client` GitHub Actions** job (Node 18 & 22), package **test** script and dependencies, **tsconfig** shift to `ESNext` / `bundler`, and **release-please** bumping `NodeInfo.ts` for version stamps.
> 
[...]

PR #1403, opened 2026-06-01
chore(node-client): harden existing platform implementation

joker23:
Hardens the Node platform layer landed in #1393, ahead of the client-side SDK implementation that builds on it. First atomic in the node-client port stack.

  - **NodeStorage**: log (no longer swallow) flush errors; validate/log on malformed cache reads and ignore non-string cache values; write the temp file via an exclusive `wx` open so the write cannot be redirected through a symlink; warn when `getNodeStorage` is called with a `localStoragePath` that differs from the process singleton's.

---

> [!NOTE]
> **Medium Risk**
> Changes on-disk cache semantics and filesystem write behavior; low blast radius but incorrect handling could affect flag persistence or local file security.
> 
> **Overview**
> Hardens the Node client **platform** layer around flag disk cache and tightens related tests.
> 
> **`NodeStorage`** now treats a missing cache file as a normal first run (no warning, no empty rewrite). Malformed or unreadable cache files trigger a **warn** and recovery write; loaded JSON must be a plain object and only **string** values are kept. Persists via an exclusive **`wx`** temp open (after removing any existing temp path) so writes cannot follow a symlink planted on the temp file path. **`getNodeStorage`** records the first `localStoragePath` and **warns** if a later call passes a different path; **`resetNodeStorage`** clears that path tracking too.
> 
> **Tests** cover first-run behavior, malformed JSON warnings, non-string cache entries, symlink safety, singleton path mismatch warnings, and **`NodeInfo`** no longer pins the SDK **version** string in expectations.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1962f54385c99a17f14d0ddd9a74a9145da1b642. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/sdk/node-client/src/platform/NodeStorage.ts
+++ b/packages/sdk/node-client/src/platform/NodeStorage.ts
@@ -89,8 +89,10 @@ export default class NodeStorage implements Storage {
       if (handle) {
         try {
           await handle.close();
-        } catch {
-          // Ignore close errors during cleanup.
+        } catch (closeError) {
+          this._logger?.warn(
+            `Failed to close storage temp file during cleanup: ${closeError}`,
+          );
         }
       }
       try {
--- a/packages/sdk/node-client/src/platform/NodeStorage.ts
+++ b/packages/sdk/node-client/src/platform/NodeStorage.ts
@@ -157,13 +157,7 @@ export default class NodeStorage implements Storage {
       await this._atomicWriteToFile(new Map(this._cache));
     });
 
-    // Batched callers chain off _flushQueue; log here so a failed write is never silently
-    // masked for callers that did not directly await this flush.
-    this._flushQueue = flush.catch((error) => {
-      this._logger?.error(
-        `Storage flush failed: ${error instanceof Error ? error.message : error}`,
-      );
-    });
+    this._flushQueue = flush.catch(() => {});
     return flush;
   }
 }
--- a/packages/sdk/node-client/src/platform/NodeStorage.ts
+++ b/packages/sdk/node-client/src/platform/NodeStorage.ts
@@ -46,10 +46,16 @@ export default class NodeStorage implements Storage {
       try {
         const data = await fs.readFile(this._storageFile, 'utf8');
         const parsed = JSON.parse(data);
-        if (parsed && typeof parsed === 'object') {
-          this._cache = new Map(Object.entries(parsed as Record<string, string>));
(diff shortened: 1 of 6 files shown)


joker23:
@cursor review

joker23:
@cursor review

joker23:
@cursor review

packages/sdk/node-client/src/platform/NodeStorage.ts
```diff
@@ -62,10 +70,26 @@ export default class NodeStorage implements Storage {
 
   private async _atomicWriteToFile(data: Map<string, string>): Promise<void> {
     const content = JSON.stringify(Object.fromEntries(data));
+    let handle: fs.FileHandle | undefined;
     try {
-      await fs.writeFile(this._tempFile, content, { encoding: 'utf8', mode: 0o600 });
+      try {
+        await fs.unlink(this._tempFile);
```
kinyoklion, line 76: I assume the file being missing isn't the only reason it cannot be deleted. For example if it was created by a different user or with different permissions.

joker23: yea that would also cause issues, but the exception will be caught further down.

packages/sdk/node-client/src/platform/NodeStorage.ts
```diff
@@ -62,10 +70,26 @@ export default class NodeStorage implements Storage {
...
+      } catch {
+        // Ignore if temp file does not exist.
```
kinyoklion:
```suggestion
        // Either the temp file didn't exist, which we don't care about or there was
        // a problem deleting the file, for example a problem with permissions, in
        // which case the subsequent write will likely also fail and be handled by its
        // exception handler.
```

packages/sdk/node-client/src/platform/NodeStorage.ts
```diff
@@ -62,10 +70,26 @@ export default class NodeStorage implements Storage {
...
+      }
+      handle = await fs.open(this._tempFile, 'wx', 0o600);
+      await handle.writeFile(content, 'utf8');
+      await handle.close();
+      handle = undefined;
       await fs.rename(this._tempFile, this._storageFile);
     } catch (error) {
```
kinyoklion, line 88: It does feel like we probably should log something in this error handler.

kinyoklion (approved):

PR #1404, opened 2026-06-01
feat(js-client-sdk): add ability to customize storage impl

joker23:
This PR will enable browser sdk to take in a custom storage implementation.

Additional details:
- added common implementation for client sdks to eventually to make this feature common to all js client sdks
- custom storage implementations will be wrapped to ensure they don't throw

---

> [!NOTE]
> **Low Risk**
> Optional API with safe wrapper; default persistence path is unchanged when `storage` is not set.
> 
> **Overview**
> Adds an optional **`storage`** setting on the browser (and React client) SDK so apps can plug in their own persistence instead of **`window.localStorage`**, while keeping today’s default when the option is omitted.
> 
> Shared client code introduces the **`LDStorage`** contract (`get` / `set` / `clear`) and **`createSafeStorage`**, which adapts custom implementations to the internal storage API: failures are logged, reads return **`null`**, and writes are dropped so a bad implementation cannot take down the host app. The browser client validates **`storage`**, wraps it with **`createSafeStorage`**, and passes it into **`BrowserPlatform`**, which uses the override when provided and otherwise still chooses **`LocalStorage`** only when supported.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a1f4117565ef0a0d9e9d4573296ffb024b59ee81. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/sdk/browser/__tests__/BrowserClient.storage.test.ts
+++ b/packages/sdk/browser/__tests__/BrowserClient.storage.test.ts
@@ -1,0 +1,97 @@
+import { AutoEnvAttributes, LDLogger } from '@launchdarkly/js-client-sdk-common';
+
+import { makeClient } from '../src/BrowserClient';
+
+const mockPlatformConstructor = jest.fn();
+
+jest.mock('../src/platform/BrowserPlatform', () => ({
+  __esModule: true,
+  default: jest.fn().mockImplementation((logger, options, storage) => {
+    mockPlatformConstructor(logger, options, storage);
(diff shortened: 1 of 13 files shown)


joker23:
@cursor review

kinyoklion (approved):
Should extend the 'logs no warnings when all configuration is valid' test with storage like the RN equivalent test.

PR #1426, opened 2026-06-08
ci: allow for prerelease to propagate from `full-release` action

joker23:
This PR should make it easier to do pre-releases without updating the `latest` tag in NPM by letting us propagate the `prerelease` input from the `full-release` action down to the `publish` action.

---

> [!NOTE]
> **Low Risk**
> CI workflow wiring only; default behavior unchanged when `prerelease` is omitted.
> 
> **Overview**
> Adds an optional **`prerelease`** input to the **`full-release`** composite action (default `false`) and forwards it to **`publish`** instead of always passing `false`, so callers can run pre-releases without moving npm’s **`latest`** tag.
> 
> When **`prerelease`** is `true`, the **`publish-docs`** step is skipped so documentation is not published for pre-releases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a47be5bc581297cff40359958703d4c0c7a1e7c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/actions/full-release/action.yml
+++ b/actions/full-release/action.yml
@@ -7,6 +7,10 @@ inputs:
   aws_assume_role:
     description: 'The ARN of an AWS IAM role to assume. Used to auth with AWS to upload results to S3.'
     required: true
+  prerelease:
+    description: 'Is this a prerelease. If so, then the latest tag will not be updated in npm. Defaults to `false`.'
+    required: false
+    default: 'false'
 
 runs:
   using: composite
@@ -32,8 +36,9 @@ runs:
       with:
         workspace_name: ${{ env.WORKSPACE_NAME }}
         workspace_path: ${{ inputs.workspace_path }}
-        prerelease: false
+        prerelease: ${{ inputs.prerelease }}
         dry_run: false
-    - uses: ./actions/publish-docs
+    - if: ${{ inputs.prerelease != 'true' }}
+      uses: ./actions/publish-docs
       with:
         workspace_path: ${{ inputs.workspace_path }}


actions/full-release/action.yml
```diff
@@ -32,8 +36,9 @@ runs:
       with:
         workspace_name: ${{ env.WORKSPACE_NAME }}
         workspace_path: ${{ inputs.workspace_path }}
-        prerelease: false
+        prerelease: ${{ inputs.prerelease }}
         dry_run: false
-    - uses: ./actions/publish-docs
+    - if: ${{ inputs.prerelease != 'true' }}
```
joker23, line 41: I am a bit iffy about this one. On one hand, I think it makes sense to hold docs until `v1`, but on the other, it would be nice to have doc links while we work on official docs.

An alternative here is to introduce different stages of release (something like `experimental`, `prerelease`, and `production`) where `experimental` stage will not have published docs.

kinyoklion: We still have the workflow for publishing docs specifically. So if we want to selectively publish docs we could use that.

Alternatively we could have a "Publish docs for pre-release" checkbox.

Generally speaking I do think it is better to default to not publishing the docs for a pre-release. 

Eventually we may want to switch to having versioned docs and just `/latest` would be used in the actual docs. Then these pre-releases could be prefixed and not typically updates `/latest`

kinyoklion (approved):

PR #1438, opened 2026-06-09
feat: add a custom proxyAgent option to the Node server SDK

bradbunce:
**Requirements**

**Related issues**

N/A — enables proxy configurations (SOCKS in particular) for environments where outbound access to LaunchDarkly is only available through a proxy the SDK does not build itself.

Adds a Node-specific `proxyAgent` option to `@launchdarkly/node-server-sdk`: an `http`/`https` agent that the SDK uses for **all** outgoing connections (polling `fetch` and the streaming event source). This is an extension point for proxy schemes the SDK does not construct itself — a SOCKS proxy being the motivating example. The application builds its own agent (e.g. `new SocksProxyAgent('socks5://host:port')`) and hands it to the SDK.

This mirrors how the Java server SDK lets developers bring their own `socketFactory`, and keeps proxy-scheme knowledge (and correctness) out of the SDK.

- `proxyAgent` is added to the server-node `LDOptions` and wired through `NodePlatform` into `NodeRequests`.
- When `proxyAgent` is set, it is used verbatim and `proxyOptions`/`tlsParams` are ignored (with a warning), because the supplied agent owns connection and TLS setup. `usingProxy()` reports `true` for a custom agent.
- When `proxyAgent` is omitted, behavior is unchanged: the existing `https-proxy-agent` path still handles `proxyOptions` with `http`/`https` schemes.
- No new runtime dependency. `socks-proxy-agent` is only a devDependency, used by the tests.

- **Teaching the SDK about SOCKS schemes directly** (the original version of this PR): adding `socks`/`socks4`/`socks4a`/`socks5`/`socks5h` handling in `proxyOptions` backed by a `socks-proxy-agent` runtime dependency. Per maintainer feedback this was reworked into the agent hook — it covers SOCKS and any other scheme without the SDK owning that logic, avoids the dependency, and matches the Java SDK's bring-your-own approach.

**Additional context**

End-to-end tests drive a user-supplied `SocksProxyAgent` through the hook in polling, streaming, and username/password-auth modes, backed by a minimal RFC 1928/1929 SOCKS5 test server (`__tests__/socksProxyServer.ts`). The full `server-node` test suite passes and lint is clean. `TESTING.md` documents how to consume a local build and configure `proxyAgent`. Relying on CI for the full supported-version matrix (hence the third box left unchecked).

---

> [!NOTE]
> **Medium Risk**
> Changes how outbound HTTP/SSE connections are established for all SDK traffic when the option is used; misconfigured agents could break connectivity, though default paths are unchanged.
> 
> **Overview**
> Adds a Node-specific **`proxyAgent`** initialization option so apps can supply their own `http`/`https` agent for **all** SDK outbound traffic (polling and streaming), enabling proxy schemes the SDK does not build (e.g. SOCKS via `SocksProxyAgent`) without new runtime dependencies.
> 
> When **`proxyAgent`** is set, it is used as-is; **`proxyOptions`** and **`tlsParams`** are ignored and a warning is logged if both are provided. **`usingProxy()`** treats a custom agent as a best-effort “proxy in use” signal. Existing **`proxyOptions`** behavior is unchanged when **`proxyAgent`** is omitted.
> 
> **`NodeRequests`** gains a new constructor parameter (shifting **`enableEventCompression`**); tests cover precedence, warnings, end-to-end client wiring for poll and stream, and agent usage on fetch.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4702973a9ed33cbff185a343a87c225976e62b5b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---

<a href="https://app.devin.ai/review/launchdarkly/js-core/pull/1438" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


bradbunce:
Heads up on CI: the failing checks are all example-app smoke tests (`run-example`, `run-bootstrap-example`, `run-getting-started-example`, `run-react-server-example`). They fail at the `aws-actions/configure-aws-credentials` step ("Could not load credentials from any providers") because this PR is from a fork, and fork PRs don't have access to the repo's AWS/OIDC secrets used to fetch the example SDK keys.

All SDK build/test checks relevant to this change pass, including `build-test-server-node (20)` / `(22)`, `build-test-sdk-server`, and `build-test-common`. The example jobs should pass when re-run with the org's credentials.

joker23:
Hey @bradbunce thanks for the contribution. I will need to speak with the team to see if we agree that this is a broad SDK feature that we want to pursue.

UPDATE
Thinking about this more, I think a better approach here is to expose a way for developers to bring their own proxy agent. For example I _think_ our Java server sdk could support SOCKS proxy because developers can specify a sockFactory implementation. In the node server sdk, allowing developers to specify a proxy agent implementation is a close equivalent (and would capture this feature)

bradbunce:
@joker23 thanks — that makes sense, and I've reworked the PR along those lines. Rather than the SDK knowing about SOCKS schemes, it now exposes a `proxyAgent` option: the app constructs its own `http`/`https` agent (e.g. `new SocksProxyAgent(...)`) and the SDK uses it for all outbound traffic (polling + streaming). This is the Node equivalent of bring-your-own `socketFactory`, covers SOCKS and any other scheme without the SDK owning that logic, and drops the `socks-proxy-agent` runtime dependency (it's now just a devDependency for the tests). When `proxyAgent` is set, `proxyOptions`/`tlsParams` are ignored with a warning. Updated description and tests are pushed — happy to adjust further.

packages/sdk/server-node/src/platform/NodeRequests.ts
```diff
@@ -112,11 +125,12 @@ export default class NodeRequests implements platform.Requests {
...
     tlsOptions?: LDTLSOptions,
     proxyOptions?: LDProxyOptions,
+    proxyAgent?: https.Agent | http.Agent,
     logger?: LDLogger,
     enableEventCompression?: boolean,
   ) {
-    this._agent = createAgent(tlsOptions, proxyOptions, logger);
-    this._hasProxy = !!proxyOptions;
+    this._agent = createAgent(tlsOptions, proxyOptions, proxyAgent, logger);
+    this._hasProxy = !!proxyOptions || !!proxyAgent;
```
joker23, line 146: this is not necessarily true since I could theoretically create a custom agent to do mtls which doesn't have anything to do with proxying.

packages/sdk/server-node/src/platform/NodeRequests.ts
```diff
@@ -80,8 +80,21 @@ function processProxyOptions(
...
   tlsOptions?: LDTLSOptions,
   proxyOptions?: LDProxyOptions,
+  proxyAgent?: https.Agent | http.Agent,
   logger?: LDLogger,
 ): https.Agent | http.Agent | undefined {
+  // A caller-supplied agent takes precedence and is used verbatim. This is the extension point for
+  // proxy schemes the SDK does not build itself (for example a SOCKS proxy via socks-proxy-agent):
+  // the application constructs the agent and the SDK simply uses it. When it is set, proxyOptions
+  // and tlsParams are ignored because the agent owns connection setup.
+  if (proxyAgent) {
```
joker23: instead of escape hatching in the function, I think it should be a branch off https://github.com/launchdarkly/js-core/pull/1438/changes#diff-411bf3104793fc710d18768374badcab71436d8255e104d4d5ce97fcaacf51fbR132 logic. Since we are providing a completely different way to creating the agent.

packages/sdk/server-node/__tests__/socksProxyServer.ts
joker23, line 1: I wonder if we can start an `examples` directory and have a "SOCKS proxy example". Similar to https://github.com/launchdarkly/js-core/tree/main/packages/sdk/react/examples

Main thing here is that, I don't feel too good about including the proxy package as part of dev deps in this project as it is not considered core to the SDK.

packages/sdk/server-node/TESTING.md
joker23, line 1: I don't think this file is in scope of this PR.

joker23:
Hey @bradbunce thanks for the revision, the PR looks a lot better... I've added some more comments. I think some of comments could be address in later PRs (eg creating example SOCKS implementation).

NOTE: I created an internal ticket to track this as `SDK-2508`.

packages/sdk/server-node/src/platform/NodeRequests.ts
joker23: thinking through this comment again, I think it would be safer to keep this logic

kinyoklion (approved):

PR #1760, opened 2026-06-22
fix: Wire timeout option through to flag polling and event delivery requests

joker23:
The timeout configuration was stored in configuration but never threaded
through the datamanger so all connections ran without any timeout
regardless of configuration. This commit will fix this for the node
server sdk (we still need to assess this for client side sdks and see if
it is a problem for other server side sdks).

In addition, this commit will also set the timeout default to 10s which
is consistent with most other server side sdks (eg python, ruby, c++,
etc).

---

> [!NOTE]
> **Medium Risk**
> Changes the default network timeout and alters failure behavior for slow or stuck upstreams during flag polling; apps that relied on 5s or sub-1 values will see different timing without code changes.
> 
> **Overview**
> Fixes **`timeout`** so it actually limits how long flag **polling** HTTP calls can hang, instead of being stored in config but never passed to the platform layer.
> 
> **`Requestor`** now converts configured seconds to milliseconds and sets **`options.timeout`** on polling **`fetch`** calls. **`NodeRequests`** adds a **`timeout`** listener that **`destroy`**s the request with **`Request timed out`** when the socket timeout fires (with a regression test against a non-responding server).
> 
> **Configuration** changes: default **`timeout`** rises from **5s to 10s** (aligned with other server SDKs), validation requires **≥ 1** (invalid **0** clamps to **1** with a warning), and diagnostics/tests reflect the new defaults.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 037606ac68372584109363a9c876aa5bd0661ef2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/shared/sdk-server/src/options/Configuration.ts
```diff
@@ -48,7 +48,7 @@ const validations: Record<string, TypeValidator> = {
   baseUri: TypeValidators.String,
   streamUri: TypeValidators.String,
   eventsUri: TypeValidators.String,
-  timeout: TypeValidators.Number,
+  timeout: TypeValidators.numberWithMin(1),
```
joker23, line 51: There is no specs on this AFAIK, open to discuss... currently the thought is that timeout should exist and must be a valid timeout. Alternatives that I think could work are:
1. accept `0` and make to "no timeout"
2. define a stricter min (eg `5` seconds) to prevent friendly fire

joker23:
@cursor review

packages/shared/sdk-server/src/options/Configuration.ts
kinyoklion: I can be fine with 1 for now, but I do think I would be fine with something larger as well.

joker23: okay, thinking about it more, I think I am more okay with leaving it as is... 1 second could be good for tests I suppose.

kinyoklion (approved):

PR #1790, opened 2026-07-07
ci(vue-sdk): add contract tests

joker23:
SDK-2194

Adding fdv1 and fdv2 contract tests to vue sdk

---

> [!NOTE]
> **Overview**
> Adds **LaunchDarkly SDK contract test** coverage for the Vue client SDK in CI, mirroring the browser SDK harness setup.
> 
> A new `@launchdarkly/vue-contract-test-service` workspace runs a Vite-built app that connects to the test adapter over WebSocket. **`ClientEntity`** maps harness config to Vue’s `createClient` / `client.start()` (including secure-mode `hash` and data-system connection modes) and implements harness commands (evaluate, identify, events, flag listeners). **`vue.yml`** builds the SDK and entity, serves static assets, opens headless Chromium via Playwright, and runs FDv1 and FDv2 harness passes with skip lists for known REPORT/tag header gaps.
> 
> Shared **`SDKConfigClientSideParams`** gains an optional **`hash`** field so secure-mode contract scenarios can be configured.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7ba8ec61bb8a8ff9a1e01da5b4feee8205d70ca1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/vue.yml
+++ b/.github/workflows/vue.yml
@@ -25,4 +25,107 @@ jobs:
           should_build_docs: false
 
   # TODO(scaffold): run-example job arrives when examples/getting-started lands
-  # TODO(scaffold): contract-tests job arrives when contract-tests land
+
+  contract-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+      - uses: ./actions/setup-yarn
+      - name: Install contract test dependencies
+        env:
+          YARN_ENABLE_IMMUTABLE_INSTALLS: 'false'
+        run: yarn workspaces focus @launchdarkly/vue-contract-test-service
+
+      - name: Build the SDK
+        run: yarn workspaces foreach -pR --topological-dev --from '@launchdarkly/vue-client-sdk' run build
+
+      - name: Install Playwright browsers
+        run: yarn workspace @launchdarkly/vue-contract-test-service install-playwright-browsers
+
+      - name: Build shared contract test utils
+        run: yarn workspaces foreach -pR --topological-dev --from '@launchdarkly/js-contract-test-utils' run build
+
+      - name: Build contract test entity (Vue app)
+        run: yarn workspace @launchdarkly/vue-contract-test-service run build
+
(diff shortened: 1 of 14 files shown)


packages/tooling/contract-test-utils/src/types/ConfigParams.ts
```diff
@@ -100,6 +100,12 @@ export interface SDKConfigClientSideParams {
   evaluationReasons?: boolean;
   useReport?: boolean;
   includeEnvironmentAttributes?: boolean;
+  /**
+   * A pre-computed secure mode hash supplied by the test harness. When present, the
+   * client-side SDK includes it as the `h` query parameter on streaming/polling requests.
+   * The SDK does not compute this value; it forwards it.
+   */
+  hash?: string;
```
tanderson-ld, line 103: Consider removing comment, I think the service spec describes what these fields are and should be the source of truth.

.github/workflows/vue.yml
```diff
@@ -25,4 +25,107 @@ jobs:
...
+              echo "HTTP server ready"
+              break
+            fi
+            if [ $i -eq 30 ]; then
+              echo "Timeout waiting for HTTP server"
+              cat /tmp/http-server.log
+              exit 1
+            fi
+            sleep 1
+          done
```
tanderson-ld, line 89: Any better options than these chunky scripts. for detecting the port is ready?

joker23: So this is the same thing we do for https://github.com/launchdarkly/js-core/blob/main/.github/workflows/browser.yml ... I think what we can do is make an executable script that does more or less the same thing in javascript in the shared contract test package?

tanderson-ld: Up to you, if there is precedent and it is consistent, I'm ok with it.

tanderson-ld (approved):

.github/workflows/vue.yml
joker23: I created a task for myself to revisit both CI implementation.

PR #1797, opened 2026-07-08
fix: Disable event processor background flush timers for edge clients

keelerm84:
## Summary

The Fastly Compute SDK creates a new `LDClient` per request and never calls `close()`. Because `sendEvents` defaults on, a real `EventProcessor` is constructed and its `start()` creates `setInterval` flush timers (the event flush timer and the context-deduplication flush timer). Those are cleared only by `EventProcessor.close()`, which the SDK and its example never call, so the timers keep the runtime event loop alive and root the whole client graph in memory -- a leak for a per-request client. (This is also what caused jest to hang without `--forceExit`.)

A background flush timer is not a viable delivery mechanism on Fastly Compute anyway: outbound requests must be made while a request is being handled, and by default each request runs in a fresh sandbox that is torn down afterward. The supported pattern is to flush explicitly with `event.waitUntil(client.flush())`, which the SDK already documents and the example already uses.

### Changes

- Add an opt-in internal option `disableBackgroundEventFlush` (`LDInternalOptions`), threaded through `LDClientImpl`'s FDv1 and FDv2 construction into the `EventProcessor`'s existing `start` parameter. It defaults to `true` when the option is unset, so behavior is unchanged for the Node server SDK and every other consumer.
- Set `disableBackgroundEventFlush: true` in the Fastly edge client.
- Document the per-request client lifecycle and event-flushing model in the README, including the reusable-sandboxes caveat (don't module-scope a client and rely on it to self-flush).
- Add a test asserting the edge client constructs the event processor with `start = false`.

`flush()` and `sendEvent()` do not depend on `start()` having run (they only use constructor-initialized state; `sdk-client` already constructs the processor with `start = false`), so events are still recorded and delivered via explicit `flush()`.

Part of the Fastly refresh epic (SDK-2643). Fixes SDK-2645.

---

> [!NOTE]
> **Medium Risk**
> Touches shared `LDClientImpl` event-processor construction for all server SDKs, though default behavior is preserved unless edge SDKs opt in.
> 
> **Overview**
> Per-request Fastly clients were starting the shared `EventProcessor` background interval timers without ever calling `close()`, which kept the runtime alive and pinned the client in memory. This change adds an internal `disableBackgroundEventFlush` flag on `ServerInternalOptions`, wires it through `LDClientImpl` (FDv1/FDv2) into the existing `EventProcessor` `start` parameter, and enables it for the Fastly `LDClient`.
> 
> **Node and other server SDKs are unchanged** when the option is unset (`startEventProcessor` stays true). Events still buffer and ship via explicit `flush()` / `sendEvent()` without background timers.
> 
> The Fastly README now documents per-request `init()`, `event.waitUntil(ldClient.flush())`, and why module-scoped clients are unsafe with reusable sandboxes. A unit test asserts the edge client passes `start = false` to `EventProcessor`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9040b2331fb936fe04e2405c755184d6b49d681a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/shared/common/src/internal/events/LDInternalOptions.ts
+++ b/packages/shared/common/src/internal/events/LDInternalOptions.ts
@@ -22,17 +22,4 @@ export type LDInternalOptions = {
    * Server - 60s.
    */
   highTimeoutThreshold?: number;
-
-  /**
-   * When true, the event processor is constructed without starting its periodic
-   * background work (the flush and context-deduplication interval timers, and the
-   * diagnostic timer when diagnostics are enabled). Events are still recorded and
-   * are delivered only via explicit flush() calls.
-   *
-   * This is intended for per-request edge SDKs that flush explicitly (for example
-   * via a runtime's waitUntil) and must not leave interval timers running. A live
-   * interval timer keeps the runtime event loop alive and roots the whole client
-   * graph in memory, which for a per-request client is a leak.
-   */
-  disableBackgroundEventFlush?: boolean;
 };
--- a/packages/shared/common/src/internal/events/LDInternalOptions.ts
+++ b/packages/shared/common/src/internal/events/LDInternalOptions.ts
@@ -22,4 +22,17 @@ export type LDInternalOptions = {
    * Server - 60s.
    */
   highTimeoutThreshold?: number;
+
+  /**
+   * When true, the event processor is constructed without starting its periodic
+   * background work (the flush and context-deduplication interval timers, and the
+   * diagnostic timer when diagnostics are enabled). Events are still recorded and
+   * are delivered only via explicit flush() calls.
+   *
+   * This is intended for per-request edge SDKs that flush explicitly (for example
+   * via a runtime's waitUntil) and must not leave interval timers running. A live
+   * interval timer keeps the runtime event loop alive and roots the whole client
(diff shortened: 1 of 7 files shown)


packages/shared/common/src/internal/events/LDInternalOptions.ts
```diff
@@ -22,4 +22,17 @@ export type LDInternalOptions = {
...
+   * background work (the flush and context-deduplication interval timers, and the
+   * diagnostic timer when diagnostics are enabled). Events are still recorded and
+   * are delivered only via explicit flush() calls.
+   *
+   * This is intended for per-request edge SDKs that flush explicitly (for example
+   * via a runtime's waitUntil) and must not leave interval timers running. A live
+   * interval timer keeps the runtime event loop alive and roots the whole client
+   * graph in memory, which for a per-request client is a leak.
+   */
+  disableBackgroundEventFlush?: boolean;
```
joker23: Since this is specific to edge sdks, I think putting this option in `packages/shared/sdk-server/src/options/ServerInternalOptions.ts` is more appropriate.

joker23 (approved):

PR #1799, opened 2026-07-08
fix(node-client-sdk): better handling for bad filesystem states

joker23:
- make sdk refuse to load persistent cache from symlinks (as a security measure) 
- fallsback to inmemory store if filesystem loading fails (instead of throwing)
- additionally, exposing `ConnectionMode` type for downstream reference

---

> [!NOTE]
> **Medium Risk**
> Changes local persistence and symlink handling for the flag cache, which affects security and whether flags survive restarts when the filesystem is misconfigured.
> 
> **Overview**
> **NodeStorage** no longer treats a broken or unsafe local cache as a hard failure. After init problems it keeps serving **get/set/clear** from an in-memory map, logs a single **warn** about the fallback (flags won’t persist across restarts), and skips disk flushes when persistence is disabled.
> 
> **Security / integrity:** Init now rejects a **symlinked storage directory** (`lstat` after `mkdir`) and refuses to load the cache unless **`ldcache.json` is a regular file** (symlinks or other node types are discarded like malformed JSON, without following links into arbitrary files).
> 
> **API:** **`ConnectionMode`** is re-exported from `LDCommon` for downstream typing.
> 
> Tests were expanded for symlinked dirs/files, a file at the storage path, uncreateable paths, and failed rewrite after discarding bad cache; platform tests expect **warn** on init failure instead of per-operation **error** logs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eca6fe615ae2bbeede67e803e0d8665ff5613897. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/sdk/node-client/src/platform/NodeStorage.ts
+++ b/packages/sdk/node-client/src/platform/NodeStorage.ts
@@ -7,7 +7,6 @@ const DEFAULT_DIR_NAME = 'ldclient-user-cache';
 const STORAGE_FILE_NAME = 'ldcache.json';
 
 const ERROR_PREFIX = {
-  get: 'Error getting key from storage',
   set: 'Error setting key in storage',
   clear: 'Error clearing key from storage',
 } as const;
@@ -18,9 +17,9 @@ export default class NodeStorage implements Storage {
   private readonly _storageDir: string;
   private readonly _storageFile: string;
   private readonly _tempFile: string;
-  private readonly _initialized: Promise<boolean>;
+  private readonly _initialized: Promise<void>;
   private readonly _logger?: LDLogger;
-  private _initError?: Error;
+  private _persistenceDisabled: boolean = false;
   private _cache: Map<string, string> = new Map();
   private _flushPending: boolean = false;
   private _flushQueue: Promise<void> = Promise.resolve();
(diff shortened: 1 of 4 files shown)


joker23:
@cursor review

packages/sdk/node-client/src/platform/NodeStorage.ts
```diff
@@ -33,17 +32,29 @@ export default class NodeStorage implements Storage {
...
+  private async _initialize(): Promise<void> {
     try {
       await fs.mkdir(this._storageDir, { recursive: true });
 
+      // fs.mkdir succeeds silently if the path already exists as a symlink to a directory, so a
+      // pre-planted symlink would otherwise redirect where the cache is read from and written to
+      // without ever surfacing as an init failure.
+      if ((await fs.lstat(this._storageDir)).isSymbolicLink()) {
+        throw new Error(`Storage directory path is a symlink, not a real directory: ${this._storageDir}`);
+      }
```
joker23, line 44: I included this check in the default storage implementation to protect against all the funny business symlinks can introduce... after some discussion and thought, this might be too opinionated at this level? I am leaning towards removing this check.

One defense for keeping this is: if an application developer wants to support symlinks or anything, they can override the `storage` implementation.

kinyoklion: My reflex is skepticism.

joker23: okay, I think I'll keep this check in for the default storage class since I do think it is safer. If users don't want this check then they can make their own storage class to override.

We can also offer to suppress this check with an option in the future.

kinyoklion (approved):

PR #1806, opened 2026-07-16
feat: Thread FDv1 fallback TTL through polling and streaming data sources

joker23:
BEGIN_COMMIT_OVERRIDE
chore: Thread FDv1 fallback TTL through polling and streaming data sources
END_COMMIT_OVERRIDE

This commit will add in fallback TTL reading support as well as allow RN native event source read headers.

NOTE: while working this PR, I found a lot of gaps in the RN eventsource. Internally tracked in ticket 2721 in sdk project.

---

> [!NOTE]
> **Overview**
> Extends FDv2 polling and streaming so **FDv1 fallback directives** carry optional TTL (`x-ld-fd-fallback` / `x-ld-fd-fallback-ttl`, plus in-band `protocolFallbackTTL` on goodbye). Result builders now take a shared `FallbackDirective` instead of a bare boolean, and header TTL is interpreted in **seconds** and converted to milliseconds.
> 
> **Orchestration and data manager:** `FDv2DataSource` checks `fdv1Fallback` on **any** initializer result (including transfer-none / HTTP 304-style payloads) and stops the initializer chain when the server directs FDv1. `FDv2DataManagerBase` no longer overwrites the stored selector when a `none` payload arrives without `state`, so a prior full payload’s basis survives 304/none confirmations.
> 
> **Polling:** Goodbye paired with a fallback directive is surfaced as `terminal_error` so the orchestrator moves off FDv1; recoverable errors with fallback skip initializer retries. **Streaming:** Open headers defer fallback until the next queued result; reconnect without fallback clears pending/committed fallback; ping-triggered polls merge deferred TTL without mutating the handler’s result object.
> 
> CI **package size limit** for `@launchdarkly/js-client-sdk-common` is raised from 39300 to 44000 bytes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 18c0cfdd40a0fecc3e6e4dc6060c29dea8b9b26f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


joker23:
@cursor review

joker23:
@cursor review

packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.ts
```diff
@@ -7,10 +7,17 @@
...
  *
  * Additional changes:
- * 1. separated event handling to use onprogress for data changes
- *    and onreadystatechange for status changes. This is to address
- *    an issue with Vega OS where they do not fire a readyStatechange
- *    event when the response is received.
+ * 1. separated event handling to use onprogress for data changes and
+ *    onreadystatechange for status changes, since some platforms (e.g. Vega OS)
+ *    never fire a readystatechange event while the response is still streaming in.
+ * 2. onprogress now also performs the CONNECTING -> OPEN transition and dispatches
```
joker23: Would like to have some thoughts here. This looks like a bug in our previous implementation, but would like to understand if there are any legacy reasons for this.

.github/workflows/sdk-client.yml
```diff
@@ -32,4 +32,4 @@ jobs:
           target_file: 'packages/shared/sdk-client/dist/esm/index.mjs'
           package_name: '@launchdarkly/js-client-sdk-common'
           pr_number: ${{ github.event.number }}
-          size_limit: 39000
+          size_limit: 40000
```
joker23: FYI - needed to bump this up to accomondate the TTL logic :(

joker23:
@cursor review

packages/shared/sdk-client/src/datasource/fdv2/PollingBase.ts
```diff
@@ -55,31 +52,32 @@ function processEvents(
...
-        earlyResult = changeSet(action.payload, fdv1Fallback, environmentId);
+        earlyResult = changeSet(action.payload, fdv1Fallback, environmentId, undefined, fdv1FallbackTtlMs);
         break;
       case 'goodbye':
-        earlyResult = goodbye(action.reason, fdv1Fallback);
+        if (fdv1Fallback) {
+          earlyResult = terminalError(errorInfoFromUnknown(action.reason), true, fdv1FallbackTtlMs);
+        } else {
+          earlyResult = goodbye(action.reason, fdv1Fallback);
+        }
```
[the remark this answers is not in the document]
joker23: Polling payloads should always have a header.

joker23:
@cursor review

packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.ts
```diff
@@ -7,10 +7,23 @@
...
+ *    open, guarded by the same this._status === CONNECTING check onreadystatechange
+ *    already used. Previously open only fired from onreadystatechange at
+ *    readyState DONE, which for a long-lived stream only happens at connection
+ *    close, so the parsed response headers carried on open were never seen while
+ *    the stream was actually running. onprogress fires as soon as the response
+ *    starts loading, so headers are available as soon as the connection opens
+ *    instead of only at teardown.
+ * 3. onprogress now also drives retryAndHandleError/reconnect on an error status,
```
joker23: I think we will need to do a pairing for this. There are lots of gaps in the RN eventsource that are being identified.

packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.ts
```diff
@@ -353,12 +421,28 @@ export default class EventSource<E extends string = never> {
     }
   }
 
+  private _parseResponseHeaders(raw: string | null): Record<string, string> {
+    const result: Record<string, string> = {};
+    if (!raw) {
+      return result;
+    }
+    raw.split('\r\n').forEach((line) => {
```
joker23: I probably should use `JSON.parse`?

packages/sdk/react-native/src/fromExternal/react-native-sse/EventSource.ts
joker23, line 1: I think there is a lot of changes needed to appropriately support FDv2... proposal now is to track this in a separate PR. I propose that we rewrite an EventSource implementation and ONLY use it if FDv2 is specified (to avoid having to major version in case there are incompatibilities).

joker23:
@cursor review

joker23:
@cursor review

packages/shared/sdk-client/src/datasource/fdv2/PollingBase.ts
```diff
@@ -13,11 +13,7 @@ import {
   interrupted,
   terminalError,
 } from './FDv2SourceResult';
-
-function getFallback(headers: { get(name: string): string | null }): boolean {
```
joker23, line 17: replacing with https://github.com/launchdarkly/js-core/blob/e430b6178b550070e0ac897207a28705bed2f7f4/packages/shared/sdk-client/src/datasource/fdv2/fallbackDirective.ts#L27

packages/shared/sdk-client/src/datasource/fdv2/PollingBase.ts
```diff
@@ -109,22 +106,26 @@ export async function poll(
...
+    // Synthesize a 'none' payload so the orchestrator's changeSet path runs
+    // normally without touching stored flags.
     if (response.status === 304) {
       const nonePayload: internal.Payload = {
         version: 0,
         type: 'none',
         updates: [],
       };
-      return changeSet(nonePayload, fdv1Fallback, environmentId);
+      return changeSet(nonePayload, fdv1Fallback, environmentId, undefined, fdv1FallbackTtlMs);
```
tanderson-ld: Possible to get the two fallback parameters next to each other?

tanderson-ld: Maybe change the type of the existing param to be a directive type and that would hold both the bool and the time ms?

packages/shared/sdk-client/src/datasource/fdv2/StreamingFDv2Base.ts
```diff
@@ -127,26 +133,80 @@ export function createStreamingBase(config: {
...
         logConnectionResult(true);
-        resultQueue.put(changeSet(action.payload, fdv1Fallback));
+        if (pendingFallback) {
+          // A fallback directive was deferred from onopen; apply it now that
+          // the payload has been delivered.
+          fdv1Fallback = true;
+          fdv1FallbackTtlMs = pendingFallbackTtlMs;
+          resultQueue.put(changeSet(action.payload, true, undefined, undefined, pendingFallbackTtlMs));
+          pendingFallback = false;
+          pendingFallbackTtlMs = undefined;
```
tanderson-ld: Should this section use `resolveFallback`?

packages/shared/sdk-client/src/datasource/fdv2/StreamingFDv2Base.ts
```diff
@@ -127,26 +133,80 @@ export function createStreamingBase(config: {
...
+    }
+    return { fdv1Fallback, fdv1FallbackTtlMs };
+  }
+
+  function handleAction(action: internal.ProtocolAction, rawData?: unknown): void {
     switch (action.type) {
       case 'payload':
         logConnectionResult(true);
-        resultQueue.put(changeSet(action.payload, fdv1Fallback));
+        if (pendingFallback) {
```
tanderson-ld: Is this if check required?  Is there a way to write it such that it just uses resolveFallback always?

packages/shared/sdk-client/src/datasource/fdv2/StreamingFDv2Base.ts
```diff
@@ -127,26 +133,80 @@ export function createStreamingBase(config: {
...
+        // from onopen (pendingFallback). Surface as a terminal fallback result.
+        const goodbyeDirective = readGoodbyeFallbackDirective(rawData);
+        if (goodbyeDirective.fdv1Fallback || pendingFallback) {
+          const ttlMs = goodbyeDirective.fdv1Fallback
+            ? goodbyeDirective.fdv1FallbackTtlMs
+            : pendingFallbackTtlMs;
+          fdv1Fallback = true;
+          resultQueue.put(terminalError(errorInfoFromUnknown(action.reason), true, ttlMs));
+          pendingFallback = false;
+          pendingFallbackTtlMs = undefined;
```
tanderson-ld: Same comment here.  I thought the purpose of the resolveFallback was to remove duplication of this routine of consuming the pending fallback and clearing values.

packages/shared/sdk-client/src/datasource/fdv2/StreamingFDv2Base.ts
```diff
@@ -238,17 +316,32 @@ export function createStreamingBase(config: {
...
+        // Trust the poll's own fallback signal (e.g. from its own HTTP
+        // response headers) over a directive deferred at onopen, so an
+        // onopen directive still surfaces even while ping-triggered polls
+        // keep succeeding. Still backfill the TTL alone when the poll's own
+        // fallback signal did not carry one, so a deferred TTL is not lost
+        // just because the poll also happened to indicate fallback.
+        const fallback = resolveFallback();
+        if (!result.fdv1Fallback && fallback.fdv1Fallback) {
+          result.fdv1Fallback = true;
+          result.fdv1FallbackTtlMs = fallback.fdv1FallbackTtlMs;
```
tanderson-ld: Is editing result appropriate?  Seems like it shouldn't be mutated from a good practices standpoint.

packages/shared/sdk-client/src/datasource/fdv2/StreamingFDv2Base.ts
tanderson-ld, line 1: Something feels off about how many spots have if checks on pending fallback and logic for moving things around and masking values on top of others.  It just seems very scattered and likely a subtle tweak will mess up the logic.  Is there a way to make it such that all results that go into the queue conform to some type / interface and fallback can be applied to that thing unconditonally, sometimes a no-op but sometimes a real fallback apply occurs.  The pending mechanism could be something that can only apply once and then turns itself into a no-op or something.

packages/shared/sdk-client/src/datasource/fdv2/PollingBase.ts
joker23: done

packages/shared/sdk-client/src/datasource/fdv2/PollingInitializer.ts
```diff
@@ -55,7 +55,7 @@ export function createPollingInitializer(
           return result;
         }
 
-        // Recoverable error — save and potentially retry
+        // Recoverable error: save it and retry if attempts remain
         lastResult = result;
 
         if (attempt < maxRetries) {
```
kinyoklion, line 73: We should be checking the fallback before the retries.

kinyoklion (changes requested):

tanderson-ld (approved):

tanderson-ld:
Probably makes sense to wait for @kinyoklion's review since he left a comment.

packages/shared/sdk-client/src/datasource/FDv2DataManagerBase.ts
```diff
@@ -329,7 +329,12 @@ export function createFDv2DataManagerBase(
       `${logTag} dataCallback: type=${payload.type}, updates=${payload.updates.length}, state=${payload.state}`,
     );
 
-    selector = payload.state;
+    // A 'none' payload (e.g. an HTTP 304, or a streaming 'none' intent)
+    // carries no state -- it confirms existing data is still current and
+    // must not clear a selector already obtained from a prior payload.
+    if (payload.type !== 'none') {
```
kinyoklion, line 336: I think we really need to verify that an intent of none would never change the selector. Because if it did, then I assume we could get drift in the value used for the basis and then trigger a full transfer on a reconnect.

An aside is that a 304 in a browser is a replay, as it is in RN, but maybe it isn't for node client.

kinyoklion: In that the HTTP library in both cases re-presents the cached values.

kinyoklion (approved):

PR #1813, opened 2026-07-20
test: fix contract tests

joker23:
- **chore: disable secure hashing capability in electron contract test**
- **test: fix contract test client entities**


joker23:
I don't think the issue is in the contract test implementation

PR #1816, opened 2026-07-21
fix: add defensive cycle guard to prerequisite evaluation

tanderson-ld:
## Summary

Adds defensive cycle detection to the recursive prerequisite walk in `LDClientImpl._variationInternal`, bringing the JS client-side SDKs' behavior into line with the LaunchDarkly server SDK evaluators.

## The fix

`_variationInternal` now threads a lazily-allocated `Set<string>` through the recursion carrying the flag keys on the current evaluation path. Before descending into a prerequisite, the walker checks whether that key is already on the path; if so, it skips that edge and continues with remaining prerequisites at the same level.

- **Lazy allocation**: variation calls on prereq-less flags (the common case) allocate zero collections. The `Set<string>` is created only when the walker descends into a prerequisite for the first time in a call, then reused for the rest of the walk.
- **Mutable `add` / `try/finally delete`**: no per-descent copy. `ancestors.add(flagKey)` before recursing, `ancestors.delete(flagKey)` in the `finally` — the invariant "the set contains exactly the current path" holds even under early exits.
- **Ancestor-set (current-path) semantics** — this is deliberate. A prerequisite reached via multiple non-cyclic paths (a diamond `A -> [B, C], B -> [D], C -> [D]`) is not a cycle; each path should emit its own prerequisite event. A global visited-set would silently drop the second event; the ancestor-set pattern correctly emits D twice. There is a unit test guarding this.

The optional `visited?: Set<string>` parameter is trailing, so the four existing entry points into `_variationInternal` (`variation`, `variationDetail`, and the two typed-eval callsites) work unchanged — none of the public `variation` / `variationDetail` / `boolVariation` / `intVariation` / `stringVariation` / `doubleVariation` / `jsonVariation` methods need to change.

## Caller-visible behavior on a cycle

The requested flag returns its cached value and reason unchanged. A client-side prerequisite cycle is not surfaced as `MALFORMED_FLAG` and does not fall back to the caller-provided default value. This differs from server-side behavior — the client already holds an authoritative pre-evaluated result from the server; only the ancillary event walk is affected by the cycle.

## Files changed

- `packages/shared/sdk-client/src/LDClientImpl.ts` — cycle guard in `_variationInternal`. Existing callers unchanged.
- `packages/shared/sdk-client/__tests__/LDClientImpl.events.test.ts` — 5 new Jest tests grouped under a nested `describe('prerequisite cycles', ...)`: self-loop, two-cycle (evaluating each end), three-cycle, and a non-cyclic diamond that asserts the shared descendant emits an event for each path (the ancestor-set regression fence).
- `packages/sdk/browser/contract-tests/entity/src/TestHarnessWebSocket.ts` — declares `client-prereq-cycle-detection` so the matching contract tests in sdk-test-harness v2.38.0+ activate for the browser SDK. Downstream wrapper packages (React, React Native, Electron, node-client, Vue) will declare the capability in follow-up PRs against SDK-2711 through SDK-2715.

## Verification

- **Unit tests**: `yarn workspace @launchdarkly/js-client-sdk-common run test` → 834 tests, 0 failures. The 5 new cycle-detection tests are visible under the `prerequisite cycles` group and passing.
- **Contract tests locally** against harness v2.38.1: full monorepo build, browser adapter + entity + headless Chromium, harness run → 831 total, 32 skipped, 799 ran, all passed. Both new suites (`events/prerequisite events handle cycles` and `events/summary events/prerequisites/handles cycles`) fire and pass — 15 subtests total between them.

## Changelog

> Updated prerequisite evaluation event emission to match server SDK behavior for cyclic prerequisite graphs.

[SDK-2707]: https://launchdarkly.atlassian.net/browse/SDK-2707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SDK-2695]: https://launchdarkly.atlassian.net/browse/SDK-2695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Touches the core `variation()` prerequisite/event path, but behavior for non-cyclic flags is unchanged and cycles only skip extra recursion while preserving cached values.
> 
> **Overview**
> **Prerequisite evaluation** in `_variationInternal` now tracks the current dependency path in a lazily allocated `Set` and **skips cyclic edges** instead of recursing forever. Returned flag values and reasons stay the same; only the prerequisite walk (and related feature events) is bounded, matching server SDK semantics. **Diamond graphs** still emit one prerequisite event per distinct path—not a single global “visited” pass.
> 
> Adds **five unit tests** for self-loops, 2/3-cycles, and a non-cyclic diamond regression. The browser contract harness advertises **`client-prereq-cycle-detection`** for harness v2.38+. CI **bundle size limit** for `@launchdarkly/js-client-sdk-common` rises slightly (39,000 → 39,300 bytes).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b03e72245fa2fcbe888a4a1e55578e24944b163. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/shared/sdk-client/src/LDClientImpl.ts
```diff
@@ -679,9 +680,28 @@ export default class LDClientImpl implements LDClient, LDClientIdentifyResult {
...
-    prerequisites?.forEach((prereqKey) => {
-      this._variationInternal(prereqKey, undefined, this._eventFactoryDefault);
-    });
+    if (prerequisites && prerequisites.length > 0) {
+      // Recurse on prerequisites to emulate prereq evaluations occurring with desirable side effects
+      // such as events for prereqs.
+      //
+      // `visited` tracks the chain of prerequisite dependencies from the top-level evaluation to
+      // (but not including) the current flag. It is allocated lazily: variation calls on
+      // prereq-less flags (the common case) allocate zero collections.
```
joker23:
```suggestion
      // Recurse on prerequisites to emulate prereq evaluations occurring with desirable side effects
      // such as events for prereqs.
      //
      // `visited` tracks the chain of prerequisite dependencies from the top-level evaluation to
      // (but not including) the current flag.
```

packages/sdk/browser/contract-tests/entity/src/TestHarnessWebSocket.ts
```diff
@@ -42,6 +42,7 @@ export default class TestHarnessWebSocket {
             'anonymous-redaction',
             'strongly-typed',
             'client-prereq-events',
+            'client-prereq-cycle-detection',
```
joker23, line 45: could we try to enable this capability on `react` sdk as well (since it is just a wrapper)

joker23 (approved):

packages/sdk/browser/contract-tests/entity/src/TestHarnessWebSocket.ts
tanderson-ld: Yes, planning to in the story for React

PR #1827, opened 2026-07-24
feat(sdk-server-common): add support for custom base uris in FDv2 datasources 

joker23:
This will update from a pinned version to follow the v3 release branch

Additionally, this PR will add the ability to customize uris for FDv2 datasources, which is a newer contract test that was failing.

The caveat for adding this is that for standard datasystem types (none custom), we will need to make sure that the baseuri property does not get inherited because standard datasystem settings should only take the top level baseuri.

---

> [!NOTE]
> **Medium Risk**
> Changes which host FDv2 custom composite sources use for polling/streaming; misconfiguration could break flag delivery, though standard modes reject stray `baseUri` and coverage is added.
> 
> **Overview**
> Adds **optional per-source `baseUri`** on FDv2 **custom** polling initializers/synchronizers and streaming synchronizers so each source can hit its own host instead of the shared `serviceEndpoints`. `LDClientImpl` builds scoped `ServiceEndpoints` via `scopedServiceEndpoints` and passes them into `Requestor` / `StreamingProcessorFDv2`; FDv1 fallback polling reuse the same helper.
> 
> **Standard / streaming-only / polling-only** `dataSystem.dataSource` shapes **cannot** set `baseUri` (types omit it; config validation drops it with a warning). Top-level `baseUri` / `streamUri` / `eventsUri` are unchanged.
> 
> Node **contract tests** map harness data-system URIs onto per-entry `baseUri` (not global client URIs). CI runs **v3** contract tests against `version: v3` instead of a pinned alpha.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eb6d22190055dee928348d7be29deb21a8e7bb05. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/server-node.yml
+++ b/.github/workflows/server-node.yml
@@ -53,7 +53,7 @@ jobs:
         with:
           test_service_port: 8000
           token: ${{ secrets.GITHUB_TOKEN }}
-          version: v3.0.0-alpha.6
+          version: v3
           extra_params: '--skip-from=./packages/sdk/server-node/contract-tests/testharness-suppressions-fdv2.txt'
 
   run-custom-socks-proxy-example:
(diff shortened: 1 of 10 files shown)


joker23:
Looks like the scope of this change is a bit bigger. We should support uri customization.

joker23:
@cursor review

joker23:
@cursor review

jsonbailey (approved):

PR #1856, opened 2026-08-13
chore: add FDv2 recovery primitives to SourceManager and Conditions

joker23:
## Summary

Adds `SourceManager.fdv2Recovery()` (reverses `fdv1Fallback()` -- blocks FDv1 slots, unblocks non-FDv1 slots, resets the synchronizer index) and a read-only `SourceManager.isCurrentSynchronizerFDv1Fallback` getter. Also guards `handleFdv1Fallback` in `FDv2DataSource` against re-triggering fallback when the active synchronizer is already the FDv1 fallback slot (regression SDK-2617).

Per review discussion (https://github.com/launchdarkly/js-core/pull/1856#discussion_r3779556509), FDv2 recovery after a directed FDv1 fallback is not modeled as a `Condition` here -- the originally-proposed `Conditions.createFDv2RecoveryCondition(ttlMs)` and `'fdv2Recovery'` condition type were removed. `Condition`s are only ever created through `getConditions()`, which produces no conditions at all when one or zero synchronizers are available -- exactly the state a directed FDv1 fallback puts the SDK in -- so a `Condition`-based recovery attempt would be silently skipped in the case DATASYSTEM v2 section 1.6 requires it to fire unconditionally.

`fdv2Recovery()` / `isCurrentSynchronizerFDv1Fallback` remain primitives with no caller in this diff. The actual recovery scheduling (a plain elapsed-time check, not a `Condition`) is being wired into PR #1858.

---

> [!NOTE]
> **Overview**
> Adds **`SourceManager.fdv2Recovery()`** (mirror of `fdv1Fallback`: block FDv1 slots, unblock FDv2, reset synchronizer index) and **`isCurrentSynchronizerFDv1Fallback`** so the orchestrator can tell when the active synchronizer is already on FDv1.
> 
> **`handleFdv1Fallback`** in `FDv2DataSource` now **skips re-running fallback** when the current synchronizer is an FDv1 fallback slot but still emits `fdv1Fallback` on its results—fixing a loop where FDv2 could be recreated repeatedly (regression **SDK-2617**).
> 
> `Conditions` gets small robustness/docs tweaks: **idempotent fallback timer start** on repeated `interrupted` statuses, and clarified comments on `close()` and the synchronizer `try/finally`.
> 
> Unit tests cover `fdv2Recovery` / `isCurrentSynchronizerFDv1Fallback` and the FDv1 re-trigger guard. **`fdv2Recovery` is not wired into the orchestration loop in this diff** (primitives for a follow-up).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit af8ee0c890258b85cd661f348a0ec58d28411dd3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/shared/sdk-client/src/datasource/fdv2/Conditions.ts
+++ b/packages/shared/sdk-client/src/datasource/fdv2/Conditions.ts
@@ -15,15 +15,15 @@ interface ConditionTimer {
 }
 
 /**
- * Creates a cancelable timer that resolves with the given {@link ConditionType}
- * when it fires. Wraps {@link cancelableTimedPromise} to convert its
- * reject-on-timeout semantics into resolve-with-type semantics.
+ * Wraps {@link cancelableTimedPromise} into a promise that resolves (not rejects)
+ * with a {@link ConditionType} on timeout. `Promise.race` in the orchestration
+ * loop cannot distinguish rejections from other racers, so we must resolve.
  */
 function conditionTimer(timeoutMs: number, type: ConditionType, taskName: string): ConditionTimer {
   const timed = cancelableTimedPromise(timeoutMs / 1000, taskName);
   return {
     promise: timed.promise.then(
-      () => new Promise<ConditionType>(() => {}), // cancelled — never settle
+      () => new Promise<ConditionType>(() => {}), // cancelled - never settle
       () => type, // timeout fired
     ),
     cancel: timed.cancel,
@@ -35,7 +35,7 @@ function conditionTimer(timeoutMs: number, type: ConditionType, taskName: string
  * - `'fallback'`: move to the next available synchronizer
  * - `'recovery'`: reset to the primary synchronizer
  */
-export type ConditionType = 'fallback' | 'recovery';
+export type ConditionType = 'fallback' | 'recovery' | 'fdv2Recovery';
 
 /**
  * A timed condition that races against `synchronizer.next()`. When the
@@ -123,7 +123,7 @@ function createCondition(
     timer = undefined;
   }
 
-  // No inform handler — start immediately (recovery behavior).
+  // No inform handler - start immediately (recovery behavior).
   if (!informHandler) {
     startTimer();
   }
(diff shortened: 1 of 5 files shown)


joker23:
@cursor review

packages/shared/sdk-client/src/datasource/fdv2/Conditions.ts
```diff
@@ -15,15 +15,15 @@ interface ConditionTimer {
 }
 
 /**
- * Creates a cancelable timer that resolves with the given {@link ConditionType}
- * when it fires. Wraps {@link cancelableTimedPromise} to convert its
- * reject-on-timeout semantics into resolve-with-type semantics.
+ * Wraps {@link cancelableTimedPromise} into a promise that resolves (not rejects)
```
kinyoklion: I don't think I am fond of this comment change.

packages/shared/sdk-client/src/datasource/fdv2/Conditions.ts
```diff
@@ -35,7 +35,7 @@ function conditionTimer(timeoutMs: number, type: ConditionType, taskName: string
  * - `'fallback'`: move to the next available synchronizer
  * - `'recovery'`: reset to the primary synchronizer
  */
-export type ConditionType = 'fallback' | 'recovery';
+export type ConditionType = 'fallback' | 'recovery' | 'fdv2Recovery';
```
kinyoklion: I don't think this is the correct modeling.

kinyoklion (changes requested):

packages/shared/sdk-client/src/datasource/fdv2/Conditions.ts
joker23: I figured that we would want to keep the recovery from fdv1 fallback separate from a normal recovery which is why this is a separate enumeration. Is there some thing specific that is bothering you?

kinyoklion: From the discussion we had FDv2 recovery would just be a time, and not a proactive condition.

kinyoklion: So so if the time had elapsed by the next fallback or recovery, then the fallback or recovery would be influenced by it. Otherwise it would be caught the next recovery/fallback that happened.

But there was some nuance about needing to keep trying recovery, when currently we only do it if there is more than one available source, because the source list wasn't expecting temporary modifications.

joker23: Removed the condition type and now recovering from FDv1 fallback is only tracked by the timer.

packages/shared/sdk-client/src/datasource/fdv2/SourceManager.ts
```diff
@@ -213,6 +223,17 @@ export function createSourceManager(
       synchronizerIndex = -1;
     },
 
+    fdv2Recovery() {
+      synchronizerSlots.forEach((slot) => {
+        slot.state = slot.isFDv1Fallback ? 'blocked' : 'available';
+      });
+      synchronizerIndex = -1;
+    },
```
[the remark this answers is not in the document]
joker23: this will be handled later in the stack https://github.com/launchdarkly/js-core/pull/1875

kinyoklion (approved):

PR #1857, opened 2026-08-13
chore: schedule FDv2 recovery orchestration after fallback TTL expires

joker23:
## Summary

Wires the FDv2-recovery primitives from the previous PR in this stack into `FDv2DataSource`'s synchronizer run loop: after falling back to FDv1, a recovery condition races alongside the existing fallback/recovery conditions, using the server-supplied TTL (or a caller-configurable override, or a 1-hour default if absent -- a TTL of `0` means indefinite, no recovery scheduled).

On firing, `SourceManager.fdv2Recovery()` re-enables FDv2 and the loop naturally attempts a fresh FDv2 connection -- if the server still signals fallback on that attempt, fallback re-engages from the new result. Also guards against the FDv1 synchronizer's own results re-triggering the fallback machinery. Ships with regression tests covering TTL present/absent/zero and the no-re-trigger guard.

---

> [!NOTE]
> **Overview**
> Adds a guard in `handleFdv1Fallback` so when the **active synchronizer is already the FDv1 fallback slot**, results that still carry `fdv1Fallback` no longer call `sourceManager.fdv1Fallback()` again. That stops redundant fallback churn (e.g. spinning FDv2 back up or re-blocking slots) while the SDK is already on FDv1.
> 
> A regression test (**SDK-2617**) covers FDv2 signaling fallback, then the FDv1 synchronizer emitting both fallback-flagged and normal changeSets, and asserts FDv2 is only created once and both FDv1 payloads are delivered.
> 
> Also clarifies the synchronizer loop comment: condition timers must be closed on throw or early break, not only on normal paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 63d2f3dacabe344f5d9674704ba9436ae0fc500f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


joker23:
@cursor review

joker23:
Folding this into #1856. After removing the Condition-based FDv2 recovery primitive per review (https://github.com/launchdarkly/js-core/pull/1856#discussion_r3779556509), this layer's only remaining content was the FDv1-fallback re-trigger guard in `handleFdv1Fallback` -- which depends directly on #1856's `isCurrentSynchronizerFDv1Fallback` primitive. Folded that guard + its regression test into #1856 rather than keeping this as a near-empty separate PR. #1858 now rebases directly on top of #1856.

PR #1868, opened 2026-08-19
docs: Document that the Node OpenFeature provider does not report connection state

kinyoklion:
Documents that the Node OpenFeature provider does not surface the state of the connection to LaunchDarkly, instead of implementing partial support for it.

- The provider only emits `ConfigurationChanged`; it never emits `ProviderEvents.Error` or `Stale`, and stays `READY` after a successful start even if the connection is later lost.
- Applications which need connection state can listen to the events of the Node.js SDK client returned by `getClient()`.
- Supporting it properly is contingent on a data source status API in the LaunchDarkly Node.js SDK, which the other server-side SDKs have.

<details>
<summary>Implementation details</summary>

**Requirements**

- [x] I have followed the repository's pull request submission guidelines

**Related issues**

Part of an audit of LaunchDarkly's OpenFeature providers against the current OpenFeature specification. The Java, Ruby, Python and .NET providers report data source state as provider events; the Node provider cannot.

This PR originally mapped the Node SDK's `failed` event to `ProviderEvents.Error`, with recovery reported on the next flag update. That implementation is reverted here: the `failed` event and the flag update event cannot express interruption and recovery, so an unchanged payload after a retried failure would leave the provider in `ERROR` indefinitely. Documenting the gap is preferable to partial support with a state the provider cannot leave.

- Listening to the SDK's `error` event — rejected. It is also emitted for evaluation-time problems such as an unknown flag key, so a single bad evaluation would flip provider status, and it is gated on listener count, which would suppress the SDK's own error logging.
- Emitting `Fatal` — rejected. Fatal makes the OpenFeature SDK short-circuit evaluation and return call-site defaults even though the LaunchDarkly client can still evaluate the flag data it holds.

**Testing**

Documentation only; no code changes remain in this PR.

**Additional context**

No visual preview applies; this is a documentation change to a server-side package.

</details>

Link to Devin session: https://app.devin.ai/sessions/0c452d209ec54b068ba120b4c92b8f6c
Requested by: @kinyoklion

---

> [!NOTE]
> **Overview**
> Documents OpenFeature provider events for the Node server provider: it emits `ConfigurationChanged` on flag updates, but **does not** emit `Error` or `Stale`, and stays `READY` after a successful start even if the LaunchDarkly connection is later lost.
> 
> The README points apps that need connection-loss handling at the underlying Node SDK client from `getClient()`, and notes that full connection-state reporting waits on a data source status API that this SDK does not yet have.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 56b25415040980952854b0c7b72c0cdff8913de0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/shared/openfeature-server-common/src/BaseOpenFeatureProvider.ts
```diff
@@ -104,6 +104,20 @@ export abstract class BaseOpenFeatureProvider<
...
 
+  /**
+   * Emit an OpenFeature Error event. Per-SDK providers call this from their
+   * event wiring when the SDK can no longer receive flag updates.
+   *
+   * This is not reported as a fatal error, because the SDK continues to
+   * evaluate using the flag data it already has.
+   */
+  protected emitError(message: string): void {
+    this.events.emit(ProviderEvents.Error, {
```
joker23: We should also have a place that emits the provider ready event for when the client recovers (https://openfeature.dev/specification/sections/events#requirement-511)

kinyoklion:
I am not sure we should support this. We may instead want to document it is unsupported and have support contingent on implementing data source status in the node server SDK. 

Change this PR to just be the documentation change.

joker23 (approved):

PR #1875, opened 2026-08-21
fix: Close FDv1 fallback and FDv2 recovery correctness gaps

joker23:
> NOTE: stacked on top of https://github.com/launchdarkly/js-core/pull/1982

## Summary

Four fixes found reviewing the recovery-timer PR in this stack, before they shipped:
- Adds the missing "10s uninitialized" leg to the FDv2 fallback condition group: before the data system has ever received data, a source that hangs without producing `interrupted` or a changeSet now still triggers fallback after `initFallbackTimeoutMs` (default 10s), complementing the existing interrupted-based leg.
- Fixes a starvation bug where a repeated fallback directive from a source that never transitions off FDv2 (no FDv1 synchronizer configured) kept re-arming the recovery deadline, so it could never elapse. Now only the first directive per engagement re-arms it.
- Fixes a `'none'`-payload changeSet incorrectly marking the data system initialized, and adds `hasExistingData` so a post-connection-mode-switch `'none'` response on an already-initialized system doesn't churn through synchronizers looking for data that was never going to come.
- Fixes a resource leak where `SourceManager` left the active source open when every synchronizer slot was blocked.

---

> [!NOTE]
> **Overview**
> Tightens **FDv2 data-source orchestration** around fallback conditions, FDv1 directives, and what counts as “initialized.”
> 
> **Uninitialized fallback (10s leg):** Before the data system has ever received real data, synchronizer fallback now includes an **init-fallback timer** (`createInitFallbackCondition`, default 10s) that fires even when the source never reports `interrupted`—so a hung primary can still move to the next synchronizer. Real changeSets cancel it; **`none` payloads do not.** `getConditions` takes an `initialized` flag and `initFallbackTimeoutMs`; `FDv2DataSource` wires `hasExistingData` so connection-mode rebuilds skip this leg.
> 
> **Initialization semantics:** Synchronizer-phase handling no longer calls `markInitialized()` on **`type: 'none'`** changeSets unless `hasExistingData` is set (manager passes prior `initialized` on rebuild). That stops empty poll/304 responses from falsely completing `start()` or disarming the init-fallback leg.
> 
> **FDv1 fallback without an FDv1 slot:** `handleFdv1Fallback` now **always halts** the current run via `sourceManager.fdv1Fallback()` when synchronizers exist (sets `INTERRUPTED` if no FDv1 slot), instead of ignoring the directive. Recovery TTL is **scheduled only once** per engagement—repeated directives from a still-running FDv2 source no longer reset the deadline and starve return to FDv2.
> 
> **Resource cleanup:** `SourceManager` **closes the active synchronizer** when every slot is blocked, so transports are not left running during recovery.
> 
> Tests cover the new conditions, orchestration paths, and a dedicated `FDv1FallbackRecovery.test.ts`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a7a57730d6c06f5e8531cccb6d2e796e82e8b9ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


joker23:
@cursor review

joker23:
@cursor review

PR #1885, opened 2026-08-25
fix(sdk-server-common): use subpath import for `semver` module

joker23:
> NOTE: this PR is stacked on top of https://github.com/launchdarkly/js-core/pull/1972

This PR will use subpath imports for semver to reduce package size per suggestion from @jthorupp

---

> [!NOTE]
> **Overview**
> **`@launchdarkly/js-server-sdk-common` is republished as a dual ESM/CJS package** (tsup instead of `tsc`), with conditional `exports`, `sideEffects: false`, and TypeScript set to `ESNext` / bundler resolution.
> 
> **Semver usage is narrowed** in `Operations.ts` via `semver/functions/parse.js` and `semver/classes/semver` (with an explicit `.js` extension so Node’s ESM loader can resolve the subpath).
> 
> **Integration test helpers** (`FileDataSourceFactory`, `TestData`, builders) are **re-exported by name** from the common package root and from `server-node`’s `integrations.ts` (direct re-export instead of destructuring). **Vercel `EdgeFeatureStore` tests** mock `reviveFullPayload` at module scope and assert call counts on that shared mock, with the real implementation restored in `beforeEach` after `resetAllMocks`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 96e8ab3944f4894c0776f12dd03f31bdacf8ced8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/shared/sdk-server/src/evaluation/Operations.ts
```diff
@@ -1,4 +1,17 @@
...
+ * `parse` uses the `import ... = require(...)` form deliberately (see
+ * https://www.typescriptlang.org/docs/handbook/2/modules.html#es-module-syntax-with-commonjs-behavior).
+ * This package compiles with `"module": "commonjs"` and without `esModuleInterop`, and
+ * `@types/semver` declares that subpath with `export =`, so a default import would
+ * type-check but emit `parse_1.default(...)`, which throws at runtime. We will need to
+ * keep this syntax until this package enables `esModuleInterop` or compiles to ESM.
+ */
+import type SemVer from 'semver/classes/semver';
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+import parse = require('semver/functions/parse');
```
kinyoklion: So, we have leaf-node packages that compile to ESM, this doesn't seem like a safe plan.

kinyoklion (changes requested):

packages/shared/sdk-server/src/evaluation/Operations.ts
joker23: yea this change does have a bad smell to it... it would probably be worth it to hold this off until we can produce shared ESM packages? I think we should head toward that direction so we can do tree-shaking.

packages/shared/sdk-server/src/evaluation/Operations.ts
```diff
@@ -1,4 +1,12 @@
-import { parse, SemVer } from 'semver';
+/*
+ * We import the `semver/functions/parse` subpath instead of the package root to minimize
+ * package size for server SDKs. The `.js` extension is required: this package now
+ * compiles with `"module": "ESNext"` and `"moduleResolution": "bundler"`, and Node's own
+ * ESM resolver (unlike `require` or a bundler's resolution) does not probe for extensions
+ * on bare subpath specifiers, so the published ESM entry point fails to load without it.
```
kinyoklion, line 6: This makes me pretty uneasy. So I am hoping for considerable exploration of how an end-2-end package behaves after this change.

kinyoklion (approved):

packages/shared/sdk-server/tsup.config.ts
```diff
@@ -0,0 +1,22 @@
+// It is a dev dependency and the linter doesn't understand.
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { defineConfig } from 'tsup';
+
+export default defineConfig({
+  entry: {
+    index: 'src/index.ts',
+  },
+  format: ['esm', 'cjs'],
```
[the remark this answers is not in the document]
joker23: This is a false positive - the final compilation happens in the leaf SDK and all of our builds are far past node 14

PR #1982, opened 2026-09-02
feat: Schedule FDv2 recovery after the FDv1 fallback TTL elapses

joker23:
> NOTE: stacked on top of https://github.com/launchdarkly/js-core/pull/1981

## Summary

Wires the recovery half of the FDv1 fallback directive: once the SDK has fallen back to FDv1, it now schedules an unconditional, elapsed-time attempt to return to FDv2 once the fallback TTL elapses, using the `fdv2Recovery()` / `isCurrentSynchronizerFDv1Fallback` primitives added in #1856.

Also corrects the shared `fallbackDirective.ts`'s TTL handling to match the spec: an absent, unparseable, negative, or out-of-range TTL (including `0`, previously treated as indefinite fallback) now clamps to a jittered 1-hour default instead. A server-supplied in-range TTL is used as-is. Directed fallback is therefore never indefinite.

`FDv2RecoveryTimer` is a new standalone primitive holding that single deadline; it is not modeled as a `Condition` since `getConditions()` produces nothing when the SDK has one or zero available synchronizers, which is exactly the state a directed fallback creates. In `FDv2DataSource`, the deadline races the active synchronizer loop, and if every synchronizer slot is blocked before the deadline fires, a background continuation takes over responsibility for the timer and restarts FDv2 sources on its own once it elapses -- without holding up the in-flight `start()`/`identify()` call.

---

> [!NOTE]
> **Overview**
> **After the server directs an FDv1 fallback, the SDK now schedules an unconditional return to FDv2 when the directive TTL elapses**, instead of staying on FDv1 indefinitely when TTL was missing or zero.
> 
> Shared **`resolveFallbackTtlMs`** (and exports of **`DEFAULT_FDV1_FALLBACK_TTL_MS`**) replace prior behavior: only TTLs in **(0, 1 hour]** are honored as given; missing, invalid, zero, negative, or over-one-hour values become a **jittered one-hour default**, so directed fallback always has a concrete retry deadline. Header and goodbye **`protocolFallbackTTL`** parsing both route through this helper.
> 
> **`FDv2RecoveryTimer`** holds that single deadline; **`FDv2DataSource`** arms it on every fallback directive (including while already on the FDv1 synchronizer—reschedule only), races it in the synchronizer loop, and calls **`sourceManager.fdv2Recovery()`** when it fires. If all synchronizer slots are blocked but a deadline remains, a **background continuation** keeps the timer alive and restarts FDv2 without blocking **`start()`** resolution; **`close()`** and orchestration teardown cancel or hand off the timer appropriately.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ec977f238bdf68e59a24281e4fce608541bc6a82. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


joker23:
@cursor review

PR #2002, opened 2026-09-10
chore: Add @launchdarkly/eventsource-node package scaffold

joker23:
## Summary

First layer of the eventsource migration stack. Adds the `@launchdarkly/eventsource-node` package scaffold: package identity (package.json, tsconfigs, tsup, jest, typedoc, docs), the dependency-free `retryDelay` and `capacity` modules with their unit tests, and the public option/event types, which derive from the `@launchdarkly/js-sdk-common` platform contract. `src/index.ts` is a scaffold that exports only the types; the `EventSource` implementation and its export arrive in the next layer. Also wires the package into the root workspaces and tsconfig references, and adds a build-test CI workflow with a TODO for contract tests (completed later in the stack).

---

> [!NOTE]
> **Overview**
> Introduces **`@launchdarkly/eventsource-node`** as the first layer of in-repo EventSource work: package metadata, build/test tooling (tsup, Jest, TypeDoc), license/README/changelog, and monorepo wiring (Yarn workspace, root `tsconfig` project reference, dedicated **`eventsource-node.yml`** CI via shared `./actions/ci`).
> 
> The public surface is intentionally minimal: **`src/index.ts` exports only types** aligned with `@launchdarkly/js-sdk-common` (init options, event payloads, Node TLS/proxy options); the **`EventSource` client is deferred** to a follow-up layer. This PR does land two ported helpers—**`retryDelay`** (backoff, jitter, reset interval) and **`capacity`** (buffer growth with a fixed 8192 minimum instead of reading `Buffer.poolSize`)—each with unit tests; Jest is pinned to **`maxWorkers: 1`** to avoid port collisions from test helpers. Contract tests in CI are left as a TODO.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6cb3662da399e43002e1688679434dd7faca15c3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #2003, opened 2026-09-10
chore: Add the EventSource implementation to @launchdarkly/eventsource-node

joker23:
## Summary

Second layer of the eventsource migration stack. Adds the `EventSource` implementation to `@launchdarkly/eventsource-node`: a close-to-verbatim port of the `launchdarkly-eventsource` npm package's SSE client onto Node `http`/`https`, together with all of its unit tests (the package suite is 114 tests). Two deliberate deviations from the original: TLS certificate verification follows Node's own default (the legacy top-level `rejectUnauthorized` alias and its verification-disabled default are removed), and the Node-inert `withCredentials` option is dropped. Finalizes `src/index.ts` to export the implementation.


joker23:
Closing: superseded by the single fetch-based @launchdarkly/eventsource redesign (see #2002 closing note). A new PR will follow.

PR #2027, opened 2026-09-16
feat: Use Electron's net module as the default HTTP transport

joker23:
This PR will make it so that electron sdk will use the native `net` module as the default HTTP transport.


packages/sdk/electron/src/platform/ElectronRequests.ts
```diff
@@ -45,24 +55,65 @@ export default class ElectronRequests implements platform.Requests {
...
-      );
+        // Node's http/https never followed redirects, sent ambient session credentials, or
+        // served requests from a shared cache; net.request() defaults to all three (`redirect:
+        // 'follow'`, session-ambient auth, `cache: 'default'`). Preserve the prior behavior:
+        // surface any redirect as an error instead of silently following it with credentials
+        // attached, don't attach ambient session auth, and don't let polling/analytics/
+        // diagnostic responses be served from or written into the app's shared HTTP cache.
+        redirect: 'error',
+        credentials: 'omit',
+        cache: 'no-store',
```
joker23, line 73: would like some opinions on this one. I personally don't see a reason to deviate from current behavior (what we have in nodes `http`/`https` implementation)

packages/sdk/electron/__tests__/platform/ElectronRequests.test.ts
```diff
@@ -1,6 +1,127 @@
 import * as http from 'http';
 import * as zlib from 'zlib';
 
+// electron.net isn't available outside a real Electron main process, so this fakes just
+// enough of the ClientRequest API to drive the local test server below over Node's real
+// http module. Verified against real Electron 42.6.0: Chromium's net module transparently
+// decompresses a compressed body while `IncomingMessage.headers` still reports the origin's
+// original `content-encoding` -- this fake reproduces exactly that (see the scrutinize
+// report for the real-Electron verification), rather than handing back a raw, never
+// -transcoded Node response the way a naive passthrough would.
```
joker23, line 10: This is something I had claude whip up (obviously), but it does seem like electron heavily favors e2e tests (https://www.electronjs.org/docs/latest/development/testing)... debating whether or not it is worth it to have an e2e suite that also runs to supplement test coverage.

